PR #52, opened 2019-02-08
Prototype to add config mediaType and annotation support to push

### LINKED ISSUE #50: Should be able to add manifest config and  annotations on push
Should optionally be able to modify a manifest's `Config` and/or `Annotations` on push:
https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go#L25

sajayantony:
This fixes #50 

This is just quick prototype on ORAS to support mediaTypes in config blobs so that registries can determine the artifacts appropriately. 

How do you push config media types ? 

```bash
$ ./oras push localhost:5000/oras:latest -t application/vnd.test -a a=1 -a b=2  ./oras
```
Produces the following manifest 
```
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.test",
    "size": 2,
    "annotations": {
      "a": "1",
      "b": "2"
    }
  },
  "layers": [
    {
      "digest": "sha256:066fed7ff6d005eea332f9ab7da7e815c5c41a6b6c1c3b60c1c497eb630d1f19",
      "size": 11035464,
      "annotations": {
        "org.opencontainers.image.title": "./oras"
      }
    }
  ]
}
```
/cc @shizhMSFT 

## Pending items 

- [ ] the public push method takes too many arguments and potentially needs refactoring, maybe through a stateful model or take a [Manifest](https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go#L20) itself .
- [ ]  If we are supporting root media type then we might want to reserve -t for that 
- [ ] Need to fix tests 
- [ ] Need to consider passing in config blob as a file as well and not just use `{}`

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -16,9 +16,11 @@ type pushOptions struct {
 	targetRef string
 	fileRefs  []string
 
-	debug    bool
-	username string
-	password string
+	configMediaType   string
+	configAnnotations map[string]string
+	debug             bool
+	username          string
+	password          string
 }
 
 func pushCmd() *cobra.Command {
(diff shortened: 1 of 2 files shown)


sajayantony:
This is just a POC so would not recommend merging at this point.  I would rather we refactor the Push method completely to support a usage model here - https://github.com/helm/helm/blob/dev-v3/pkg/registry/client.go#L68
/cc @shizhMSFT

sajayantony:
#56  supersedes this.

PR #53, opened 2019-03-29
Support oras login

shizhMSFT:
This PR supports
- `oras login` to log in to a remote registry.
- `oras logout` to log out from a remote registry.

It re-uses the auth config from docker cli, so as docker credential helpers. By default, it uses the default docker config and it can be overridden by setting the `--config` option. Multiple config files can be taken that if the credential is not found in the first config, `oras` continue to search from the next config file.

For login and logout, the credential changes always occur in the first config file (i.e. all config files except the first one are read-only).

@sajayantony @jdolitsky This is a supportive solution to https://github.com/helm/helm/issues/5461

### DIFF
--- a/pkg/auth/docker/client.go
+++ b/pkg/auth/docker/client.go
@@ -45,6 +45,10 @@ func NewClient(configPaths ...string) (auth.Client, error) {
 	}, nil
 }
 
+func (c *Client) primaryCredentialsStore(hostname string) credentials.Store {
+	return c.configs[0].GetCredentialsStore(hostname)
+}
+
 // loadConfigFile reads the configuration files from the given path.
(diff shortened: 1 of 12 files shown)


pkg/auth/docker/client.go
```diff
@@ -0,0 +1,71 @@
...
+		configs = []*configfile.ConfigFile{cfg}
+	}
+
+	return &Client{
+		configs: configs,
+	}, nil
+}
+
+func (c *Client) primaryCredentialsStore(hostname string) credentials.Store {
+	return c.configs[0].GetCredentialsStore(hostname)
```
jdolitsky: should we check for length >= 1 here?

PR #55, opened 2019-04-03
Directory Support

### LINKED ISSUE #54: Enable folder upload
Push a folder to the remote registry as gzipped tar ball.

shizhMSFT:
This PR enables `oras` to push / pull a directory to / from a remote registry.

To push, the target directory is compressed into a `tar.gz` temproray file (on Linux, it sits at `/tmp`). Then the `tar.gz` file is pushed normally with media type `application/vnd.oci.image.layer.v1.tar+gzip` and extra annotations
- `io.deis.oras.content.digest`: the `sha256` digest of the internal `tar` file. It is different from the digest of the `tar.gz` file.
- `io.deis.oras.content.unpack`: marked as `true` for extration when pull.

To pull, the `tar.gz` file is pulled to a temprorary directory. Then it is extracted and verified against `io.deis.oras.content.digest`.

Notes:
- Since the temp files are introduced, please call `defer store.Close()` for `*FileStore` to release resources.
- The pushed directory can be directly used as a layer of a container image without any modification.

Resolves #54

### DIFF
--- a/pkg/content/file.go
+++ b/pkg/content/file.go
@@ -198,49 +198,11 @@ func (s *FileStore) Writer(ctx context.Context, opts ...content.WriterOpt) (cont
 	if !ok {
 		return nil, ErrNoName
 	}
-	path := s.ResolvePath(name)
-	if !s.AllowPathTraversalOnWrite {
-		base, err := filepath.Abs(s.root)
-		if err != nil {
-			return nil, err
-		}
-		target, err := filepath.Abs(path)
-		if err != nil {
-			return nil, err
-		}
-		rel, err := filepath.Rel(base, target)
-		if err != nil || strings.HasPrefix(filepath.ToSlash(rel), "../") {
-			return nil, ErrPathTraversalDisallowed
-		}
+	path, err := s.resolveWritePath(name)
+	if err != nil {
+		return nil, err
 	}
-	if s.DisableOverwrite {
-		if _, err := os.Stat(path); err == nil {
-			return nil, ErrOverwriteDisallowed
-		} else if !os.IsNotExist(err) {
-			return nil, err
-		}
-	}
-
-	var (
-		file        *os.File
-		err         error
-		afterCommit func() error
-	)
-	if value, ok := desc.Annotations[AnnotationUnpack]; ok && value == "true" {
-		if err := os.MkdirAll(path, 0755); err != nil {
-			return nil, err
-		}
-		file, err = s.tempFile()
-		afterCommit = func() error {
-			checksum := desc.Annotations[AnnotationDigest]
-			return extractTarGzip(path, name, file.Name(), checksum)
-		}
-	} else {
-		if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
-			return nil, err
-		}
-		file, err = os.Create(path)
-	}
+	file, afterCommit, err := s.createWritePath(path, desc, name)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/content/utils.go
```diff
@@ -1,9 +1,163 @@
...
+			if link, err = os.Readlink(path); err != nil {
+				return err
+			}
+		}
+		header, err := tar.FileInfoHeader(info, link)
+		if err != nil {
+			return errors.Wrap(err, path)
+		}
+		header.Name = name
+		header.Uid = 0
```
jdolitsky: Could you explain the uid/gid/uname/gname here? Should we just rely on defaults for these?

pkg/content/utils.go
```diff
@@ -1,9 +1,163 @@
...
+			return err
+		}
+
+		// Name check
+		name := header.Name
+		path, err := filepath.Rel(prefix, name)
+		if err != nil {
+			return err
+		}
+		if filepath.HasPrefix(path, "../") {
```
jdolitsky: I believe this method is deprecated: https://godoc.org/path/filepath#HasPrefix

pkg/content/utils.go
shizhMSFT: They are all information about local machine.
`uid` --> user id in `/etc/passwd`
`uname` --> user name in `/etc/passwd`
`gid` --> group id in `/etc/group`
`gname` --> group name in `/etc/group`

They are meaningful when using `tar` as a backup method, and they are meaningless and leaking privacy when for sharing

pkg/content/utils.go
shizhMSFT: yea... I should just use `strings.HasPrefix` instead

pkg/content/utils.go
```diff
@@ -1,9 +1,164 @@
...
+		header.Uid = 0
+		header.Gid = 0
+		header.Uname = ""
+		header.Gname = ""
+
+		// Write file
+		if err := tw.WriteHeader(header); err != nil {
+			return errors.Wrap(err, "tar")
+		}
+		if mode.IsRegular() {
```
yuwaMSFT2: for non-regular files, they are omitted on purpose. Can you log these so user knows?

pkg/content/file.go
```diff
@@ -138,10 +221,26 @@ func (s *FileStore) Writer(ctx context.Context, opts ...content.WriterOpt) (cont
 		}
 	}
 
-	if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
-		return nil, err
+	var (
```
yuwaMSFT2: The format feels a bit like two funcs. Do think it will look better if moving the validation before this line into a func something like validateWriteOptions?

pkg/content/utils.go
```diff
@@ -1,9 +1,164 @@
...
+	if checksum != "" {
+		if digest, err := digest.Parse(checksum); err == nil {
+			verifier = digest.Verifier()
+			r = io.TeeReader(r, verifier)
+		}
+	}
+	if err := extractTarDirectory(root, prefix, r); err != nil {
+		return err
+	}
+	if verifier != nil && !verifier.Verified() {
```
yuwaMSFT2: move this validation before extractTarDirectory to return early on error

pkg/content/utils.go
shizhMSFT: For non-regular files, they are not ommited. Instead, they are recorded in the tar header.

pkg/content/utils.go
shizhMSFT: No. We cannot do that. The verifier needs to read data first to calculate the digest.

With `io.TeeReader`, here is how data stream flows:
```
gzip.Reader <-- file
     |
     v         +--> extractTarDirectory
io.TeeReader --+
               +--> digest.Verifier
```

pkg/content/file.go
shizhMSFT: refactored to three funcs

pkg/content/utils.go
yuwaMSFT2: I see. So for these file, as long as they appear in the tar header, they will be extracted when doing untar?

yuwaMSFT2:
LGTM

pkg/content/utils.go
shizhMSFT: Hard links,  symbolic links, regular files, directories are extracted. For other files, like unix socks, are ignored.

Issue #63, opened 2019-04-11
Windows absolute path support


shizhMSFT:
Support windows file paths beginging with `C:\`, `D:\`...

Currently, `oras` cannot identify `:` properly as it is the media type separater in oras cli. Since the media type does not contain `:` ( https://tools.ietf.org/html/rfc6838#section-4.2 ), we can safely use the last `:` as the separater.

PR #65, opened 2019-04-12
Cross-platform support on path

### LINKED ISSUE #62: Cross-platform support on path
On windows, running
```
oras push localhost:5000/oras:win dir\hello.txt
```
results in 
```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 2
  },
  "layers": [
    {
      "digest": "sha256:0d74c782a0fd750336d9703eb9995985255bb5dd383c28d1828a6379a5418e4e",
      "size": 14,
      "annotations": {
        "org.opencontainers.image.title": "dir\\hello.txt"
      }
    }
  ]
}
```
Pulling that image under Linux/Unix will result in
```
-rw-rw-rw- 1 shizh shizh 14 Apr 11 18:12 'dir\hello.txt'
```
All paths should be converted to slash-seperated.

shizhMSFT:
Better cross-platform support on Windows.

With this patch,
- Windows path are automatically converted to "/" except absolute paths. For instances,
  - `foo\bar\hello.txt` --> `foo/bar/hello.txt`
  - `C:\hello.txt` --> `C:\hello.txt` (no change)
- Allow file reference with Windows absolute paths. For instance,
  - `C:\hello.txt`
  - `C:\hello.txt:vnd.your.media.type`

Fixes #62, #63

### DIFF
--- a/cmd/oras/push_windows.go
+++ b/cmd/oras/push_windows.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"strings"
+	"unicode"
+)
+
+// parseFileRef parse file reference on windows.
+// Windows systems does not allow ':' in the file path except for drive letter.
+func parseFileRef(ref string, mediaType string) (string, string) {
+	i := strings.Index(ref, ":")
+	if i < 0 {
+		return ref, mediaType
+	}
+
+	// In case it is C:\
+	if i == 1 && len(ref) > 2 && ref[2] == '\\' && unicode.IsLetter(rune(ref[0])) {
+		i = strings.Index(ref[3:], ":")
+		if i < 0 {
+			return ref, mediaType
+		}
+		i += 3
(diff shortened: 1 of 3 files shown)


cmd/oras/push_windows.go
```diff
@@ -0,0 +1,26 @@
+package main
```
jdolitsky: do we need a `// +build` here too?

shizhMSFT: No need for that. Just like `*_test.go`

See https://golang.org/pkg/go/build/#hdr-Build_Constraints

shizhMSFT:
> What is the override flag here to indicate that you are allowing breakout from current directory?

I think you are commeting on #66

PR #66, opened 2019-04-12
Default path validation on push

### LINKED ISSUE #57: Do not allow path traversal by default when pushing package
Do not allow path traversal by default when pushing package. It's a gesture to indicate that the tool considers this as a potential security issue, and it helps to avoid generating package with security hole unexpected. User may choose to allow it but it should not be the default

shizhMSFT:
To generate oras packages / images with good qualities, i.e. secure and cross-platform, `oras` now checks the image title / path name to push with the following rules:
1. Path should not be empty.
2. Path should be clean.
  - `foo/bar` OK
  - `foo/../bar` **NOT OK**
  - `bar` OK
  - `./bar` **NOT OK**
3. Path should be slash-separated. i.e. path must be in the unix format.
  - `foo/bar` OK
  - `foo\bar` **NOT OK**
4. Absolute path is not allowed.
  - `/foo/bar` **NOT OK**
  - `C:\foo\bar` **NOT OK**
  - `C:/foo/bar` **NOT OK**
5. Path traversal is not allowed.
  - `..` **NOT OK**
  - `../bar` **NOT OK**
  - `foo/../../bar` **NOT OK**

Note: All checks are at the `oras` package level, not the CLI level.

If the user is using `oras` CLI, the rule (1) is always true.
With #65, the rules (2)-(4) are always true.
Therefore, this validation is basically equvilant to the rule (5).

To disable path validation, you can either
- Pass option `oras.WithNameValidation(nil)` to `oras.Push()`.
- Run `oras` with `--disable-path-validation`.

Resolves #57


shizhMSFT:
\test

pkg/oras/push_opts_test.go
```diff
@@ -0,0 +1,64 @@
...
+	err = ValidateNameAsPath(descFromName("C:/foo/bar"))
+	suite.Error(err, "windows: absolute path disallowed")
+
+	// disallow path traversal
+	err = ValidateNameAsPath(descFromName(".."))
+	suite.Error(err, "path traversal disallowed")
+	err = ValidateNameAsPath(descFromName("../bar"))
+	suite.Error(err, "path traversal disallowed")
+	err = ValidateNameAsPath(descFromName("foo/../../bar"))
+	suite.Error(err, "path traversal disallowed")
```
jdolitsky: what about something like `./../bar`? Will this get caught w "dirty path" ?

jdolitsky:
Tests failing w
```
        push_opts_test.go:43:                                                                                                                               
                Error Trace:    push_opts_test.go:43                                                                                                        
                Error:          An error is expected but got nil.                                                                                           
                Test:           TestPushOptsSuite/TestValidateNameAsPath                                                                                    
                Messages:       windows: absolute path disallowed 
```

pkg/oras/push_opts_test.go
shizhMSFT: This will be both `dirty_path` and `path_traversal_disallowed`

shizhMSFT:
> Tests failing w
> 
> ```
>         push_opts_test.go:43:                                                                                                                               
>                 Error Trace:    push_opts_test.go:43                                                                                                        
>                 Error:          An error is expected but got nil.                                                                                           
>                 Test:           TestPushOptsSuite/TestValidateNameAsPath                                                                                    
>                 Messages:       windows: absolute path disallowed 
> ```

fixed

PR #67, opened 2019-04-15
Update example code

shizhMSFT:
Close `FileStore` as introduced by #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,15 +191,16 @@ import (
 	"fmt"
 
 	"github.com/containerd/containerd/remotes/docker"
-	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/deislabs/oras/pkg/content"
 	"github.com/deislabs/oras/pkg/oras"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 func check(e error) {
 	if e != nil {
 		panic(e)
 	}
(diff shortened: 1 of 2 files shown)


sajayantony:
Does pull display the OCI manifest digest? If not could we add this to the output?

shizhMSFT:
> Does pull display the OCI manifest digest? If not could we add this to the output?

OCI manifest digest is not visible outside of `oras` package. To be clear, the `oras` users should not bother how the underlying stuffs work. They only need to care about the files to be pushed or pulled. What's purpose of printing the OCI or the docker manifest digest?

In addition, `oras pull` supports index / manifest list with multiple sub-manifests, although `oras push` does not. Which manifest? or all manifests should we print?

sajayantony:
The PR shows the power of ORAS. The contentment addressability is always through digest and unlike a tag is immutable. 
Docker pull today shows you the tag digest so that you can, you want reference by digest. 
In the index case, I believe it is a index manifest since the tag dereferences to that index.

Also ORAS should highlight content addressability since it’s a library in the pure sense. It’s good to not have the user worry about the digest if not required but ORAS having the ability to obtain/save the manifest and output the digest in a sample will be useful. 
This aligns with docker manifest commands in someway I feel.

shizhMSFT:
@sajayantony Let's continue discussion in #68

Issue #71, opened 2019-04-16
Error on logging out a non-existing registry


shizhMSFT:
Logging out a non-existing registry should be considerred as an error.

Requested by https://github.com/helm/helm/pull/5597/commits/f1793e9ca122cdbc9335194d4dd9057a28bbd1b6

PR #73, opened 2019-04-16
Logout returns error if not logged in

shizhMSFT:
`Logout` returns `ErrNotLoggedIn` if a registry is not logged in.
If the multiple config paths are used, the check will verify the login record against all configs. However, it only logs out from the primary config setting since the backup configs are supposed to be read-only.

Resovles #71

### DIFF
--- a/pkg/auth/client.go
+++ b/pkg/auth/client.go
@@ -2,8 +2,14 @@ package auth
 
 import (
 	"context"
+	"errors"
 
 	"github.com/containerd/containerd/remotes"
+)
+
+// Common errors
+var (
+	ErrNotLoggedIn = errors.New("not_logged_in")
 )
 
 // Client provides authentication operations for remotes.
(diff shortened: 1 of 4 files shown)


pkg/auth/client.go
```diff
@@ -2,10 +2,16 @@ package auth
...
 import (
 	"context"
+	"errors"
 
 	"github.com/containerd/containerd/remotes"
 )
 
+// Common errors
+var (
+	ErrNotLoggedIn = errors.New("not_logged_in")
```
bacongobbler: since this is a human-facing error code, would it be better to make this more human readable? "not logged in" might be a better error message than "not_logged_in", especially considering clients like Helm will be returning this error message as output.

pkg/auth/docker/client_test.go
```diff
@@ -0,0 +1,109 @@
+package docker
```
jdolitsky: kudos for new test file!

PR #74, opened 2019-04-16
Error Log Level on Push/Pull by default

shizhMSFT:
By default, the log level is set to `logrus.ErrorLevel` to avoid warnings like
```
WARN[0000] unknown_type: application/vnd.oci.image.config.v1+json
```
To debug, users can always toggle `--debug, -d` options to set the log level to `logrus.DebugLevel`.

@jdolitsky For `helm` invoking `oras` while do not want to bother log level for omitting warnings, please set the proper `GetLogger` at https://github.com/containerd/containerd/blob/master/log/context.go#L31

@sajayantony In terms of the user experiences, what do you think about those warnings?

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -64,11 +64,10 @@ Example - Pull all files, any media type:
 }
 
 func runPull(opts pullOptions) error {
-	if !opts.verbose {
-		logrus.SetLevel(logrus.ErrorLevel)
-	}
 	if opts.debug {
 		logrus.SetLevel(logrus.DebugLevel)
+	} else {
+		logrus.SetLevel(logrus.ErrorLevel)
 	}
 	if opts.allowAllMediaTypes {
 		opts.allowedMediaTypes = nil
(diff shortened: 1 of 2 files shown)


bacongobbler:
Perhaps it'd be better to allow a "verbosity" string setting rather than a debug setting. That way the user can configure between the different log levels logrus provides. e.g.

```
level, err := logrus.ParseLevel(opts.verbosity)
if err != nil {
    ...
}
logrus.SetLevel(level)
```

jdolitsky:
@shizhMSFT I'd like to see the warnings go away, and it seems odd that we would need to import logrus in Helm (or oras for that matter)...  is this common?

shizhMSFT:
@jdolitsky `oras` can use any loggers. The `logrus` package is required by `containerd` as it uses logrus for logging. See: https://github.com/containerd/containerd/blob/master/remotes/handlers.go#L29

shizhMSFT:
Abandoned and succeeded by #80

PR #87, opened 2019-04-19
Encoding check

### LINKED ISSUE #86: Updating from commit 550ed34caf4a to 40d20966e9ea introduces ^M newline characters in code
Example:

```
diff --git a/vendor/github.com/deislabs/oras/pkg/oras/push_opts.go b/vendor/github.com/deislabs/oras/pkg/oras/push_opts.go
index b4d75daf5..fd60d666a 100644
--- a/vendor/github.com/deislabs/oras/pkg/oras/push_opts.go
+++ b/vendor/github.com/deislabs/oras/pkg/oras/push_opts.go
@@ -4,6 +4,7 @@ import (
        "path/filepath"
        "strings"
 
+       "github.com/containerd/containerd/images"^M
        orascontent "github.com/deislabs/oras/pkg/content"
 
@@ -15,6 +16,7 @@ type pushOpts struct {
        configAnnotations   map[string]string
        manifestAnnotations map[string]string
        validateName        func(desc ocispec.Descriptor) error
+       baseHandlers        []images.Handler^M
 }
 
 func pushOptsDefaults() *pushOpts {
@@ -98,3 +100,12 @@ func ValidateNameAsPath(desc ocispec.Descriptor) error {
 
        return nil
 }
+^M
+// WithPushBaseHandler provides base handlers, which will be called before^M
+// any push specific handlers.^M
+func WithPushBaseHandler(handlers ...images.Handler) PushOpt {^M
+       return func(o *pushOpts) error {^M
+               o.baseHandlers = append(o.baseHandlers, handlers...)^M
+               return nil^M
+       }^M
+}^M
```

This is just a subset of the diff, but these `^M` characters are scattered throughout

jdolitsky:
Add test for checking line-endings, also command to fix line endings.

Fixes #86


Makefile
```diff
@@ -46,6 +46,14 @@ build-windows: vendor
 	GOARCH=amd64 CGO_ENABLED=0 GOOS=windows go build -v --ldflags="$(LDFLAGS)" \
 		-o bin/windows/amd64/$(CLI_EXE).exe $(CLI_PKG)
 
+.PHONY: check-encoding
+check-encoding:
+	! find {cmd,pkg,internal} -not -type d -exec file "{}" ";" | grep CRLF
```
shizhMSFT: how about the `examples` folder?

shizhMSFT: How about
```bash
! find {cmd,pkg,internal,examples} -name "*.go" -type f -exec file "{}" ";" | grep CRLF
! find scripts -name "*.sh" -type f -exec file "{}" ";" | grep CRLF
```

Makefile
```diff
@@ -46,6 +46,14 @@ build-windows: vendor
...
+
+.PHONY: fix-encoding
+fix-encoding:
+	find {cmd,pkg,internal} -type f -name "*.go" -exec sed -i -e "s/\r//g" {} +
```
shizhMSFT: `\r$` is safer than `\r`. or should we use something like `dos2unix`?

shizhMSFT: Similar as above. Go files and shell files.

shizhMSFT: NVM with `\r$` since `\r` won't be in the middle anyway

shizhMSFT:
@jdolitsky I have updated the yaml files for codefresh. We are now using smaller image `golang:1.11-alpine` instead of `golang:1.11`. Feel free to release `v0.4.0` once you merged this PR.

.codefresh/release.yml
```diff
@@ -8,11 +8,12 @@ steps:
   UnitTest:
     title: Run unit tests
     stage: test
-    image: golang:1.11
+    image: golang:1.11-alpine
     working_directory: &workdir /go/src/github.com/${{CF_REPO_OWNER}}
     commands:
+    - apk add git make bash file
     - &link ln -s /codefresh/volume/${{CF_REPO_NAME}} ${{CF_REPO_NAME}} && cd ${{CF_REPO_NAME}}
-    - echo make test
```
shizhMSFT: I don't why there is an `echo`. Feel free to revert it.

jdolitsky: must have been from an earlier test

Issue #99, opened 2019-05-01
Move from dep to go modules


jdolitsky:

olegsu:
@jdolitsky I can try to implement it, wdyt?

jdolitsky:
@olegsu go for it. check out this comment which is relevant: https://github.com/deislabs/oras/issues/98#issuecomment-488446081

jdolitsky:
closed by #100

PR #100, opened 2019-05-04
migrate to gomodules

olegsu:
Related to #99
Changes have to be done
1. As @jdolitsky pointed to @thaJeztah https://github.com/deislabs/oras/issues/98#issuecomment-488446081 , docker dependency have to be replaced to direct revision that was used in [Gopkg.lock](https://github.com/deislabs/oras/blob/master/Gopkg.lock#L196) and applied as [psuedo-version](https://golang.org/cmd/go/#hdr-Pseudo_versions) to avoid go mod downgrade (the format used is `vX.Y.(Z+1)-0.yyyymmddhhmmss-abcdefabcdef `) 
2. Indirect dependency of `rsc.io/letsencrypt` have to be replaced with the [forked](github.com/dmcgowan/letsencrypt) version using by `docker/distribution` https://github.com/deislabs/oras/blob/master/Gopkg.toml#L22
3. Remove all vendor rules from `Makefile`
4. Update pipelines of Codefresh to work from the cloned working directory instead of under `GOPATH`

`make test` passed


go.mod
```diff
@@ -0,0 +1,66 @@
...
+	google.golang.org/grpc v1.18.0 // indirect
+	gopkg.in/square/go-jose.v1 v1.1.2 // indirect
+	gopkg.in/yaml.v2 v2.2.2 // indirect
+	gotest.tools v2.2.0+incompatible // indirect
+	rsc.io/letsencrypt v0.0.1 // indirect
+)
+
+replace (
+	github.com/docker/docker => github.com/docker/docker v0.0.0-20190131205458-8a43b7bb99cd
+	rsc.io/letsencrypt => github.com/dmcgowan/letsencrypt v0.0.0-20160928181947-1847a81d2087
```
thaJeztah: Looking at the changes that were made in the fork (related to vendoring), perhaps the fork is no longer needed if I read the description of this release: https://github.com/rsc/letsencrypt/releases/tag/v0.0.1 (but probably best to make that change he upstream in the github.com/docker/distribution package first)

/cc @dmcgowan @caervs

olegsu: Trying to replace it to `replace rsc.io/letsencrypt => github.com/rsc/letsencrypt v0.0.1` breaking the tests at `github.com/deislabs/oras/pkg/auth/docker`
```
+ go test -v -covermode=atomic -coverprofile=.cover/github.com_deislabs_oras_pkg_auth_docker.cover.out github.com/deislabs/oras/pkg/auth/docker
# rsc.io/letsencrypt
../../../go/pkg/mod/github.com/rsc/letsencrypt@v0.0.1/lets.go:674:41: too many arguments in call to c.ObtainCertificate
	have ([]string, bool, nil, bool)
	want ([]string, bool, crypto.PrivateKey)
FAIL	github.com/deislabs/oras/pkg/auth/docker [build failed]
```

thaJeztah: Hm, interesting; https://github.com/rsc/letsencrypt/commit/33926faef6d434b854ea994228f11d0185faa0c1

looks like that may be due to a different version of https://github.com/go-acme/lego being vendored (see https://github.com/go-acme/lego/commit/01974a95cd6e66e2a9a6e49a4b9ef94ff0ae9ec6).

Don't think it has top-priority now to get rid of that fork, and it can be done upstream first (in the docker/distribution repository)

olegsu: Agree, as long as the model gives reproducibility and availability of the modules.

PR #116, opened 2019-08-25
Update to containerd v1.3.0 (beta.2+)

jdolitsky:
Closes https://github.com/deislabs/oras/issues/111


pkg/oras/store.go
```diff
@@ -65,3 +67,54 @@ func (s *hybridStore) Writer(ctx context.Context, opts ...content.WriterOpt) (co
 	}
 	return s.ingester.Writer(ctx, opts...)
 }
+
+// TODO: implement (needed to create a content.Store)
+// TODO: do not return empty content.Info
+// Abort completely cancels the ingest operation targeted by ref.
+func (s *hybridStore) Info(ctx context.Context, dgst digest.Digest) (content.Info, error) {
```
jdolitsky: This is the only method that appears to be used by PushContents(). All tests are all passing, this object appears unused by oras at the moment..

pkg/oras/pull_opts.go
```diff
@@ -23,7 +23,7 @@ type PullOpt func(o *pullOpts) error
 
 func pullOptsDefaults() *pullOpts {
 	return &pullOpts{
-		dispatch:   images.Dispatch,
+		dispatch:   images.Walk,
```
shizhMSFT: We should remain using `images.Dispatch` since it dispatches in parallel (i.e. download in parallel) while `images.Walk` is synchronous.

pkg/oras/store.go
```diff
@@ -6,13 +6,15 @@ import (
...
 	"github.com/containerd/containerd/content"
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 // ensure interface
 var (
 	_ content.Provider = &hybridStore{}
 	_ content.Ingester = &hybridStore{}
+	_ content.Store    = &hybridStore{}
```
shizhMSFT: `content.Store` only is enough since it includes `Provider` and `Ingester`.

pkg/oras/store.go
shizhMSFT: `containerd` introduced `annotateDistributionSourceHandler` in their code which requires `content.Manager`. Also, you are right that `Info()` is the only additional method used by `PushContents()`.

shizhMSFT: From the `containerd` source code, the `annotateDistributionSourceHandler` tries to fetch the `Annotations` for the `Descriptor` from the  `content.Manager` interface. It is safe to return `content.Info{}` to bypass this process since `oras` has already taken care of the annotations in `oras.pack()`.

pkg/oras/store.go
```diff
@@ -65,3 +67,54 @@ func (s *hybridStore) Writer(ctx context.Context, opts ...content.WriterOpt) (co
...
+}
+
+// TODO: implement (needed to create a content.Store)
+// Update updates mutable information related to content.
+// If one or more fieldpaths are provided, only those
+// fields will be updated.
+// Mutable fields:
+//  labels.*
+func (s *hybridStore) Update(ctx context.Context, info content.Info, fieldpaths ...string) (content.Info, error) {
+	panic("not yet implemented: Update (content.Store interface)")
```
shizhMSFT: Should we return `error` instead of `panic()`?

pkg/oras/push.go
```diff
@@ -38,20 +39,27 @@ func Push(ctx context.Context, resolver remotes.Resolver, ref string, provider c
 		return ocispec.Descriptor{}, err
 	}
 
-	desc, provider, err := pack(provider, descriptors, opt)
+	store := newHybridStoreFromProvider(provider)
+	desc, err := pack(store, descriptors, opt)
```
shizhMSFT: Please let `pack()` return the store. i.e.
```go
func pack(provider content.Provider, descriptors []ocispec.Descriptor, opts *pushOpts) (ocispec.Descriptor, content.Store, error)
```

pkg/oras/push.go
```diff
@@ -38,20 +39,27 @@ func Push(ctx context.Context, resolver remotes.Resolver, ref string, provider c
...
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
 
-	if err := remotes.PushContent(ctx, pusher, desc, provider, nil, opt.baseHandlers...); err != nil {
+	var wrapper func(images.Handler) images.Handler
+	if len(opt.baseHandlers) > 0 {
+		wrapper = func(h images.Handler) images.Handler {
+			h = images.Handlers(append(opt.baseHandlers, h)...)
+			return h
```
shizhMSFT: nit: one line?
```golang
return images.Handlers(append(opt.baseHandlers, h)...)
```

jdolitsky:
@shizhMSFT ok, thanks. updated per your comments I've also update readme in prep for a 0.7.0 release

pkg/oras/store.go
jdolitsky: done

pkg/oras/store.go
jdolitsky: done

pkg/oras/pull_opts.go
jdolitsky: done. had to modify signature

PR #126, opened 2019-09-26
strip file times to make tarballs more reproducible

mxey:


mxey:
I also think it would be sensible to strip all file modes except for executable, similar to Git. Should I add this?

jdolitsky:
@mxey sure that sounds good.. @shizhMSFT any thoughts on this?

shizhMSFT:
@mxey @jdolitsky File times are useful for many scenarios, especially for the scenario that we can use the uploaded directory (tar.gz) as a layer of a real image. Thus we should not strip file times.

mxey:
> File times are useful for many scenarios, especially for the scenario that we can use the uploaded directory (tar.gz) as a layer of a real image. Thus we should not strip file times.

@shizhMSFT the same could be said for UID and GID, which you already strip

shizhMSFT:
It is not stripping UID and GID but `chown` to `root`.

BTW, the scenario I have mentioned above can be found at https://github.com/sajayantony/dotnet-oci

jdolitsky:
Perhaps this is something we can pass as feature flag on push if necessary?

mxey:
> It is not stripping UID and GID but chown to root.

OK, in the same sense, I am not stripping times, I am just setting them to epoch 0

mxey:
> Perhaps this is something we can pass as feature flag on push if necessary?

That would work for me.

jzelinskie:
We definitely should have parameters that clarify the manipulation of the tar headers. Different use cases will definitely have different requirements and may want the same tarball with different file attributes to be unique.

It's best to have an opinion, but leave a backdoor. Make the easy things easy, and the hard things possible.

mxey:
@jzelinskie So should I modify my pull request to make this into an option?

jzelinskie:
I think so. What say you @jdolitsky?

jdolitsky:
@mxey @jzelinskie sorry for delay - yes that sounds good

pkg/content/file.go
```diff
@@ -28,6 +28,9 @@ type FileStore struct {
 	DisableOverwrite          bool
 	AllowPathTraversalOnWrite bool
 
+	// Reproducible enables stripping times from added files
+	Reproducible bool
```
shizhMSFT: Maybe `StripTimes` is better?

mxey: @shizhMSFT I decided to make it more generic, in case there are more things to strip

PR #132, opened 2019-10-10
Update module file for 1.13

jpreese:
The dependency resolution problem with `docker` in oras current mod file is causing some pain points when depending on oras (https://github.com/deislabs/oras/issues/125).

This updates the `go.mod` file without the need for replace directives and zero time.Time pseudo versions.


jpreese:
Updating base branch

jdolitsky:
thanks @jpreese !

PR #133, opened 2019-10-11
Allow overriding the ManifestDescriptor when pushing

ecordell:
**What this does**
This PR adds a new field to `PushOpt`, and corresponding config functions, to allow giving an entire manifest descriptor to ORAS when pushing.

**Why**

I've been experimenting with building small, static images using ORAS, while still making them appear as standard docker v2-2 images rather than OCI. 

I found that I could do almost everything I needed to aside from overriding the ManifestDescriptor - which I needed to do to set the MediaType on the manifest (`image-spec` omits this field, but docker registry complains about it).

See where I'm using this here: https://github.com/ecordell/bndlr/blob/master/pkg/bundle/bundle.go#L58-L79

Also open to feedback and suggestions if there are better approaches to what I'm doing. Thanks!

### DIFF
--- a/pkg/oras/push.go
+++ b/pkg/oras/push.go
@@ -83,6 +83,10 @@ func pack(provider content.Provider, descriptors []ocispec.Descriptor, opts *pus
 	}
 
 	// Manifest
+	if opts.manifest != nil {
+		return *opts.manifest, store, nil
+	}
+
 	manifest := ocispec.Manifest{
 		Versioned: specs.Versioned{
 			SchemaVersion: 2, // historical value. does not pertain to OCI or docker version
--- a/pkg/oras/push_opts.go
+++ b/pkg/oras/push_opts.go
@@ -15,6 +15,7 @@ type pushOpts struct {
 	config              *ocispec.Descriptor
 	configMediaType     string
 	configAnnotations   map[string]string
+	manifest            *ocispec.Descriptor
 	manifestAnnotations map[string]string
 	validateName        func(desc ocispec.Descriptor) error
 	baseHandlers        []images.Handler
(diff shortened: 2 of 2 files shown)


pkg/oras/push.go
```diff
@@ -83,24 +83,29 @@ func pack(provider content.Provider, descriptors []ocispec.Descriptor, opts *pus
...
-	manifestBytes, err := json.Marshal(manifest)
-	if err != nil {
-		return ocispec.Descriptor{}, nil, err
-	}
-	manifestDescriptor := ocispec.Descriptor{
-		MediaType: ocispec.MediaTypeImageManifest,
-		Digest:    digest.FromBytes(manifestBytes),
-		Size:      int64(len(manifestBytes)),
+	var manifestDescriptor ocispec.Descriptor
+	if opts.manifest == nil {
```
jzelinskie: if you invert this logic, you can improve readability:

```go
if opts.manifest != nil {
  return *opts.manifest, store, nil
}

manifest := ocispec.Manifest{
   ...
}
...
```

pkg/oras/push_opts.go
```diff
@@ -53,6 +54,14 @@ func WithConfigAnnotations(annotations map[string]string) PushOpt {
 	}
 }
 
+// WithManifest overrides the manifest
```
jzelinskie: this doc should be a little more alarming that it'll invalidate other options like: opts.ManifestAnnotations etc...

pkg/oras/push.go
ecordell: I copied the syntax from the section above that allows overriding the `config` as well. I can update both?

pkg/oras/push_opts.go
ecordell: It should! Again copied from the `WithConfig` which does the same thing for the config manifest

PR #134, opened 2019-10-17
Revert "Update module file for 1.13"

jdolitsky:
Reverts deislabs/oras#132

PR #135, opened 2019-10-17
Update oras modules and pipelines

jpreese:
As theorized in #134, this was Go version difference. I think there's value in wiring up some of these configs (docker builds, maybe have Makefile leverage docker build, codefresh, etc), into the repository itself.

I just ran make build on my local initially, so side stepped all of the CI steps.

For this, I just signed up for Codefresh, added the ymls, and did a Docker build to verify.


jpreese:
As an aside, if we feel this is too big of a jump, we could just remove the line

`github.com/docker/docker v0.0.0-00010101000000-000000000000` in the `go.mod`

and instead make it (without the `replace` directive)

`github.com/docker/docker v0.7.3-0.20190826074503-38ab9da00309`

The _ultimate_ goal here is to remove the zero.Time depedency so others can import this module.

I feel this PR is ultimately where we want to get to with 1.13, but the alternative is still a step forward.

PR #139, opened 2019-10-31
Remove encoding from default mediatype

### LINKED ISSUE #137: Default mediatype should not include encoding
Based on OCI call discussion today, we should remove +json as a default

jdolitsky:
Closes #137

### DIFF
--- a/implementors.md
+++ b/implementors.md
@@ -195,7 +195,7 @@ ACR Artifact Documentation: [aka.ms/acr/artifacts](https://aka.ms/acr/artifacts)
 
   ```sh
   oras push myregistry.azurecr.io/samples/artifact:1.0 \
-      --manifest-config /dev/null:application/vnd.unknown.config.v1+json \
+      --manifest-config /dev/null:application/vnd.unknown.config.v1 \
       ./artifact.txt:application/vnd.unknown.layer.v1+txt
   ```
 
--- a/pkg/artifact/consts.go
+++ b/pkg/artifact/consts.go
@@ -3,5 +3,5 @@ package artifact
 const (
 	// UnknownConfigMediaType is the default mediaType used when no
 	// config media type is specified.
-	UnknownConfigMediaType = "application/vnd.unknown.config.v1+json"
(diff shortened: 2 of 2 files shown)


implementors.md
```diff
@@ -195,7 +195,7 @@ ACR Artifact Documentation: [aka.ms/acr/artifacts](https://aka.ms/acr/artifacts)
 
   ```sh
   oras push myregistry.azurecr.io/samples/artifact:1.0 \
-      --manifest-config /dev/null:application/vnd.unknown.config.v1+json \
+      --manifest-config /dev/null:application/vnd.unknown.config.v1 \
       ./artifact.txt:application/vnd.unknown.layer.v1+txt
```
SteveLasker: Based on our call, the layer should be `application/txt`, however I'm not seeing an existing entry. 
However, if we change the defaults, below, the above examples would be:
```sh
 oras push myregistry.azurecr.io/samples/artifact:1.0 \
      --manifest-config /dev/null \
      ./artifact.txt
```
as the defaults within ORAS would set these to:
`application/unknown.config`
and
`application/txt`

pkg/artifact/consts.go
```diff
@@ -3,5 +3,5 @@ package artifact
 const (
 	// UnknownConfigMediaType is the default mediaType used when no
 	// config media type is specified.
-	UnknownConfigMediaType = "application/vnd.unknown.config.v1+json"
```
SteveLasker: This got a little more complex.
We agreed to pull the encoding off, when the object is null.
When the object is set, it would have the encoding. `+json` seems like a good default.

implementors.md
SteveLasker: I had this blocked for requesting review, but I don't want to block you cutting a build tonight. So, digest the feedback, make a call, and cut the release. The main thing I want to change in this release is having new artifacts default to an OCI image. Only container images, that can be run with docker and other container runtimes should use `application/vnd.oci.image.config.v1`

jdolitsky:
@SteveLasker since there is still some things to sort out, im going to close this and cut the release. we can resubmit later

PR #169, opened 2020-06-27
Refactor docker/NewClient to use early return

KeisukeYamashita:
## What

As title.

## Why

For better code. Before `for`-looping, IMO, it is better to check the length and early return if the length is zero.

### DIFF
--- a/pkg/auth/docker/client.go
+++ b/pkg/auth/docker/client.go
@@ -22,13 +22,6 @@ type Client struct {
 // All changes will only be written to the first config file.
 func NewClient(configPaths ...string) (auth.Client, error) {
 	var configs []*configfile.ConfigFile
-	for _, path := range configPaths {
-		cfg, err := loadConfigFile(path)
-		if err != nil {
-			return nil, errors.Wrap(err, path)
-		}
-		configs = append(configs, cfg)
-	}
 	if len(configs) == 0 {
 		cfg, err := config.Load(config.Dir())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/auth/docker/client.go
```diff
@@ -22,13 +22,6 @@ type Client struct {
...
 func NewClient(configPaths ...string) (auth.Client, error) {
 	var configs []*configfile.ConfigFile
-	for _, path := range configPaths {
-		cfg, err := loadConfigFile(path)
-		if err != nil {
-			return nil, errors.Wrap(err, path)
-		}
-		configs = append(configs, cfg)
-	}
 	if len(configs) == 0 {
```
jdolitsky, line 25: I think this will always be 0 .. should we check `len(configPaths)` instead ?

KeisukeYamashita: Oh, sorry! I've fixed it. Thank you for catching this.

pkg/auth/docker/client.go
```diff
@@ -22,22 +22,26 @@ type Client struct {
 // All changes will only be written to the first config file.
 func NewClient(configPaths ...string) (auth.Client, error) {
 	var configs []*configfile.ConfigFile
```
jdolitsky, line 24: Once last thing is that this var can be defined after the early return (`configs`)

KeisukeYamashita: Thank you 🙇

KeisukeYamashita:
@jdolitsky Is it /test ? https://github.com/deislabs/oras/pull/122#issuecomment-531494124

Do I need to do something? I read the "contributions" section but it didn't tell me what to do after submitting a PR.

jdolitsky:
@KeisukeYamashita yes thanks :)

jdolitsky:
cc @shizhMSFT

PR #179, opened 2020-09-14
move reporting of status to library

### LINKED ISSUE #140: Move verbose logging into the library and out of cmd?
Would you consider moving the verbose logging defined in [pushStatusTrack()](https://github.com/deislabs/oras/blob/master/cmd/oras/push.go#L186-L196) and [pullStatusTrack()](https://github.com/deislabs/oras/blob/master/cmd/oras/pull.go#L117-L133) into `pkg/oras` and out of `cmd/oras`? The logging is highly image pull/push specific, and thus quiet useful to others using the oras library, not just the CLI.

I needed it, ended up duplicating almost the entire bit of code. 

What I did is pass an `io.Writer` to `push`/`pull` so rather than having the library write to stdout with `fmt.Println` (which may not be desired), it writes to the passed writer, if provided. It looks something like this:

```go
func pushStatusTrack(writer io.Writer) images.Handler {
        var printLock sync.Mutex
        return images.HandlerFunc(func(ctx context.Context, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
                if name, ok := content.ResolveName(desc); ok {
                        printLock.Lock()
                        defer printLock.Unlock()
                        writer.Write([]byte(fmt.Sprintf("Uploading %s %s", desc.Digest.Encoded()[:12], name)))
                }
                return nil, nil
        })
}
```

And then you can pass the `io.Writer` and `verbose` option via `opts ...PushOpt`

I don't mind opening a PR for it. 

Thanks!

deitch:
Resolves #140, specifically [this comment](https://github.com/deislabs/oras/issues/140#issuecomment-691729131) by @SteveLasker 

One interesting point. I took it "as is" and gave it the ability to accept an `io.Writer`. Personally, I would prefer something that accepts the logging interface, so that I can pass it as:

```go
oras.WithPushStatusTrack(log)
```

and then rather than calling `fmt.Fprintf(writer)`, we can just call `log.Debug()` or similar.

The problem is that log _isn't_ an interface, it is an actual implementation that people pick. This may be good enough for now, and then we can iterate upon it.


deitch:
This is still open. Can I get a review on it @jdolitsky @SteveLasker ? Let's get it in?

jdolitsky:
We have CI broken right now.. opened #183 to address, then we should be good w/ this

jdolitsky (approved):

PR #182, opened 2020-09-30
create pass-through writer and iowriter

deitch:
As discussed with @SteveLasker , this is the first of 2-3 PRs expanding the capabilities of the `store` package.

There are two additions here:

* `PassthroughWriter` that lets you wrap any other underlying store (e.g. file, memory) and have a processor function in the middle. Thus if I wanted to decrypt the content in transit, or uncompress, or any other action, I have a simple framework for doing so.
* `IoWriter` is very similar to `FileStore`, except instead of taking a directory and opening files, it takes an actual `io.Writer` and passes it through. I have used it to, e.g. write directly to block devices, and to write to another stream over a network. It is more fine-grained control as you can pass it any `io.Writer` you want.

Once these are in, I have two more that take advantage of the `PassthroughWriter`: `GunzipWriter` and `UntarWriter`.

### DIFF
--- a/pkg/content/iowriter.go
+++ b/pkg/content/iowriter.go
@@ -1,0 +1,82 @@
+package content
+
+import (
+	"context"
+	"io"
+	"io/ioutil"
+
+	"github.com/containerd/containerd/content"
+	"github.com/opencontainers/go-digest"
+)
+
+// IoContentWriter writer that wraps an io.Writer, so the results can be streamed to
+// an open io.Writer. For example, can be used to pull a layer and write it to a file, or device.
+type IoContentWriter struct {
+	writer   io.Writer
+	digester digest.Digester
+	size     int64
+}
+
+// NewIoContentWriter create a new IoContentWriter. blocksize is the size of the block to copy,
+// in bytes, between the parent and child. The default, when 0, is to simply use
+// whatever golang defaults to with io.Copy
+func NewIoContentWriter(writer io.Writer, blocksize int) content.Writer {
+	w := writer
+	if w == nil {
+		w = ioutil.Discard
+	}
+	ioc := &IoContentWriter{
+		writer:   w,
+		digester: digest.Canonical.Digester(),
+	}
+	return NewPassthroughWriter(ioc, func(r io.Reader, w io.Writer, done chan<- error) {
+		// write out the data to the io writer
+		var (
+			err error
+		)
+		if blocksize == 0 {
+			_, err = io.Copy(w, r)
+		} else {
+			b := make([]byte, blocksize, blocksize)
+			_, err = io.CopyBuffer(w, r, b)
+		}
+		done <- err
+	})
+}
+
+func (w *IoContentWriter) Write(p []byte) (n int, err error) {
+	n, err = w.writer.Write(p)
+	if err != nil {
+		return 0, err
+	}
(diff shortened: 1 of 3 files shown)


pkg/content/iowriter.go
```diff
@@ -0,0 +1,100 @@
...
+	"io"
+
+	"github.com/containerd/containerd/content"
+	"github.com/opencontainers/go-digest"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// Blocksize size of byte slice for each io copy
+	Blocksize = 20480
```
jdolitsky, line 14: should this be configurable?

deitch: Yeah, it should. I also passed all of my logging stuff through, which is a mistake. I will clean that up too.

deitch: OK, now it is `DefaultBlocksize` and it is an option when creating. Take a look.

deitch:
Thanks for rebasing on master @jdolitsky . Any other comments on it?

jdolitsky:
@deitch code lgtm - can we add a few unit tests for these? Or I suppose they can be added w/ the Gunzip/Untar writers?

deitch:
Yeah, that is a good idea. At least for the `io.Writer` one. I will find some time to do it.

deitch:
A bit odd to do, but I think this works for the test.

jdolitsky:
@deitch can you check failing CI?

deitch:
@jdolitsky I ran it multiple times locally, I cannot see the failure. Could be transient. I am repushing to kick off again

deitch:
Nope that failed too. The failure doesn't appear to have anything to do with this PR, but I am not sure.

```
--- FAIL: TestORASTestSuite (0.27s)
    --- FAIL: TestORASTestSuite/Test_0_Push (0.09s)
    --- FAIL: TestORASTestSuite/Test_1_Pull (0.06s)
    --- PASS: TestORASTestSuite/Test_2_MediaType (0.05s)
    --- PASS: TestORASTestSuite/Test_3_Conditional_Pull (0.05s)
```

Any ideas? Here is the output https://github.com/deislabs/oras/pull/182/checks?check_run_id=1210937435

deitch:
What is wrong with the CI run @jdolitsky ?

deitch:
@jdolitsky I rebased onto master, but also took advantage to simplify the interface, which ended up simplifying the code and blocksize stuff materially, while hewing much more closely to the classic golang style. Take a look, and let's get this in? It has been open for a good few weeks.

pkg/content/iowriter.go
deitch: And now it is gone. It uses `io.Copy` if a blocksize is not set, an `io.CopyBuffer` if a blocksize is set. Much more golang-native-like.

deitch:
And CI is all clean. Even better.

deitch:
Shoot. It deleted all my changes! I have to recreate them!

deitch:
Phew. Managed to recover it all. Back at you @jdolitsky

deitch:
_And_ CI is green.

jdolitsky (approved):
lgtm!

deitch:
Excellent.

I have gunzip and untar writers ready to go based on this.

PR #191, opened 2020-10-27
gunzip and untar passthrough writers

deitch:
This leverages the `PassthroughWriter` to create gunzip and untar writers.

These can be chained. I use it as follows (skipping error handling):

```go
        writer, err := ingester.Writer(ctx, opts...)
        // figure out which writer we need
        hasGzip, hasTar := checkCompression(desc.MediaType)
        if hasTar {
                writer = NewUntarWriter(writer, d.blocksize)
        }
        if hasGzip {
                writer = NewGunzipWriter(writer, d.blocksize)
        }
        // `writer` now will process the stream through gunzip and then through untar before passing it on to the original ingester writer
```

I actually have all of the above wrapped up in a `DecompressStore` which I pass to `oras.Pull`.

### DIFF
--- a/pkg/content/gunzip.go
+++ b/pkg/content/gunzip.go
@@ -1,0 +1,48 @@
+package content
+
+import (
+	"compress/gzip"
+	"fmt"
+	"io"
+
+	"github.com/containerd/containerd/content"
+)
+
+// NewGunzipWriter wrap a writer with a gunzip, so that the stream is gunzipped
+func NewGunzipWriter(writer content.Writer, blocksize int) content.Writer {
+	if blocksize == 0 {
+		blocksize = DefaultBlocksize
+	}
+	return NewPassthroughWriter(writer, func(r io.Reader, w io.Writer, done chan<- error) {
+		gr, err := gzip.NewReader(r)
+		if err != nil {
+			done <- fmt.Errorf("error creating gzip reader: %v", err)
+			return
+		}
+		// write out the uncompressed data
+		b := make([]byte, blocksize, blocksize)
+		for {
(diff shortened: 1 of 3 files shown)


jdolitsky:
@deitch can you rebase when you have a moment?

deitch:
Done.

deitch:
Once this is in, an interesting question would be how to incorporate it directly into the store.

pkg/content/gunzip.go
```diff
@@ -0,0 +1,63 @@
...
+	gzipSuffix := "+gzip"
+	if strings.HasSuffix(mt, gzipSuffix) {
+		mt = mt[:len(mt)-len(gzipSuffix)]
+		gzip = true
+	}
+	if strings.HasSuffix(mt, ".tar") {
+		tar = true
+	}
+	return
+}
```
jdolitsky, line 63: is this used anywhere?

jdolitsky (changes requested):

pkg/content/gunzip.go
deitch: Ah shoot. No, it isn't (for now). I have a `DecompressStore` that wraps any of the other stores (`FileStore` or `MemoryStore`, etc.) that uses the `GunzipWriter` and `UntarWriter`; it uses the above to determine what to wrap.

I will take it out so this is clean to merge in, and then we can discuss if we want to use add that store and how.

deitch: Done. Back to you @jdolitsky

jdolitsky (approved):

PR #192, opened 2020-10-29
options for content.Writers; options to pre-provide input and output hashing, and blocksize

deitch:
In most cases, e.g. a remote OCI registry store, you want to hash the retrieved object and validate it. But sometimes you don't, e.g. if you truly trust the store, or if the trade-off in hashing cost is too high. This gives you the option to skip hashing.

### DIFF
--- a/pkg/content/gunzip.go
+++ b/pkg/content/gunzip.go
@@ -9,9 +9,18 @@ import (
 )
 
 // NewGunzipWriter wrap a writer with a gunzip, so that the stream is gunzipped
-func NewGunzipWriter(writer content.Writer, blocksize int) content.Writer {
-	if blocksize == 0 {
-		blocksize = DefaultBlocksize
+//
+// By default, it calculates the hash when writing. If the option `skipHash` is true,
+// it will skip doing the hash. Skipping the hash is intended to be used only
+// if you are confident about the validity of the data being passed to the writer,
+// and wish to save on the hashing time.
+func NewGunzipWriter(writer content.Writer, opts ...WriterOpt) content.Writer {
+	// process opts for default
+	wOpts := DefaultWriterOpts()
+	for _, opt := range opts {
+		if err := opt(&wOpts); err != nil {
+			return nil
+		}
 	}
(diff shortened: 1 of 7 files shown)


deitch:
@jdolitsky as discussed. I have no idea why CI is failing, looks like some pulls from docker hub?

deitch:
I restructured this to have the options be variadic functions, so the API is unchanged. Very similar to how https://godoc.org/github.com/containerd/containerd/content#WriterOpt works.

deitch:
@jdolitsky I completely restructured this. I realized that the main reason to suppress hashes is likely because you already have them. Also, perhaps more importantly, you sometimes need those hashes for the underlying store, e.g. MemoryStore or FileStore.

So instead of an option to `SkipHash`, this has an option to provide the hash. If the hash is provided, it skips calculations. If you don't care what it is, provide the fixes `BlankHash`.

This also enabled me to set up some tests on it.

pkg/content/gunzip.go
```diff
@@ -9,10 +9,22 @@ import (
 )
 
 // NewGunzipWriter wrap a writer with a gunzip, so that the stream is gunzipped
-func NewGunzipWriter(writer content.Writer, blocksize int) content.Writer {
+//
+// By default, it calculates the hash when writing. If the option `skipHash` is true,
+// it will skip doing the hash. Skipping the hash is intended to be used only
+// if you are confident about the validity of the data being passed to the writer,
+// and wish to save on the hashing time.
+func NewGunzipWriter(writer content.Writer, blocksize int, opts ...WriterOpt) content.Writer {
```
jdolitsky, line 17: Does it make sense to make blocksize an option too? This code has not yet been released yet

jdolitsky (changes requested):

pkg/content/gunzip.go
deitch: Yes! Really good idea! I am putting this in right now. A few minutes...

deitch: Done.

deitch:
Done. Back at you @jdolitsky

jdolitsky (approved):
👍

PR #197, opened 2020-11-10
prevent anyone from accidentally setting the blocksize to zero

deitch:
Although we do have a default, in theory, but if someone actually calls: `WithBlocksize(0)`, it will override the default, and break all sorts of things. This adds a safety check.

### DIFF
--- a/pkg/content/opts.go
+++ b/pkg/content/opts.go
@@ -1,6 +1,8 @@
 package content
 
 import (
+	"errors"
+
 	"github.com/opencontainers/go-digest"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/content/opts.go
```diff
@@ -47,10 +47,13 @@ func WithOutputHash(hash digest.Digest) WriterOpt {
...
 
 // WithBlocksize set the blocksize used by the processor of data.
-// The default is DefaultBlocksize, which is the same as that used by io.Copy
+// The default is DefaultBlocksize, which is the same as that used by io.Copy.
+// Includes a safety check to ensure the caller doesn't actively set it to <= 0.
 func WithBlocksize(blocksize int) WriterOpt {
 	return func(w *WriterOpts) error {
-		w.Blocksize = blocksize
+		if blocksize > 0 {
+			w.Blocksize = blocksize
```
jdolitsky, line 55: Should we check for inverse and just throw an error here?

jdolitsky (changes requested):

pkg/content/opts.go
deitch: I don't understand.

deitch: Do you mean, rather than ignoring/fixing it if the `blocksize <= 0`, treat `blocksize <= 0` as an error?

jdolitsky: > treat blocksize <= 0 as an error?

Yes

deitch: Done.

jdolitsky (approved):

PR #198, opened 2020-11-10
add optional decompressstore

deitch:
The doc describes it. If I pass `FileStore` to `oras.Pull` as the `ingester`, then it just takes the artifact in whatever format it is. That is fine in some use cases. In others, I might have the artifact compressed with gzip, or wrapped in a single tar file. This is most common when using a registry that doesn't have proper artifact support, so every artifact is just tarred and gzipped and given the legacy media-type. It also is common with compressing artifacts, using just gzip.

This creates an optional "wrapper" store, so you can do:

```go
fileStore := content.NewFileStore(path)
decompressStore := NewDecompressStore(fileStore)
oras.Pull(... decompressStore ...)
```

This leverages the `UntarWriter` and `GunzipWriter` from earlier commits.

I think the options to `DecompressStore` could use some work, but let's get the idea out first.

### DIFF
--- a/pkg/content/decompressstore.go
+++ b/pkg/content/decompressstore.go
@@ -1,0 +1,90 @@
+package content
+
+import (
+	"context"
+	"strings"
+
+	ctrcontent "github.com/containerd/containerd/content"
+)
+
+// DecompressWriter store to decompress content and extract from tar, if needed, wrapping
+// another store. By default, a FileStore will simply take each artifact and write it to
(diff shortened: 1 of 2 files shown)


jdolitsky (changes requested):
possible to add unit test?

deitch:
Done. Back at you.

jdolitsky (approved):

PR #199, opened 2020-11-10
add multireader

deitch:
Provides option to wrap multiple stores to get a single `content.Provider` that can be used anywhere a `Provider` is needed.

### DIFF
--- a/pkg/content/multireader.go
+++ b/pkg/content/multireader.go
@@ -1,0 +1,40 @@
+package content
+
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


jdolitsky (changes requested):
possible to add unit test?

deitch:
Yeah, this one, I think, should be doable.

deitch:
Updated with a test

jdolitsky (approved):

PR #204, opened 2020-12-14
Fix `pkg/auth/docker` test

shizhMSFT:
Fix build failure due to errors in `pkg/auth/docker/client_test.go`. The root cause is that the test registry may not be running before any test cases.


jdolitsky (approved):
👍

PR #214, opened 2021-01-28
multiwriteringester support

deitch:
This is a first cut attempt at this. It needs proper tests, which I am willing to add once we have agreement on the idea.

@jdolitsky this is what we discussed over Slack.

## Problem

The [DecompressStore](https://github.com/deislabs/oras/blob/0bffdbc8b717fdf774259c96f7ac285089f8ceaa/pkg/content/decompressstore.go#L22) has support for determining that a particular blob is tarred and then untarring it. This is great, but it assumes that the tar contains only one file. It even [ignores the filename](https://github.com/deislabs/oras/blob/0bffdbc8b717fdf774259c96f7ac285089f8ceaa/pkg/content/untar.go#L30-L35):

```go
			_, err := tr.Next()
			if err == io.EOF {
				// clear the error, since we do not pass an io.EOF
				err = nil
				break // End of archive
			}
```

One could argue about whether we should or should not support tar with multiple files in it, but as long as tar supports it, people _will_ use it, break this, and we will have issues.

The heart of the issue is that the standards assume one writer per descriptor. But if a descriptor references a tar file (or any other archiving format, but really only tar is used), then it might need multiple writers.

## Solution

This PR is a proposed solution. It creates two kinds of "untar writer" - one that has a single writer passed through, and another that has a "multi-writer" passed through. The multi-writer creates multiple writers, one for each filename, and untarwriter picks which one based on the filename.

It continues to have a regular `Writer`, so that non-tar layers can be handled.

The usage would be something like:

```go
        multiStore := NewMultiStore(rootPath) // some store that can handle different filenames
        decompressStore := store.NewDecompressStore(multiStore, WithBlocksize(blocksize), WithMultiWriterIngester())
```

where `multiStore` implements `MultiWriterIngester`:

```go
type MultiWriterIngester interface {
	ctrcontent.Ingester
	Writers(ctx context.Context, opts ...ctrcontent.WriterOpt) (map[string]ctrcontent.Writer, error)
}
```

Calling `.Writers()` returns a map of filename to `content.Writer`, which is passed to `UntarWriter`, which uses the appropriate one.

## Alternate

I did consider an alternate approach, one where, instead of returning `map[string]Writer`, it returned a func that can be called to get the writer:

```go
	Writers(ctx context.Context, opts ...ctrcontent.WriterOpt) (func (filename string) ctrcontent.Writer, error)
```

But I honestly couldn't see how the complexity added value. 

Looking forward to feedback.

### DIFF
--- a/pkg/content/decompressstore.go
+++ b/pkg/content/decompressstore.go
@@ -2,12 +2,13 @@ package content
 
 import (
 	"context"
+	"errors"
 	"strings"
 
 	ctrcontent "github.com/containerd/containerd/content"
 )
 
-// DecompressWriter store to decompress content and extract from tar, if needed, wrapping
+// DecompressStore store to decompress content and extract from tar, if needed, wrapping
 // another store. By default, a FileStore will simply take each artifact and write it to
 // a file, as a MemoryStore will do into memory. If the artifact is gzipped or tarred,
 // you might want to store the actual object inside tar or gzip. Wrap your Store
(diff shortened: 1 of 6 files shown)


jdolitsky (approved):
lgtm!

PR #215, opened 2021-01-28
Fix OCIStore inconsistent state after AddReference

distorhead:
Update descriptor in the nameMap as well as in index.Manifests.

nameMap used in the ListReferences operation and should return actual descriptor.


jdolitsky:
thank you. Is there any way to add a test for the inconsistent state to prevent in the future?

distorhead:
Sure, I will add unit test for this case ASAP.

jdolitsky (approved):

PR #218, opened 2021-02-09
actual versions in go.mod

deitch:
Because `replace` directive is ignored when this is used as a library import.


deitch:
I don't understand the test failures. All work well locally.

jdolitsky:
@deitch, i retriggered it twice and appears to be passing now

jdolitsky (approved):

deitch:
Works now!

PR #220, opened 2021-02-10
use function instead of static map

deitch:
The multiwriter passed a `map[string]content.Writer`. The problem is that the thing that creates these writers sometimes doesn't know what writer it will use until later. This changes it to a func, so that it can determine it on the fly. The net result is the same, with some more flexibility.

Discussed on slack with @jdolitsky 

I still have some work to do to be sure this is ready, but the initial view can get looked at.


deitch:
Removed `[WIP]`

deitch:
And added a test for the new PassthroughMultiWriter

deitch:
Fully ready for a review @jdolitsky

jdolitsky (approved):

Issue #225, opened 2021-02-22
default hybrid store on pull does not pass manifest and index through to ingester


deitch:
Looking at this code [here](https://github.com/deislabs/oras/blob/a64b0ac4faf26c58016fbd7ed1a9eb9d049e54ee/pkg/oras/store.go#L56-L68):

```go
// Writer begins or resumes the active writer identified by desc
func (s *hybridStore) Writer(ctx context.Context, opts ...content.WriterOpt) (content.Writer, error) {
	var wOpts content.WriterOpts
	for _, opt := range opts {
		if err := opt(&wOpts); err != nil {
			return nil, err
		}
	}

	if isAllowedMediaType(wOpts.Desc.MediaType, ocispec.MediaTypeImageManifest, ocispec.MediaTypeImageIndex) || s.ingester == nil {
		return s.cache.Writer(ctx, opts...)
	}
	return s.ingester.Writer(ctx, opts...)
}
```

If the media type is a manifest or an index, it stores it in the cache - which can be a good thing, especially when the need to walk children comes about - but it doesn't pass it through to the ingester, as `s.cache` is just a `MemoryStore`.

Why would we want only to send the config and layers (and whatever other media types there are) through to the ingester? We don't know how it uses it, shouldn't it be up to the ingester to decide?

Strangely, I have had this work, so maybe I am missing something, or maybe it didn't work and I didn't fully realize it?

cc @jdolitsky @shizhMSFT

deitch:
FWIW, if we can agree on this, I am happy to open a PR for this one and pass them through. I already solved it locally. Let me know.

jdolitsky:
I would be fine with an an additional option such as `WithCachedMediaTypes` that allows you to supply your own, defaulting to `[]string{wOpts.Desc.MediaType, ocispec.MediaTypeImageManifest, ocispec.MediaTypeImageIndex}`

deitch:
That could solve both. I will provide a PR for both the caching pass through and the mediatypes next week.

PR #227, opened 2021-02-22
cache docker types too

### LINKED ISSUE #226: hybrid store caching misses docker media types but ChildrenHandler expects them
The caching code is [here](https://github.com/deislabs/oras/blob/a64b0ac4faf26c58016fbd7ed1a9eb9d049e54ee/pkg/oras/store.go#L64-L66):

```go
	if isAllowedMediaType(wOpts.Desc.MediaType, ocispec.MediaTypeImageManifest, ocispec.MediaTypeImageIndex) || s.ingester == nil {
		return s.cache.Writer(ctx, opts...)
	}
```

However, the containerd Children handler [here](https://github.com/containerd/containerd/blob/v1.4.3/images/image.go#L330-L369):

```go
	switch desc.MediaType {
	case MediaTypeDockerSchema2Manifest, ocispec.MediaTypeImageManifest:
		p, err := content.ReadBlob(ctx, provider, desc)
		if err != nil {
			return nil, err
		}

		// TODO(stevvooe): We just assume oci manifest, for now. There may be
		// subtle differences from the docker version.
		var manifest ocispec.Manifest
		if err := json.Unmarshal(p, &manifest); err != nil {
			return nil, err
		}

		descs = append(descs, manifest.Config)
		descs = append(descs, manifest.Layers...)
	case MediaTypeDockerSchema2ManifestList, ocispec.MediaTypeImageIndex:
		p, err := content.ReadBlob(ctx, provider, desc)
		if err != nil {
			return nil, err
		}

		var index ocispec.Index
		if err := json.Unmarshal(p, &index); err != nil {
			return nil, err
		}

		descs = append(descs, index.Manifests...)
	default:
		if IsLayerType(desc.MediaType) || IsKnownConfig(desc.MediaType) {
			// childless data types.
			return nil, nil
		}
		log.G(ctx).Debugf("encountered unknown type %v; children may not be fetched", desc.MediaType)
	}
```

expects the ability to read from the provider (in our case, the cache) for all _four_ media types (OCI manifest and index, docker manifest and manifest list). This means that if we actually do pull down some such from the docker types, it will try to read it, but it isn't cached, and will return a "not found" error.

The simplest solution appears to be to add the two media types - I have a PR about to submit for it - but I am concerned that the lack of passthrough (see #225) might mean this would break something else.

deitch:
Fixes #226 

But worth awaiting the result of #225 before merging in to see any impact.

### DIFF
--- a/pkg/oras/pull.go
+++ b/pkg/oras/pull.go
@@ -61,7 +61,7 @@ func fetchContent(ctx context.Context, fetcher remotes.Fetcher, desc ocispec.Des
 
 	store := opts.contentProvideIngester
 	if store == nil {
-		store = newHybridStoreFromIngester(ingester)
+		store = newHybridStoreFromIngester(ingester, opts.cachedMediaTypes)
 	}
 	handlers := []images.Handler{
 		filterHandler(opts, opts.allowedMediaTypes...),
--- a/pkg/oras/pull_opts.go
+++ b/pkg/oras/pull_opts.go
@@ -21,6 +21,7 @@ type pullOpts struct {
 	callbackHandlers       []images.Handler
 	contentProvideIngester orascontent.ProvideIngester
 	filterName             func(ocispec.Descriptor) bool
+	cachedMediaTypes       []string
 }
 
 // PullOpt allows callers to set options on the oras pull
@@ -28,8 +29,26 @@ type PullOpt func(o *pullOpts) error
 
 func pullOptsDefaults() *pullOpts {
 	return &pullOpts{
-		dispatch:   images.Dispatch,
-		filterName: filterName,
+		dispatch:         images.Dispatch,
+		filterName:       filterName,
+		cachedMediaTypes: []string{ocispec.MediaTypeImageManifest, ocispec.MediaTypeImageIndex},
+	}
+}
+
+// WithCachedMediaTypes sets the media types normally cached in memory when pulling.
+func WithCachedMediaTypes(cachedMediaTypes ...string) PullOpt {
+	return func(o *pullOpts) error {
+		o.cachedMediaTypes = cachedMediaTypes
+		return nil
+	}
+}
+
+// WithAdditionalCachedMediaTypes adds media types normally cached in memory when pulling.
+// This does not replace the default media types, but appends to them
+func WithAdditionalCachedMediaTypes(cachedMediaTypes ...string) PullOpt {
+	return func(o *pullOpts) error {
+		o.cachedMediaTypes = append(o.cachedMediaTypes, cachedMediaTypes...)
+		return nil
 	}
 }
 
--- a/pkg/oras/push.go
+++ b/pkg/oras/push.go
@@ -57,7 +57,7 @@ func Push(ctx context.Context, resolver remotes.Resolver, ref string, provider c
 
 //func pack(store *hybridStore, descriptors []ocispec.Descriptor, opts *pushOpts) (ocispec.Descriptor, error) {
 func pack(provider content.Provider, descriptors []ocispec.Descriptor, opts *pushOpts) (ocispec.Descriptor, content.Store, error) {
-	store := newHybridStoreFromProvider(provider)
+	store := newHybridStoreFromProvider(provider, nil)
 
 	// Config
 	var config ocispec.Descriptor
--- a/pkg/oras/store.go
+++ b/pkg/oras/store.go
@@ -20,22 +20,25 @@ var (
 )
 
 type hybridStore struct {
-	cache    *orascontent.Memorystore
-	provider content.Provider
-	ingester content.Ingester
+	cache            *orascontent.Memorystore
+	cachedMediaTypes []string
+	provider         content.Provider
+	ingester         content.Ingester
 }
 
-func newHybridStoreFromProvider(provider content.Provider) *hybridStore {
+func newHybridStoreFromProvider(provider content.Provider, cachedMediaTypes []string) *hybridStore {
 	return &hybridStore{
-		cache:    orascontent.NewMemoryStore(),
-		provider: provider,
+		cache:            orascontent.NewMemoryStore(),
+		cachedMediaTypes: cachedMediaTypes,
+		provider:         provider,
 	}
 }
 
-func newHybridStoreFromIngester(ingester content.Ingester) *hybridStore {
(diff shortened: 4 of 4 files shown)


deitch:
This is the last of the 3 (I think) PRs to handle the caching issue, discussed with @jdolitsky. We had discussed whether or not we should be adding the legacy docker types, in addition to the OCI types, to the list of types to be cached.

This PR assumes they should. You had suggested at one point to make the OCI types be the default, and have an option to override the types to be cached.

I am ok with that, in principle, but it creates some unanswered questions:

### append vs replace 

Should the option _append_ types, or _replace_ them? 

I think _replace_ is just simpler, especially if we make the default types available via a function that `hybridStore` uses. So if I want to replace, I just replace; if I want to append, I call `getDefaultTypes()` (or whatever) and then just append to the slice, and pass that as a replace.

### where to pass the option

The bigger issue is where to pass this option? It only is relevant to [newHybridStoreFromIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/store.go#L35), which currently does not have options, although we can add them. But `hybridStore` (and `newHybridStoreFrom*()`) are private. [newHybridStoreFromIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/store.go#L35) is called only by [fetchContent](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L47) (again private), which in turn is called by [Pull](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L19).

The only exposed place is [Pull](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L19). We could pass an option there, but that seems distant from what it is dealing with. To have an option to modify how the deep down cache behaves when calling [Pull](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L19) seems a bit disjoint. Then again, we have an option [WithContentProvideIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull_opts.go#L78), so perhaps it is fine?

I see three options:

1. Just take this as is. It certainly is the simplest option by far.
2. Create a new func that returns a [PullOpt](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull_opts.go#L27), named something like `WithCacheMediaTypes(...MediaType)`
3. Make `hybridStore` public/exported, give it a creation-time option to modify cached types, and then consumers can use [WithContentProvideIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull_opts.go#L78)

My obvious personal preference is just this PR. It is by far the simplest (2-line change). Josh, thoughts?

deitch:
This is the last of the 3 (I think) PRs to handle the caching issue, discussed with @jdolitsky. We had discussed whether or not we should be adding the legacy docker types, in addition to the OCI types, to the list of types to be cached.

This PR assumes they should. You had suggested at one point to make the OCI types be the default, and have an option to override the types to be cached.

I am ok with that, in principle, but it creates some unanswered questions:

### append vs replace 

Should the option _append_ types, or _replace_ them? 

I think _replace_ is just simpler, especially if we make the default types available via a function that `hybridStore` uses. So if I want to replace, I just replace; if I want to append, I call `getDefaultTypes()` (or whatever) and then just append to the slice, and pass that as a replace.

### where to pass the option

The bigger issue is where to pass this option? It only is relevant to [newHybridStoreFromIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/store.go#L35), which currently does not have options, although we can add them. But `hybridStore` (and `newHybridStoreFrom*()`) are private. [newHybridStoreFromIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/store.go#L35) is called only by [fetchContent](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L47) (again private), which in turn is called by [Pull](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L19).

The only exposed place is [Pull](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L19). We could pass an option there, but that seems distant from what it is dealing with. To have an option to modify how the deep down cache behaves when calling [Pull](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull.go#L19) seems a bit disjoint. Then again, we have an option [WithContentProvideIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull_opts.go#L78), so perhaps it is fine?

I see three options:

1. Just take this as is. It certainly is the simplest option by far.
2. Create a new func that returns a [PullOpt](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull_opts.go#L27), named something like `WithCacheMediaTypes(...MediaType)`
3. Make `hybridStore` public/exported, give it a creation-time option to modify cached types, and then consumers can use [WithContentProvideIngester](https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/pkg/oras/pull_opts.go#L78)

My obvious personal preference is just this PR. It is by far the simplest (2-line change). Josh, thoughts?

jdolitsky:
@deitch - option 2 is still my preference. This enables oras to be fully agnostic to mediatypes

deitch:
You would want the harder one, wouldn't you? 

Let's see what we can do.

deitch:
@jdolitsky take a look at it now. See if this works.

jdolitsky (approved):
this looks great

deitch:
I was waiting for you to merge, then realized, wait a sec, I don't need to. I can do it once you approved (which you did) 😃

PR #228, opened 2021-02-23
handle different gzip extensions

deitch:
I should have done this from the beginning. The decompress store treats `+gzip` as gzip and `.tar` as tar. This is correct, but there still its plenty out there that uses `.gzip`

This just handles that equally sanely. Adds it to the test as well, because we are well-behaved citizens. 😄


jdolitsky (approved):
this lgtm!

PR #230, opened 2021-02-28
pass through on write

### LINKED ISSUE #225: default hybrid store on pull does not pass manifest and index through to ingester

deitch:
Fixes #225 

This does not change what gets cached or not - expect a separate PR on that - but it has the cache be write-through.

Current behaviour: if the item is to be cached, cache it; if not, pass it through to the underlying store. This means the underlying store doesn't even have the opportunity to decide if it wants to do anything with the cached data. It is to cache or to underlying store, but not both.

PR behaviour: if the item is to be cached, cache it *and* pass it through to the underlying store; if not, *just* pass it through to the underlying store. The cache becomes a write-through cache.

Of course, the underlying store still might decide to reject or discard.

### DIFF
--- a/pkg/oras/oras_test.go
+++ b/pkg/oras/oras_test.go
@@ -362,7 +362,7 @@ func (suite *ORASTestSuite) Test_4_GHSA_g5v4_5x39_vwhx() {
 			suite.FailNow("error creating temp directory", err)
 		}
 		defer os.RemoveAll(tempDir)
-		store := orascontent.NewFileStore(tempDir)
+		store := orascontent.NewFileStore(tempDir, orascontent.WithIgnoreNoName())
 		defer store.Close()
 		ref = fmt.Sprintf("%s/evil:%s", suite.DockerRegistryHost, tag)
 		_, _, err = Pull(newContext(), newResolver(), ref, store)
--- a/pkg/oras/store.go
+++ b/pkg/oras/store.go
@@ -3,12 +3,15 @@ package oras
 import (
 	"context"
 	"errors"
+	"io"
+	"time"
 
 	orascontent "github.com/deislabs/oras/pkg/content"
 
 	"github.com/containerd/containerd/content"
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/errgroup"
 )
 
 // ensure interface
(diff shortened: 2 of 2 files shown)


deitch:
Trying to figure out what CI is complaining about.

deitch:
I figured it out. The pass-through works fine. The problem is that if the underlying store cannot handle certain things - e.g. `content.FileStore` requires a name annotation on everything, hence the CI `no_name` unexpected error - then the fact that we did not used to pass through config and manifest and index, and now do, causes the issue.

Essentially, the `hybridStore` was acting as an implicit filter for certain types that the `FileStore` cannot handle, and so FIleStore errors out.

What is the right way to handle this?

jdolitsky:
I think you need to modify the test to pull using `WithPullEmptyNameAllowed()` here: https://github.com/deislabs/oras/blob/main/pkg/oras/oras_test.go#L368

Although that makes me think this is a backwards-incompatible change? I'm surprised if we dont have pull tests for the FileStore besides related to the recent vulnerability

deitch:
> I think you need to modify the test to pull using WithPullEmptyNameAllowed()

I had already tried that; it doesn't work. `WithPullEmptyNameAllowed()` only affects the pull filters. The problem here is once it passes that, and it sends the content to the `FileStore`. That doesn't know what to do with content from descriptors that have no name, so it returns an error.

I _think_ the right thing to do is to modify `FileStore` so that it has an option to ignore, rather than error on, content that has no name. If you think that is right, I am happy to modify `NewFileStore()` to do that, then this one will start to make sense.

deitch:
I definitely want to move this ahead. @jdolitsky do you think the above approach is correct, i.e. an option to `FileStore` to ignore content it cannot handle? Or should we handle it differently?

jdolitsky:
@deitch sry for delay-

> I think the right thing to do is to modify FileStore so that it has an option to ignore

This sounds good to me

jdolitsky:
@deitch - OK to close this now after #239 ?

deitch:
Not yet. I'll update this one tomorrow.

deitch:
@jdolitsky I rebased this to main, and updated it to use `WithIgnoreNoName()` on creating a FIleStore. All of those tests pass now.

Once this is in place, I will do one more for having options to override allowed media-types [here](https://github.com/deislabs/oras/blob/a64b0ac4faf26c58016fbd7ed1a9eb9d049e54ee/pkg/oras/store.go#L64) as you had suggested.

jdolitsky (approved):

PR #239, opened 2021-03-11
filestore option to ignore noname error

deitch:
When we try to write to a FileStore (or more accurately, create a Writer from a FileStore), using a Descriptor that doesn't have a title annotation - `ocispec.AnnotationTitle` - then FileStore returns an `ErrNoName`.

This adds an option to ignore that error, i.e. if you pass it a descriptor without a name, `FileStore` will just ignore it, writing it to `ioutil.Discard`. This option allows one to blindly pass an entire OCI manifest and all of its dependencies to FileStore, which only will write down the ones it knows how to handle.

The default, of course, remains unchanged.

Also added a basic test.

As discussed with @jdolitsky in #230

### DIFF
--- a/pkg/content/file.go
+++ b/pkg/content/file.go
@@ -31,19 +31,28 @@ type FileStore struct {
 	// Reproducible enables stripping times from added files
 	Reproducible bool
 
-	root       string
-	descriptor *sync.Map // map[digest.Digest]ocispec.Descriptor
-	pathMap    *sync.Map
-	tmpFiles   *sync.Map
+	root         string
+	descriptor   *sync.Map // map[digest.Digest]ocispec.Descriptor
+	pathMap      *sync.Map
+	tmpFiles     *sync.Map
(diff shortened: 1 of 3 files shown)


pkg/content/file.go
```diff
@@ -35,15 +35,24 @@ type FileStore struct {
 	descriptor *sync.Map // map[digest.Digest]ocispec.Descriptor
 	pathMap    *sync.Map
 	tmpFiles   *sync.Map
+	ignoreNoName bool
```
jdolitsky, line 38: can you gofmt this file?

jdolitsky (changes requested):

pkg/content/file.go
deitch: That's funny. I thought I had. Clearly, I had not.

Updated. Ran `gofmt` on all of the files changed.

jdolitsky (approved):

PR #241, opened 2021-03-15
Add GitHub Container Registry Implementor

markphelps:
👋🏻 Hi from GitHub! We'd like to add our [Container Registry](https://docs.github.com/en/packages/guides/about-github-container-registry) implementation to your implementors guide.

Please let me know if there is anything else needed!


awakecoding:
Nice! I was actually wondering if the GitHub container registry supported OCI artifacts, now I have my answer :)

jdolitsky (approved):

PR #247, opened 2021-03-17
fix example for FileStore

### LINKED ISSUE #246: Failed to run the example with version 0.11.0
Here is the output of running `github.com/deislabs/oras/examples`:
```
Pushing hello.txt to localhost:5000/oras:test...
time="2021-03-17T23:50:21+08:00" level=warning msg="reference for unknown type: application/vnd.unknown.config.v1+json" digest="sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a" mediatype=application/vnd.unknown.config.v1+json size=2
time="2021-03-17T23:50:21+08:00" level=warning msg="reference for unknown type: my.custom.media.type" digest="sha256:03ba204e50d126e4674c005e04d82e84c21366780af1f43bd54a37816b6ab340" mediatype=my.custom.media.type size=13
Pushed to localhost:5000/oras:test with digest sha256:d3551d28503fc53dbb2f316bc71f3e0fd27772395c58ff32c35c9f137b70609a
Pulling from localhost:5000/oras:test and saving to hello.txt...
panic: no_name
```

It is a regression introduced by #230

deitch:
Fixes #246

### DIFF
--- a/examples/simple_push_pull.go
+++ b/examples/simple_push_pull.go
@@ -37,7 +37,7 @@ func main() {
 
 	// Pull file(s) from registry and save to disk
 	fmt.Printf("Pulling from %s and saving to %s...\n", ref, fileName)
-	fileStore := content.NewFileStore("")
+	fileStore := content.NewFileStore("", content.WithIgnoreNoName())
 	defer fileStore.Close()
 	allowedMediaTypes := []string{customMediaType}
 	desc, _, err = oras.Pull(ctx, resolver, ref, fileStore, oras.WithAllowedMediaTypes(allowedMediaTypes))


jdolitsky:
So as not to break people who may be using this -

Does it make sense to make the default `true`, and instead replace `WithIgnoreNoName` with `WithAllowNoName` ?

deitch:
I have no problem with reversing it, but didn't know if that would break anything, so I opted for not changing the default.

By default, if `FileStore` sees a descriptor without a proper title/name annotation, it returns an error. `WithIgnoreNoName` means, "don't error, just ignore it". If we make that the default, then, it would be `WithErrorOnNoName` rather than `WithAllowNoName`. 

I don't care much either way. I don't _think_ anyone is depending on `FileStore` returning an error when it hits a descriptor that is missing a name, but 🤷‍♂️

jdolitsky:
i guess, was the example not breaking with 0.10 ?

cc @shizhMSFT

shizhMSFT:
Tested against `v0.10.0`. The example works fine.

jdolitsky:
closing in favor of #250

PR #248, opened 2021-03-17
Build & release arm64 binaries

wmedlar:
This PR adds arm64 binaries for linux and darwin (Apple Silicon) to the release pipeline. Darwin arm64 is [supported as of golang 1.16](https://golang.org/doc/go1.16#darwin). Windows arm64, however, is not yet supported.

Additionally this updates a couple typos in the build docs.


shizhMSFT (approved):
LGTM

PR #263, opened 2021-04-28
initial allow for custom user agent

### LINKED ISSUE #240: Permit user-agent configuration
Currently ORAS uses containerd under the hood, which is setting the user agent to be the containerd default of `containerd/{version}` here - https://github.com/containerd/containerd/blob/38604a76ca9b1a7c3e00eeeb92d2d3975c76e5d1/remotes/docker/resolver.go#L144-L146

As a registry maintainer it'd be preferable to have ORAS extend or overwrite that user agent so that we can discern between different types of tooling.

Since ORAS can be used as a library, it'd also be fantastic if the end user could overwrite that value as well and have it propagate up, so that tooling which is wanting to use ORAS (i.e. [Homebrew](https://github.com/Homebrew/brew/pull/10835/files#diff-4890b590cb90a2c9ef5af6c48b020b952b712a1b080fa39be1c377de8e8a7c62R59-R64)) would also be able to set their own user agent to a more context-appropriate value.

Containerd allows for overriding the user agent value when [creating a Resolver](https://pkg.go.dev/github.com/containerd/containerd@v1.4.4/remotes/docker#NewResolver) by setting headers in the [`ResolverOptions`](https://pkg.go.dev/github.com/containerd/containerd@v1.4.4/remotes/docker#ResolverOptions).

Once the config is added, it should be able to be included in the docker resolver options here - https://github.com/deislabs/oras/blob/75801fef923e9d9945f512494f4790cdf1c022cd/cmd/oras/resolver.go#L18-L20

jdolitsky:
resolves #240

### DIFF
--- a/cmd/oras/resolver.go
+++ b/cmd/oras/resolver.go
@@ -39,7 +39,7 @@ func newResolver(username, password string, insecure bool, plainHTTP bool, confi
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "WARNING: Error loading auth file: %v\n", err)
 	}
-	resolver, err := cli.Resolver(context.Background(), client, plainHTTP)
+	resolver, err := cli.Resolver(context.Background(), client, plainHTTP, "")
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "WARNING: Error loading resolver: %v\n", err)
 		resolver = docker.NewResolver(opts)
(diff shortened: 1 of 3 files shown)


jdolitsky:
Some of this should change depending on how https://github.com/deislabs/oras/issues/176 is addressed

jdolitsky:
Closing. This can now be achieved using LoginWithOpts for authn/authz requests and ResolverWithOpts for normal requests

PR #257, opened 2021-04-07
Add OCI Conformance validation with ORAS

SteveLasker:
Add several steps to validate if a registry supports OCI Artifact types using `oras push|pull` 

Signed-off-by: Steve Lasker <stevenlasker@hotmail.com>

PR #265, opened 2021-05-04
Migrate to oras-go library

jdolitsky:
Git history has been migrated to the following new repositories:

- Go library: https://github.com/oras-project/oras-go
- Documentation/website: https://github.com/oras-project/oras-www

This PR removes all code/docs that have been migrated to the above repos, and leaves behind only code related to the ORAS CLI.

CI/testing has been affected since the unit tests and acceptance tests are now in oras-go. Instead, I've added "make build-linux" and "oras version" to GHitHub actions to make sure the code at least builds. We should soon add some acceptance tests in this repo to make sure the CLI operates as expected (possibly a replacement for #257 )


jdolitsky:
Note: I've also added commit c8ce7374f735434a8dd6acf8699e90d93039e924 to prep for a v0.12.0 release after this is merged

deitch (approved):
We should do this. Separating it out gives us a whole future path for doing things with the CLI at its own pace.

The lack of tests is a little disconcerting, but the reality is that we never had CLI tests anyways. Should we? Sure. Does this change it? Actually it makes it better, but adding a few very basic smoke tests.

We have an open question about the library name, which should be resolved first, then this should get merged in.

shizhMSFT (approved):
LGTM

PR #297, opened 2021-08-17
Oras discover with output tree format 

### LINKED ISSUE #295: ORAS Discover use -o for output format
The current `oras discover` api has a handy `--output-json` flag. Many other cli's use `-o json`, enabling different output formats.

Change requested:

change `--output-json` to `-o json`

mnltejaswini:
Add support for tree output to `oras discover` command. This output format traverses all the references and display them in a hierarchy tree structure. 

Fixes #295


mnltejaswini:
@SteveLasker  please review

cmd/oras/discover.go
```diff
@@ -17,12 +17,13 @@ import (
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/xlab/treeprint"
```
sajayantony, line 20: This is currently under MIT. Is that a concern or can we take dependency on this? 
/cc @lachie83 @SteveLasker

lachie83: The use of dependencies with permissive licenses in fine in the CNCF. It's copyleft licenses like MPL and GPL that you need to explicitly ask for an exception to be granted. Here is the doc from the CNCF stating that MIT is allowed - https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md#approved-licenses-for-allowlist

lachie83: Also - thanks for paying attention to this @sajayantony! It's not easy.

SteveLasker (approved):
The treeview experience in the CLI really helps visualize the graph. Thanks @mnltejaswini for the great experience.

PR #298, opened 2021-08-18
Adding script to download/install oras

juliusl:
Also, Setting the release version to "0.11.0-experimental"

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -19,6 +19,7 @@ type pullOptions struct {
 	allowedMediaTypes  []string
 	allowAllMediaTypes bool
 	allowEmptyName     bool
+	handleEmptyName    bool
 	keepOldFiles       bool
 	pathTraversal      bool
 	output             string
@@ -72,6 +73,7 @@ Example - Pull files with local cache:
 	cmd.Flags().StringArrayVarP(&opts.allowedMediaTypes, "media-type", "t", nil, "allowed media types to be pulled")
 	cmd.Flags().BoolVarP(&opts.allowAllMediaTypes, "allow-all", "a", false, "allow all media types to be pulled")
 	cmd.Flags().BoolVarP(&opts.allowEmptyName, "allow-empty-name", "", false, "allow pulling files with empty name")
+	cmd.Flags().BoolVarP(&opts.handleEmptyName, "handle-empty-name", "", false, "handle empty names, format will be {digest}.dat")
 	cmd.Flags().BoolVarP(&opts.keepOldFiles, "keep-old-files", "k", false, "do not replace existing files when pulling, treat them as errors")
 	cmd.Flags().BoolVarP(&opts.pathTraversal, "allow-path-traversal", "T", false, "allow storing files out of the output directory")
 	cmd.Flags().StringVarP(&opts.output, "output", "o", "", "output directory")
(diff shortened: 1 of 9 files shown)


juliusl:
@shizhMSFT  @sajayantony

internal/version/version.go
```diff
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version is the current version of the oras.
-	Version = "0.11.1"
+	Version = "0.11.0-experimental"
```
sajayantony, line 5: @SteveLasker should we rename this to alpha similar to distribution?

scripts/get-oras.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/bash
+
+# This script find the release by tag, downloads it to the user's home directory
```
northtyphoon, line 3: finds

internal/version/version.go
shizhMSFT: Should it be `0.11.1-experimental`?

scripts/get-oras.sh
```diff
@@ -0,0 +1,143 @@
...
+    log "${CYAN}$*${NC}"
+}
+
+# Setup logging for troubleshooting
+logsetup
+
+# Search Parameters
+REPO='oras-project/oras'
+TAG='v0.11.0' # TODO: This tag will need to change later
+OS='linux'    # TODO: Can add OS/Architecture detection later
```
shizhMSFT, line 40: How to override those environment variables?

juliusl: So the way to do this now is: 
`REPO='juliusl/oras' TAG='v0.11.5-alpha' ./get-oras.sh;. ./install-oras.sh`

or from the repo 

`curl https://raw.githubusercontent.com/juliusl/oras/juliusl/prototype-3/scripts/get-oras.sh | REPO='juliusl/oras' TAG='v0.11.5-alpha'  sh;. ./install-oras.sh`

juliusl:
@sajayantony  Could you give this a shot: 

curl https://raw.githubusercontent.com/juliusl/oras/juliusl/prototype-3/scripts/get-oras.sh | REPO='juliusl/oras' TAG='v0.11.5-alpha' sh;. ./install-oras.sh

cmd/oras/pull.go
```diff
@@ -19,6 +19,7 @@ type pullOptions struct {
 	allowedMediaTypes  []string
 	allowAllMediaTypes bool
 	allowEmptyName     bool
+	handleEmptyName    bool
```
sajayantony, line 22: I'm guessing these came in from another PR.

juliusl:
@sajayantony @northtyphoon do we not want this ?

sajayantony:
Given the move to just curling the binary and feedback from Lachie about
not piping shell maybe we should avoid it. But I know helm does have this
in the instructions so not sure why we would not take this in.

@stevelasker should we do this?

juliusl:
@sajayantony  Ok well if you guys want it, someone can copy and paste it from here: 

https://github.com/juliusl/azorasrc/tree/main/scripts

I have documentation. One thing I want to mention as well, is that this is actually a mini package manager as well, so you can have SxS oras binaries. 

Also, re: piping shell. For this script I do not require elevated permissions, nor do I touch anything in root. It's non destructive and adds files. That being said it comes from the internet so YMMV. (rustup also pipes shell btw)

PR #299, opened 2021-08-25
Adding release-oras.yml workflow

juliusl:
Copy/pasted from oras-project/distribution


sajayantony:
Can we make go-release just do all this without needing another config? 
https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/release/release.go#L93

sajayantony:
@juliusl - Does this look reasonable? https://github.com/oras-project/oras/pull/310

Issue #217, opened 2021-02-08
Enable configuring a certificate authority file for registry client.


absoludity:
Just as Helm allows as `--kube-ca-file` option to use a specific certificate authority file for the Kubernetes API Server connection, it would be great if ORAS enabled requests to the registry to be made with a specific certificate authority, so that connections to internal private registries are not susceptible to machine-in-the-middle.

Currently it appears the only option is to use the `--insecure` flag which I assume results in a TLS connection configured with [`InsecureSkipVerify`](https://golang.org/src/crypto/tls/common.go#L603). As per the docs there "This should be used only for testing or in combination with VerifyConnection or VerifyPeerCertificate".

souleb:
Hi,

I would like to work on this issue if possible.

Regards,
Soule

sajayantony:
Should we target using oras-go v2?

sabre1041:
@sajayantony I would recommend it be part of `v1` and `v2` since it is a capability that would be desired by consumers of each version

scottrigby:
👍 should be fixed in v2 first, then would need a backport to v1 for Helm users to take advantage.

PS, when there is a plan for the `oras-project/oras-go` stable v2.0.0 release, we should coordinate to also plan the Oras dependency upgrade in Helm.

scottrigby:
> 👍 should be fixed in v2 first, then would need a backport to v1 for Helm users to take advantage.
> 
> PS, when there is a plan for the `oras-project/oras-go` stable v2.0.0 release, we should coordinate to also plan the Oras dependency upgrade in Helm.

Update to this note. I was not aware that v2 has removed docker auth and implemented their own auth. Because of this v2 would have to implement things from scratch that v1 could do more easily using Docker client auth. So we may need to propose a working change for v1 to start which would then need to be ported to v2 (not backported from v2 to v1).

However, to keep with our agreement to ensure v1 and v2 have the opportunity to stay in sync in terms of features, we should in that same proposal give a recommendation on how the new v2 auth would want to implement the same thing. This way, we can get signoff from the v2 maintainers before approving/merging into v1. So we at least stay on the same page with features even if the code is completely different.

@sajayantony does this make sense to you?

PR #367, opened 2022-04-15
Upgrade login&logout commands to oras-go v2

### LINKED ISSUE #366: Upgrade login/logout commands to oras-go v2
Add login/logout packages for docker and general registries. Re-implement `login` and `logout` commands based on the new packages.

qweeah:
This PR 
- Upgrades login/logout commands based on oras-go v2
- Adds [config file](https://hackmd.io/A1AujCjMQaqnyQCsaMtEqQ?both#1-Docker-Config-Files) support on CLI side
- Adds [debug logs](https://hackmd.io/A1AujCjMQaqnyQCsaMtEqQ?both#5%EF%BC%89Debug-Tracing) for HTTP tracing

Once merged, will resolve #366


sajayantony:
@shizhMSFT @qweeah  - Is there a possibility to make this bite sized PRs where we move only part at a time so that it is incremental merge rather than one single big merge?

shizhMSFT:
@qweeah Looks like both `v0.4.0` and `v2.0.0` of `oras-go` can exist at the same time.

```
require (
	oras.land/oras-go v0.4.0
	oras.land/oras-go/v2 v2.0.0-20220415135518-c34895e747dc
)
```

Let's split this PR / issue into 3 parts:

1. Login / logout
2. Push
3. Pull

qweeah:
> @shizhMSFT @qweeah - Is there a possibility to make this bite sized PRs where we move only part at a time so that it is incremental merge rather than one single big merge?
> 
> For e.g. can we have the first one just say the pull path or just the copy path instead of 1 issue that would take multi month to merge?

I will check if the oras-go 0.4.0 and oras-go V2 can be imported at the same time. If it's possible, I will split this PR and also the update issue.

pkg/auth/client.go
```diff
@@ -0,0 +1,71 @@
...
+		Credential: func(ctx context.Context, reg string) (auth.Credential, error) {
+			return auth.Credential{
+				Username: settings.Username,
+				Password: settings.Secret,
+			}, nil
+		},
+		Client: &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: settings.Insecure,
```
qweeah, line 64: @sajayantony As @shizhMSFT mentioned, we can add some code here if customer need to use a specific certificate authority file #217

cmd/oras/login.go
```diff
@@ -60,20 +79,28 @@ Example - Login with insecure registry from command line:
 
 	cmd.Flags().BoolVarP(&opts.debug, "debug", "d", false, "debug mode")
 	cmd.Flags().StringArrayVarP(&opts.configs, "config", "c", nil, "auth config path")
+	cmd.Flags().StringVarP(&opts.credType, "cred-type", "t", "docker", "type of the saved credential")
 	cmd.Flags().StringVarP(&opts.username, "username", "u", "", "registry username")
 	cmd.Flags().StringVarP(&opts.password, "password", "p", "", "registry password or identity token")
 	cmd.Flags().BoolVarP(&opts.fromStdin, "password-stdin", "", false, "read password or identity token from stdin")
 	cmd.Flags().BoolVarP(&opts.insecure, "insecure", "k", false, "allow connections to SSL registry without certs")
+	cmd.Flags().BoolVarP(&opts.insecure, "plain-http", "", false, "allow connectsion to registry without SSL")
```
pjbgf, line 87: It would be great if the flag name was more meaningful. Flags that directly weakens existing security controls should name them, so my suggestion would be: `--insecure-allow-plain-http`.

qweeah: Added `allow` in the full parameter name and did some other corrections. Added insecure in the flag description to shorten the flag name.

shizhMSFT: Since this breaking change impact all commands, let's create an issue and address it separately.

/cc @pjbgf @qweeah

pkg/auth/docker/client.go
```diff
@@ -0,0 +1,166 @@
...
+	}, nil
+}
+
+func (c *Client) primaryCredentialsStore(hostname string) credentials.Store {
+	return c.configs[0].GetCredentialsStore(hostname)
+}
+
+// loadConfigFile reads the configuration files from the given path.
+func loadConfigFile(path string) (*configfile.ConfigFile, error) {
+	cfg := configfile.New(path)
```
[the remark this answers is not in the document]
qweeah: A new config will be created and credential will be stored into the files if auth passes.

qweeah:
Convert to draft, making changes to make it align to the design doc: https://hackmd.io/@Pnr77lOqTXm6moRbwyHlgg/r1I642QB5

internal/trace/context.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if debug {
+		log.SetLevel(logrus.DebugLevel)
+	} else if !verbose {
+		log.Out = ioutil.Discard
+	}
+
+	ctx = context.WithValue(ctx, loggerKey{}, log.WithContext(ctx))
+	return ctx
+}
```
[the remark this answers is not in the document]
shizhMSFT: Optionally, we can have 
```go
func WithLoggerLevel(ctx context.Context, level logrus.Level) (context.Context, logrus.FieldLogger)
```

internal/trace/context.go
```diff
@@ -0,0 +1,41 @@
...
+	"io/ioutil"
+
+	"github.com/sirupsen/logrus"
+)
+
+// loggerKey is the associated key type for logger entry in context.
+type loggerKey struct{}
+
+// ContextWithLogger returns a context with logrus log entry.
+func ContextWithLogger(ctx context.Context, verbose, debug bool) context.Context {
```
shizhMSFT, line 29: `verbose` and `debug` have their own meaning in general.

`debug` is toggled to print out all the debug information. Usually, they are debug-level logs. 
`verbose` is usually info level.

internal/trace/context.go
```diff
@@ -0,0 +1,41 @@
...
+	log := logrus.New()
+	ctx = context.WithValue(ctx, loggerKey{}, log.WithContext(ctx))
+
+	if debug {
+		log.SetLevel(logrus.DebugLevel)
+	} else if !verbose {
+		log.Out = ioutil.Discard
```
shizhMSFT, line 36: It will be better to set the right level instead of redirecting to `Discard`. The internal logic and perf implications are different.

internal/trace/context.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	"io/ioutil"
+
+	"github.com/sirupsen/logrus"
+)
+
+// loggerKey is the associated key type for logger entry in context.
+type loggerKey struct{}
```
shizhMSFT, line 26: A custom-typed int logger key would be better as we can have a constant key so that we don't need to create an object each time.

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+// Transport is an http.RoundTripper that keeps track of the in-flight
+// request and add hooks to report HTTP tracing events.
+type Transport struct {
+	base    http.RoundTripper
+	request *http.Request
+}
```
shizhMSFT, line 30: It can be simplified as 
```suggestion
type Transport struct {
	http.RoundTripper
}
```
And, the `request` field looks redundant to me.

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+	t.request = req
+	resp, err = t.base.RoundTrip(req)
+
+	ctx := req.Context()
+	e := ctx.Value(loggerKey{}).(*logrus.Entry)
+
+	e.Debugf(" Request URL: '%v'\n", req.URL)
+	e.Debugf(" Request method:'%v'\n", req.Method)
+	e.Debugf(" Request headers:\n")
+	logHeader(req.Header, e)
```
shizhMSFT, line 47: We may print the request before sending it out.

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if resp != nil {
```
shizhMSFT, line 49: It is a good practice to check `err` first, then check `resp`.

Since it is for tracing, we might need to log if error is not nil or the response is nil.

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return resp, err
+}
+
+// logHeader prints out the provided header keys and values, with auth header
+// scrubbed.
+func logHeader(header http.Header, e *logrus.Entry) {
+	if len(header) > 0 {
+		for k, v := range header {
+			if k == "Authorization" {
```
shizhMSFT, line 62: Header names are case insensitive. We might need to consider that.

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+				v = append(v[0:0], "*****")
```
shizhMSFT, line 63:
```suggestion
				v = []string{"*****"}
```

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+
+// RoundTrip calls base roundtrip while keeping track of the current request.
+func (t *Transport) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	t.request = req
+	resp, err = t.base.RoundTrip(req)
+
+	ctx := req.Context()
+	e := ctx.Value(loggerKey{}).(*logrus.Entry)
+
+	e.Debugf(" Request URL: '%v'\n", req.URL)
```
shizhMSFT, line 44: I am curios that do we really need to have `\n` in our logs? Should that be added automatically?

internal/trace/transport.go
```diff
@@ -0,0 +1,70 @@
...
+func logHeader(header http.Header, e *logrus.Entry) {
+	if len(header) > 0 {
+		for k, v := range header {
+			if k == "Authorization" {
+				v = append(v[0:0], "*****")
+			}
+			e.Debugf("   '%s': '%s'\n", k, strings.Join(v, ", "))
+		}
+	} else {
+		e.Debugf("   There is no header\n")
```
shizhMSFT, line 68: It would be better reworded to `Empty header`.

shizhMSFT (commented):
To gain more confidence on what we have, we need unit tests, at least for those internal packages.

internal/credential/store.go
```diff
@@ -0,0 +1,114 @@
...
+	authConf := types.AuthConfig{
+		Username:      cred.Username,
+		Password:      cred.Password,
+		ServerAddress: registry,
+		IdentityToken: cred.RefreshToken,
+		RegistryToken: cred.AccessToken,
+	}
+	for _, c := range s.configs {
+		return c.GetCredentialsStore(registry).Store(authConf)
+	}
```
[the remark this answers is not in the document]
qweeah: Is this by design? How about the Erase?

internal/trace/transport.go
Wwwsylvia, line 68: Just curious, is it a `tab` or a `space`?  There are a bunch of these, and it looks a bit confusing. Would something like `\t` help?

internal/credential/store.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, err
+	}
+
+	if !cfg.ContainsAuth() {
+		cfg.CredentialsStore = credentials.DetectDefaultStore(cfg.CredentialsStore)
+	}
+	return cfg, nil
+}
+
+// Store retrieves a credential for a given registry.
```
Wwwsylvia, line 74: retrieves? stores?

internal/credential/store.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		configs = append(configs, cfg)
+	}
+
+	return &Store{
+		configs: configs,
+	}, nil
+}
+
+// loadConfigFile reads the configuration files from the given path.
```
Wwwsylvia, line 51: Sorry I still don't get this. It reads configuration _files_ from the given path? Can we clarify what do we do if the path does not exist?

internal/credential/store.go
qweeah: Good catch

internal/trace/context.go
qweeah: I am following the suggestion in [context doc](https://pkg.go.dev/context#:~:text=should%20not%20be%20of%20type%20string%20or%20any%20other%20built%2Din%20type%20to%20avoid%20collisions%20between%20packages%20using%20context) for the key

internal/trace/transport.go
qweeah: It's space. I am reproducing the output format of `az acr` here.

internal/credential/store.go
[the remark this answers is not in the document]
qweeah: Will make the store and erase operation both only effect the first config file.

internal/credential/store.go
qweeah: 1) It only reads the configuration
- If the config file exists, then read from file
- If not, then create a in-memory object, which will be persistent later when store is called

2) why the code here is hard to understand
The [LoadFromReader](https://pkg.go.dev/github.com/docker/cli@v20.10.14+incompatible/cli/config/configfile#ConfigFile.LoadFromReader) function doesn't have a static counter part, so a new config always need to be initalized first, which make the if-else structure weird. I have re-structured the code, although the `cfg = configfile.New(path)` is called twice, the code is more understandable. Please check

cmd/oras/login.go
qweeah: reverted the chanage and created a issue for that

internal/credential/store.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, c := range s.configs {
+		authConf, _ := c.GetCredentialsStore(registry).Get(registry)
+		return auth.Credential{
+			Username:     authConf.Username,
+			Password:     authConf.Password,
+			AccessToken:  authConf.RegistryToken,
+			RefreshToken: authConf.IdentityToken,
+		}, nil
+
+	}
```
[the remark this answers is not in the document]
qweeah: >Also, it seems the error was omitted. Is this intentional?
Yes, The error will always be nil. 

I have changed the implementation to: iterate all the config and try to find a non-empty cred in a best-effort way.

internal/trace/context.go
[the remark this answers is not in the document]
qweeah: I changed the implementation to the second option

internal/trace/context.go
qweeah: Changed default log level to warning, which is info+1

shizhMSFT (approved):
LGTM

internal/trace/context.go
```diff
@@ -0,0 +1,44 @@
...
+
+type contextKey int
+
+// loggerKey is the associated key type for logger entry in context.
+const loggerKey contextKey = iota
+
+// WithLoggerLevel returns a context with logrus log entry.
+func WithLoggerLevel(ctx context.Context, level logrus.Level) (context.Context, logrus.FieldLogger) {
+	logger := logrus.New()
+	logger.SetFormatter(&logrus.TextFormatter{ForceColors: true})
```
shizhMSFT, line 32: Will this work if the output is piped to a file, not a TTY?

qweeah: This will work but contains unreadable characters. Changed this to disable quote to make log more readable.

shizhMSFT (approved):
LGTM

internal/credential/credential.go
```diff
@@ -0,0 +1,15 @@
+package credential
+
+import "oras.land/oras-go/v2/registry/remote/auth"
+
+func Credential(username, password string) auth.Credential {
+	if username == "" {
+		return auth.Credential{
+			RefreshToken: password,
```
binbin-li, line 8: I'm a bit confused here. Why do we have `password` as `token`?

internal/http/client.go
```diff
@@ -0,0 +1,46 @@
...
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras/internal/credential"
+	"oras.land/oras/internal/trace"
+	"oras.land/oras/internal/version"
+)
+
+// client option struct
+type ClientOptions struct {
+	Credential      auth.Credential
```
binbin-li, line 17: Just out of curiosify, could you help me explain the difference between the Credential and CredentialStore.

binbin-li, line 17: Just out of curiosify, could you help me explain the difference between the Credential and CredentialStore.

internal/trace/transport.go
```diff
@@ -0,0 +1,71 @@
...
+
+// logHeader prints out the provided header keys and values, with auth header
+// scrubbed.
+func logHeader(header http.Header, e logrus.FieldLogger) {
+	if len(header) > 0 {
+		for k, v := range header {
+			if strings.EqualFold(k, "Authorization") {
+				v = []string{"*****"}
+			}
+			e.Debugf("   %q: %q", k, strings.Join(v, ", "))
```
binbin-li, line 66: I noticed in `e.Debugf`, sometimes we have a leading space, but we have 3 spaces here. How do we determine the space number?

internal/credential/credential.go
qweeah: This is a by-design behavior: if user name is provided, then the password is consider as a refresh token, which is a authorization grant from resource owner in the OAuth model.

internal/trace/transport.go
qweeah: I am just copying the output format from Azure CLI to make output more readable.

qweeah: @binbin-li Do you have any idea on how to improve the hard-coded indent here?

internal/http/client.go
qweeah: 1) For code here, `Credential` and `CredentialStore` will be used for [auth function](https://github.com/oras-project/oras-go/blob/main/registry/remote/auth/client.go#L71) in oras-go client, please consider the scenarios:
 - if a username and password options are provided, then we can directly construct a `Credential` and pass it in
 - if username and password are not provided, then we need to find existed auth credentials in config files, which will be taken care of `CredentialStore`.

2) This code is already refactored, please check https://github.com/oras-project/oras/blob/main/cmd/oras/internal/option/remote.go#L108-L119, which is strictly aligned to the two scenarios,

PR #380, opened 2022-05-11
Support certificate authority configuration

qweeah:
This PR enables configuring certificate authority files for registry client.

For now this was only used in `login` command but will also be applied to `pull` and `push` commands after revamping.

#217

### DIFF
--- a/cmd/oras/login.go
+++ b/cmd/oras/login.go
@@ -44,7 +44,7 @@ type loginOptions struct {
 	username   string
 	password   string
 	insecure   bool
-	plainHttp  bool
+	plainHTTP  bool
 	verbose    bool
 }
 
@@ -87,7 +87,7 @@ Example - Login with insecure registry from command line:
 	cmd.Flags().BoolVarP(&opts.fromStdin, "password-stdin", "", false, "read password or identity token from stdin")
 	cmd.Flags().BoolVarP(&opts.insecure, "insecure", "k", false, "allow connections to SSL registry without certs")
 	cmd.Flags().StringVarP(&opts.caFilePath, "ca-file", "", "", "server certificate authority file for the remote registry")
-	cmd.Flags().BoolVarP(&opts.plainHttp, "plain-http", "", false, "allow insecure connections to registry without SSL")
+	cmd.Flags().BoolVarP(&opts.plainHTTP, "plain-http", "", false, "allow insecure connections to registry without SSL")
 	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", false, "verbose output")
 	return cmd
 }
@@ -147,7 +147,7 @@ func runLogin(opts loginOptions) (err error) {
 	if err != nil {
 		return err
 	}
-	remote.PlainHTTP = opts.plainHttp
+	remote.PlainHTTP = opts.plainHTTP
 	cred := credential.Credential(opts.username, opts.password)
 	rootCAs, err := http.LoadCertPool(opts.caFilePath)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/oras/login.go
```diff
@@ -85,6 +86,7 @@ Example - Login with insecure registry from command line:
 	cmd.Flags().StringVarP(&opts.password, "password", "p", "", "registry password or identity token")
 	cmd.Flags().BoolVarP(&opts.fromStdin, "password-stdin", "", false, "read password or identity token from stdin")
 	cmd.Flags().BoolVarP(&opts.insecure, "insecure", "k", false, "allow connections to SSL registry without certs")
+	cmd.Flags().StringVarP(&opts.caFilePath, "ca-file", "", "", "user specified certificate authority file for the registry target")
```
shizhMSFT, line 89:
```suggestion
	cmd.Flags().StringVarP(&opts.caFilePath, "ca-file", "", "", "server certificate authority file for the remote registry")
```

cmd/oras/login.go
```diff
@@ -38,13 +38,14 @@ type loginOptions struct {
...
-	plainHttp bool
-	verbose   bool
+	debug      bool
+	configs    []string
+	caFilePath string
+	username   string
+	password   string
+	insecure   bool
+	plainHttp  bool
+	verbose    bool
```
shizhMSFT, line 48: It seems that those options are common for all operations like `oras login/pull/push/copy/discover`. Could we refactor it to `commonOptions` in `common.go`?

cmd/oras/login.go
```diff
@@ -38,13 +38,14 @@ type loginOptions struct {
...
-	insecure  bool
-	plainHttp bool
-	verbose   bool
+	debug      bool
+	configs    []string
+	caFilePath string
+	username   string
+	password   string
+	insecure   bool
+	plainHttp  bool
```
shizhMSFT, line 47:
```suggestion
	plainHTTP  bool
```

internal/http/certificate.go
```diff
@@ -0,0 +1,22 @@
+package http
+
+import (
+	"crypto/x509"
+	"errors"
+	"os"
+)
+
+func LoadRootCAs(path string) (*x509.CertPool, error) {
```
shizhMSFT, line 9: From functionality wise, it is better to be named as `LoadCertPool()` as it can be used for intermediate certs.

internal/http/certificate.go
```diff
@@ -0,0 +1,22 @@
...
+	pool, err := x509.SystemCertPool()
```
shizhMSFT, line 10: The motivation of having user specified root CA pool is that the user cert is not in the system cert pool or the system cert pool is not trusted for this particular operation / scenario.

Therefore, we should start with a new cert pool instead.

cmd/oras/login.go
shizhMSFT: The refactor effort can be in another PR.

internal/http/certificate.go
```diff
@@ -0,0 +1,22 @@
...
+	if err != nil {
+		return nil, err
+	}
+	pemBytes, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	if ok := pool.AppendCertsFromPEM(pemBytes); !ok {
+		return nil, errors.New("Failed to add certificate authority in file: " + path)
```
shizhMSFT, line 19: The error message needs to be revised as this function will not be coupled with root CAs.

internal/http/client_test.go
```diff
@@ -54,3 +58,34 @@ func Test_NewClient_tlsConfig(t *testing.T) {
...
+		case p == "/v2/" && m == "GET":
+			w.WriteHeader(nhttp.StatusOK)
+		}
+	}))
+	defer ts.Close()
+
+	// Test CA pool
+	pool := x509.NewCertPool()
+	c := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: ts.Certificate().Raw})
+	pool.AppendCertsFromPEM(c)
```
shizhMSFT, line 77: This can be simplified by

```
pool.AddCert(ts.Certificate())
```

shizhMSFT (commented):
The PR needs to revisited and merged after migrating to oras-go v2 for oras pull and push as root CA options not only apply to login but also to pull and push.

cmd/oras/login.go
qweeah: Created an issue for it https://github.com/oras-project/oras/issues/385

shizhMSFT (approved):
LGTM for login. Still pending PRs for pull and push.

Issue #377, opened 2022-05-07
Refine release pipeline


shizhMSFT:
This issue to track the enhancement on auto releasing `oras` when a new version is cut.

The current [release pipeline](https://oras.land/cli/5_developer_guide/#release-checklist) requires lots of manual steps and personal credentials. Instead of personal credentials, we can leverage [GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) for releases. An example can be found in the [notation](https://github.com/notaryproject/notation/blob/main/.github/workflows/release-github.yml) repo. Similarly, GITHUB_TOKEN can be used to release packages to `ghcr.io` according [this guide](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio).

Before having a new release, it is better to revisit the following issues:
- #373
- #364 
- #251

Related issues:
- #256 

Successor issues:
- https://github.com/oras-project/oras-www/issues/10

qweeah:
Ack

PR #381, opened 2022-05-12
Refine release pipeline

qweeah:
Use Github token instead of personal access token for releasing.

Related to #377


qweeah:
Closing this since it's covered in #384

PR #393, opened 2022-05-23
Revamped push and pull commands based on orag-go v2

### LINKED ISSUE #147: pushing the same blob twice with a different name fails
Pushing the same blob, with different names to a repo that requires authorization only for pushing fails.
The reason is, that since the blob exists, only the manifest needs to be pushed.
The manifest is pushed in a flow that doesn't allow for retries. The request fails as 401, because it is the first request that needs auth, and not retried.

I wasn't sure what's the best approch to fix this in ORAS, will be happy to contribute a PR once we agree on a solution.

The error message received:
```
{"errors:" [{"code":"DENIED","message": "not authorized","detail": "not authorized"}]}
Error: failed commit on ref "manifest-sha256:cfdd09ec9b5584b26b9d21caf8abff7122fe29ea74f5bd92b48d927775876e3e": unexpected status: 401 Unauthorized
```

# Workaround:
Manually calling the root of the api and adding it to the authorizer before pusing. we implemented it here:
https://github.com/solo-io/wasme/blob/839c7818e16d715d4139393179004899a9d619c3/pkg/push/push.go#L114

# to reproduce:
```
docker run -dp 5001:5000 --restart=always --name registry registry

# proxy to require auth on put\post
envoy -c silly.yaml &

# use foo\foo as user password
docker login localhost:5000
echo hi > foo
oras push localhost:5000/hello:v1 foo
oras push localhost:5000/hello:v2 foo # <- this will fail with a 401
```

silly.yaml:
```
static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address: { address: 0.0.0.0, port_value: 5000 }
    filter_chains:
    - filters:
      - name: envoy.http_connection_manager
        config:
          stat_prefix: ingress_http
          route_config:
            name: local_route
            virtual_hosts:
            - name: namespace.local_service
              domains: ["*"]
              routes:
              # if it is POST or PUT and Authorizaiton header doesn't exist deny
              - match:
                  prefix: "/"
                  headers:
                  - name: "authorization"
                    exact_match: "Basic Zm9vOmJhcg=="
                route: { cluster: dockerregistry }
                # require auth for /v2/
              - match:
                  path: "/v2/"
                  headers:
                  - name: ":method"
                    exact_match: "GET"
                  - name: "authorization"
                    present_match: true
                    invert_match: true
                response_headers_to_add:
                - header:
                    key: www-authenticate
                    value: Basic realm="foo"
                direct_response:
                  status: 401
                  body: 
                    inline_string: not-authorized
                # allow get and head
              - match:
                  prefix: "/"
                  headers:
                  - name: ":method"
                    exact_match: "GET"
                route: { cluster: dockerregistry }
              - match:
                  prefix: "/"
                  headers:
                  - name: ":method"
                    exact_match: "HEAD"
                route: { cluster: dockerregistry }
              - match: { prefix: "/" }
                response_headers_to_add:
                - header:
                    key: www-authenticate
                    value: Basic realm="foo"
                direct_response:
                  status: 401
                  body: 
                    inline_string: not-authorized
                name: catchall
          http_filters:
          - name: envoy.router
  clusters:
  - name: dockerregistry
    connect_timeout: 0.25s
    type: STRICT_DNS
    lb_policy: ROUND_ROBIN
    hosts: [{ socket_address: { address: 127.0.0.1, port_value: 5001 }}]
```

For extra context, stack trace of the failing request (just before it is sent):
```
github.com/containerd/containerd/remotes/docker.dockerPusher.Push (/home/yuval/go/pkg/mod/github.com/containerd/containerd@v1.3.0/remotes/docker/pusher.go:225)
github.com/containerd/containerd/remotes/docker.(*dockerPusher).Push (Unknown Source:1)
github.com/containerd/containerd/remotes.push (/home/yuval/go/pkg/mod/github.com/containerd/containerd@v1.3.0/remotes/handlers.go:154)
github.com/containerd/containerd/remotes.PushHandler.func1 (/home/yuval/go/pkg/mod/github.com/containerd/containerd@v1.3.0/remotes/handlers.go:146)
github.com/containerd/containerd/remotes.PushContent (/home/yuval/go/pkg/mod/github.com/containerd/containerd@v1.3.0/remotes/handlers.go:216)
github.com/deislabs/oras/pkg/oras.Push (/home/yuval/go/pkg/mod/github.com/deislabs/oras@v0.7.0/pkg/oras/push.go:54)
github.com/solo-io/wasme/pkg/push.(*PusherImpl).Push (/home/yuval/Projects/solo/extend-envoy/pkg/push/push.go:115)
github.com/solo-io/wasme/pkg/cmd/push.runPush (/home/yuval/Projects/solo/extend-envoy/pkg/cmd/push/push.go:61)
github.com/solo-io/wasme/pkg/cmd/push.PushCmd.func1 (/home/yuval/Projects/solo/extend-envoy/pkg/cmd/push/push.go:45)
github.com/spf13/cobra.(*Command).execute (/home/yuval/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826)
github.com/spf13/cobra.(*Command).ExecuteC (/home/yuval/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914)
github.com/spf13/cobra.(*Command).Execute (/home/yuval/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864)
github.com/solo-io/wasme/pkg/cmd.Run (/home/yuval/Projects/solo/extend-envoy/pkg/cmd/cmd.go:49)
main.main (/home/yuval/Projects/solo/extend-envoy/main.go:16)
runtime.main (/home/yuval/bin/go/src/runtime/proc.go:203)
runtime.goexit (/home/yuval/bin/go/src/runtime/asm_amd64.s:1357)
```

### LINKED ISSUE #172: finalizer on Pusher returned by Resolver?
Looking at how `oras.Push()` is called [here](https://github.com/deislabs/oras/blob/master/pkg/oras/push.go#L17), we pass it a `remotes.Resolver`:

```go
func Push(ctx context.Context, resolver remotes.Resolver, ref string, provider content.Provider, descriptors []ocispec.Descriptor, opts ...PushOpt) (ocispec.Descriptor, error) {
```

`Push()` , in turn, gets a pusher by calling `resolver.Pusher()` [here](https://github.com/deislabs/oras/blob/master/pkg/oras/push.go#L35)

```go
	pusher, err := resolver.Pusher(ctx, ref)
```

and then passing that pusher to `remotes.PushContent` [here](https://github.com/deislabs/oras/blob/master/pkg/oras/push.go#L52):

```go
	if err := remotes.PushContent(ctx, pusher, desc, store, nil, wrapper); err != nil {
```

What if my pusher (returned by `resolver.Pusher()`) needs some finalizing done? There are `Commit()` and `Close()` calls on the `content.Writer` returned by _each_ `Push()`, but not overall. I might have a custom resolver (and therefore pusher) that needs to do some activity at the end of a whole `PushContent`.

`remotes.PushContent()` has an option for doing so via its handlers, which can be both pre- and post-, but `oras.Push()` only accepts pre-push handlers for each push, see [here](https://godoc.org/github.com/deislabs/oras/pkg/oras#WithPushBaseHandler). In any case, those are called with each push, and so no way to know, "this is the final push".

I can do something at the `Resolver` level, rather than `Pusher`, and thus have control after calling `oras.Push()`, but that is one level too high for it, and could bleed into other things. For example, I should be able to push 3 images up, each time calling `oras.Push()` (and hence `resolver.Pusher()`) from the same instance of `resolver`; finalizing would really mess that up.

### LINKED ISSUE #217: Enable configuring a certificate authority file for registry client.

### LINKED ISSUE #392: If a blob already exists before push, status tracker should output the information

qweeah:
This PR upgrades `push` and `pull` commands towards oras-go v2.

Resolve #172,  resolves #217, resolves #147, resolves #392.


qweeah:
Will split this into two PRs, for push and pull separately.

PR #394, opened 2022-05-23
Refactor push command

### LINKED ISSUE #147: pushing the same blob twice with a different name fails

### LINKED ISSUE #172: finalizer on Pusher returned by Resolver?

### LINKED ISSUE #392: If a blob already exists before push, status tracker should output the information

qweeah:
This PR upgrades push command towards oras-go v2.

Resolves #172.
Resolves #147.
Resolves #392.


internal/status/tracker.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+)
+
+type StatusTracker struct {
+	oras.Target
+	out          io.Writer
+	printLock    sync.Mutex
+	printAfter   bool
```
binbin-li, line 19: Just out of curiosity, what does `printAfter` mean?

internal/status/tracker.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	}
+
+	print()
+	return t.Target.Push(ctx, expected, content)
+}
+
+func (t *StatusTracker) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+	existed, err := t.Target.Exists(ctx, target)
+	if t.printExisted && err == nil && existed {
```
binbin-li, line 71: Wonder if it's possible that `t.printExisted` is false but `existed` is true. What's the expected behavior in this case?

qweeah: No output for the existence of the blob.

internal/status/tracker.go
qweeah: Print before(false) or after(true) the push.

cmd/oras/push.go
```diff
@@ -63,76 +72,62 @@ Example - Push file to the HTTP registry:
...
+
+	// Prepare manifest
+	store := file.New("")
+	defer store.Close()
+	store.AllowPathTraversalOnWrite = opts.PathValidationDisabled
+
+	// Ready to push
+	tracker := status.NewPushTracker(dst, opts.Verbose)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
shizhMSFT, line 121: What's the purpose of having a cancellable context? I'm not sure about the intention here.

cmd/oras/push.go
```diff
@@ -63,76 +72,62 @@ Example - Push file to the HTTP registry:
...
-		file, err := store.Add(annotationConfig, mediaType, filename)
-		if err != nil {
+	reg, err := opts.NewRegistry(ref.Registry, opts.Common)
+	if err != nil {
+		return err
+	}
+	dst, err := reg.Repository(ctx, ref.Repository)
+	if err != nil {
+		return err
+	}
```
shizhMSFT, line 103: Can we have this block of code refactored?

Note that we can directly call `remote.NewRepository(opts.targetRef)` internally, instead of calling `reg := remote.NewRegistry()` and then `repo := reg.Repository()`. Also, keeping the `*remote.Repository` type is better for future `oras discover` development.

cmd/oras/push.go
```diff
@@ -181,5 +176,35 @@ func loadFiles(store *content.FileStore, annotations map[string]map[string]strin
...
 	return files, nil
 }
+
+func packManifest(ctx context.Context, store *file.Store, annotations map[string]map[string]string, opts *pushOptions) (ocispec.Descriptor, error) {
+	var packOpts oras.PackOptions
+	packOpts.ConfigAnnotations = annotations[annotationConfig]
+	packOpts.ManifestAnnotations = annotations[annotationManifest]
+	if opts.ManifestConfigRef != "" {
+		filename, mediaType := parseFileRef(opts.ManifestConfigRef, ocispec.MediaTypeImageConfig)
```
shizhMSFT, line 190: Originally, `artifact.UnknownConfigMediaType` is used for the default config type. However, it is replaced by `ocispec.MediaTypeImageConfig`. Is it intended?

internal/status/tracker.go
shizhMSFT: Note: This `StatusTracker` is a workaround. Implementation will be enhanced once https://github.com/oras-project/oras-go/pull/150 is merged.

internal/status/tracker.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"sync"
+
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+)
+
+// StatusTracker is used to track status when interacting with the target CAS.
+type StatusTracker struct {
```
shizhMSFT, line 31: nit: Better naming `Tracker` as `status.Tracker` is better than `status.StatusTracker`.

internal/status/tracker.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+			name = expected.MediaType
+		}
+
+		digestString := expected.Digest.String()
+		if err := expected.Digest.Validate(); err == nil {
+			if algo := expected.Digest.Algorithm(); algo == digest.SHA256 {
+				digestString = expected.Digest.Encoded()[:12]
+			}
+		}
```
shizhMSFT, line 71: This block of code is to generate digest string safely. We may refactor this block of code into a function so that it can be reused in `Exists()`.

cmd/oras/internal/option/push.go
```diff
@@ -0,0 +1,32 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package option
+
+import "github.com/spf13/pflag"
+
+// Push option struct.
+type Push struct {
```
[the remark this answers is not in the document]
qweeah: Yes, as well as some functions related to the common options. Are you suggesting that options here are not common so don't need a separate file?

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #395, opened 2022-05-23
Refactor pull command

### LINKED ISSUE #325: Error: not_found downloading the manifest-config from a remote artifact
I am trying to download the manifest config from a remote container image as seen [here](https://github.com/oras-project/oras/pull/274) using the following command:

```
oras pull --manifest-config config.json:application/vnd.docker.distribution.manifest.v2+json OCI_REFERENCE
```

The config.json file is downloaded properly with the correct data, however, the cli is returning `Error: not_found` and exiting with error code 1.

Sorry if this issue is already duplicated. I tried looking similar ones but didn't fount it.

qweeah:
This PR upgrades pull command towards oras-go v2.

Resolves #325


cmd/oras/pull.go
```diff
@@ -1,38 +1,47 @@
...
+		option.Remote
+
+		targetRef         string
+		cacheRoot         string
+		KeepOldFiles      bool
+		PathTraversal     bool
+		Output            string
+		ManifestConfigRef string
+	}
+)
```
shizhMSFT, line 44:
```suggestion
type pullOptions struct {
	option.Common
	option.Remote

	targetRef         string
	cacheRoot         string
	keepOldFiles      bool
	pathTraversal     bool
	output            string
	manifestConfigRef string
}
```

cmd/oras/pull.go
```diff
@@ -61,64 +70,77 @@ Example - Pull files with local cache:
...
+	if err != nil {
+		return err
+	}
+
+	dir := opts.Output
+	if dir == "" {
+		dir, err = os.Getwd()
+		if err != nil {
+			return err
+		}
```
shizhMSFT, line 104: Curious if `dir = "."` works or not.

cmd/oras/pull.go
```diff
@@ -61,64 +70,77 @@ Example - Pull files with local cache:
...
 	}
-	if opts.manifestConfigRef != "" {
-		pullOpts = appendPullManifestConfigHandlers(pullOpts, opts.manifestConfigRef)
+	desc, err := oras.Copy(ctx, src, ref.Reference, tracker, ref.Reference)
+	if err != nil {
+		return err
 	}
-
-	desc, artifacts, err := oras.Pull(ctx, resolver, opts.targetRef, store, pullOpts...)
+	artifacts, err := content.DownEdges(ctx, src, desc)
```
shizhMSFT, line 139: All named successors are "artifacts", not just children.

cmd/oras/pull.go
```diff
@@ -61,64 +70,77 @@ Example - Pull files with local cache:
...
-		cachedStore, err := newStoreWithCache(store, opts.cacheRoot)
+		cache, err := oci.New(opts.cacheRoot)
 		if err != nil {
 			return err
 		}
-		pullOpts = append(pullOpts, oras.WithContentProvideIngester(cachedStore))
+		if _, err = oras.Copy(ctx, src, ref.Reference, cache, ref.ReferenceOrDefault()); err != nil {
+			return err
+		}
+		src = cache
```
shizhMSFT, line 133: This way of caching has an issue that the progress is not shown in the real time.

cmd/oras/pull.go
qweeah: It works. Changed accordingly.

cmd/oras/pull.go
qweeah: Updated to do the caching per-reference.

cmd/oras/pull.go
qweeah: Updated to detect empty artifact during copying.

cmd/oras/internal/option/remote.go
```diff
@@ -144,8 +145,12 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
...
 		return nil, err
 	}
-
-	repo.PlainHTTP = opts.PlainHTTP
+	switch host, _, _ := net.SplitHostPort(repo.Reference.Registry); host {
+	case "localhost":
+		repo.PlainHTTP = true
+	default:
+		repo.PlainHTTP = opts.PlainHTTP
+	}
```
shizhMSFT, line 153: `NewRegistry()` also requires the same logic for `localhost`. Recommend to refactor.

cmd/oras/internal/option/remote.go
```diff
@@ -144,8 +145,12 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
 	if err != nil {
 		return nil, err
 	}
-
-	repo.PlainHTTP = opts.PlainHTTP
+	switch host, _, _ := net.SplitHostPort(repo.Reference.Registry); host {
```
shizhMSFT, line 148: `host` will be empty if `repo.Reference.Registry` is `localhost` since `net.SplitHostPort` will return error in this case.

cmd/oras/pull.go
```diff
@@ -61,106 +66,60 @@ Example - Pull files with local cache:
...
+		cache, err = oci.New(opts.cacheRoot)
 		if err != nil {
 			return err
 		}
-		pullOpts = append(pullOpts, oras.WithContentProvideIngester(cachedStore))
-	}
-	if opts.manifestConfigRef != "" {
-		pullOpts = appendPullManifestConfigHandlers(pullOpts, opts.manifestConfigRef)
 	}
+	tracker := status.NewPullTracker(dstStore, mco, cache)
```
shizhMSFT, line 111: The cache implementation should be decoupled with the status tracker.

A suggested way of implementation is to create a cache of type  `oras.Target` by wrapping the destination `oras.Target`.

cmd/oras/pull.go
```diff
@@ -61,106 +66,60 @@ Example - Pull files with local cache:
...
-		}
 		return err
 	}
-	if len(artifacts) == 0 {
+	if tracker.PulledEmpty() {
 		fmt.Println("Downloaded empty artifact")
+	} else {
+		fmt.Println("Pulled", opts.targetRef)
+		fmt.Println("Digest:", desc.Digest)
 	}
```
shizhMSFT, line 123: You may want to remove the `else` keyword as those lines should always be printed.

internal/status/pull.go
```diff
@@ -0,0 +1,112 @@
...
+	cache       oras.Target
+	pulledEmpty bool
+}
+
+// NewPullTracker returns a new status tracking object for pull command.
+func NewPullTracker(target oras.Target, option *ManifestConfigOption, cache oras.Target) *PullTracker {
+	return &PullTracker{
+		Target:               target,
+		out:                  os.Stdout,
+		prompt:               "Downloaded",
```
shizhMSFT, line 48: `prompt` seems redundant.

internal/status/push.go
```diff
@@ -27,35 +27,34 @@ import (
...
-	return &Tracker{
+// NewPushTracker returns a new status tracking object for push command.
+func NewPushTracker(target oras.Target, verbose bool) *PushTracker {
+	return &PushTracker{
 		Target:       target,
 		out:          os.Stdout,
 		prompt:       "Uploading",
 		verbose:      verbose,
-		printAfter:   true,
 		printExisted: true,
```
shizhMSFT, line 49: This option seems redundant.

internal/status/push.go
```diff
@@ -64,38 +63,33 @@ func (t *Tracker) Push(ctx context.Context, expected ocispec.Descriptor, content
...
-		}
-		print()
-		return nil
+	if err := t.Target.Push(ctx, expected, content); err != nil {
+		return err
 	}
-
 	print()
-	return t.Target.Push(ctx, expected, content)
+	return nil
```
shizhMSFT, line 75: For push, we need to print then push.

qweeah: Changed the print to be done before the pushing.

qweeah:
@shizhMSFT I have refactored the pull caching. It borrows the [cas proxy implementation](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go) in oras-go, can we just make that package public?

shizhMSFT:
> @shizhMSFT I have refactored the pull caching. It borrows the [cas proxy implementation](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go) in oras-go, can we just make that package public?

/cc @Wwwsylvia for inputs

internal/status/push.go
```diff
@@ -85,17 +73,18 @@ func (t *Tracker) Push(ctx context.Context, expected ocispec.Descriptor, content
...
 // option, see https://github.com/oras-project/oras-go/issues/59.
-func (t *Tracker) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+func (t *PushTracker) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
 	existed, err := t.Target.Exists(ctx, target)
+
 	if t.printExisted && err == nil && existed {
 		t.printLock.Lock()
 		defer t.printLock.Unlock()
-		fmt.Fprintln(t.out, digestString(target), "already exists")
+		fmt.Fprintln(t.out, "Existed  ", digestString(target), target.Annotations[ocispec.AnnotationTitle])
```
Wwwsylvia, line 82: > "Existed "
nit: Do we need the space here?

internal/status/pull.go
```diff
@@ -0,0 +1,95 @@
...
+	return &PullTracker{
+		Target:               target,
+		ManifestConfigOption: option,
+		out:                  os.Stdout,
+		verbose:              false,
+		pulledEmpty:          true,
+	}
+}
+
+// ManifestConfigOption contains options for manifest config.
```
Wwwsylvia, line 52: What is "manifest config"?

cmd/oras/pull.go
```diff
@@ -61,106 +67,64 @@ Example - Pull files with local cache:
...
 		}
-		pullOpts = append(pullOpts, oras.WithContentProvideIngester(cachedStore))
+		src = cache.New(repo, ociStore)
 	}
-	if opts.manifestConfigRef != "" {
-		pullOpts = appendPullManifestConfigHandlers(pullOpts, opts.manifestConfigRef)
+
+	// Copy Options
+	ctx, _ := opts.SetLoggerLevel()
+	var dstStore = file.New(opts.Output)
```
Wwwsylvia, line 105: nit: why not `dstStore := file.New(opts.Output)`?

Wwwsylvia:
> > @shizhMSFT I have refactored the pull caching. It borrows the [cas proxy implementation](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go) in oras-go, can we just make that package public?
> 
> /cc @Wwwsylvia for inputs

Agreed. Since it's common for clients to implement caching, I think we can promote the `proxy.go` to the `content` package.

internal/status/push.go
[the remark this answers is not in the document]
qweeah: Yes, it is to align `Existed` with the width of `Uploading`

internal/status/pull.go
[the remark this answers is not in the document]
qweeah: Could be the configuration object of a [container ](https://github.com/opencontainers/image-spec/blob/main/config.md)or artifact. Are you suggesting we can use another term here?

internal/status/pull.go
```diff
@@ -0,0 +1,95 @@
...
+	return t.pulledEmpty
+}
+
+// Push pushes the content, matching the expected descriptor with status
+// tracking.
+// Current implementation is a workaround before oras-go v2 supports copy
+// option, see https://github.com/oras-project/oras-go/issues/59.
+func (t *PullTracker) Push(ctx context.Context, expected ocispec.Descriptor, content io.Reader) error {
+	option := t.ManifestConfigOption
+	if option != nil && option.MediaType == expected.MediaType && option.Name != "" {
```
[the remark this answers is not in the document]
qweeah: Background: The manifest config option is set through the `--manifest-config` flag, via which user can specify a file name and a media type to download the target config. If the media type is not provided, `application/vnd.unknown.config.v1+json` will be used.

So here we should **only rewrite the file name when a blob matches the target config media type**. If the media type doesn't match, then current blob is not the target config blob.

internal/status/push.go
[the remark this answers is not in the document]
Wwwsylvia: BTW, do we need PM to review these?

internal/status/pull.go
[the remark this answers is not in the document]
Wwwsylvia: OK interesting. Maybe we can add some comments here?

sajayantony:
> public



> @shizhMSFT I have refactored the pull caching. It borrows the [cas proxy implementation](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go) in oras-go, can we just make that package public?

Are we planning to support it and do die Diligence?

qweeah:
>

qweeah:
Closing since the copy options are supported in oras-go and no need to hack status tracking.

PR #412, opened 2022-06-22
Add manifest config support

### LINKED ISSUE #409: Support pulling manifest config
User cannot pull manifest config and `--manifest-config` option is disabled for `pull` command.

Should add the option back after oras-go support cutomized down-edge filter option in oras.Copy oras-project/oras-go#189.

We should fix this before v0.13.0 release

qweeah:
This PR resolves #409.

Signed-off-by: Billy Zha <qweeah@gmail.com>


cmd/oras/pull.go
```diff
@@ -103,17 +104,47 @@ func runPull(opts pullOptions) error {
...
+				if s.Annotations == nil {
+					s.Annotations = make(map[string]string)
+				}
+				s.Annotations[ocispec.AnnotationTitle] = configFileName
+			}
+
+			if s.Annotations[ocispec.AnnotationTitle] == "" {
+				if opts.Verbose {
+					// Display skipped
+					if err = display.Print("Skipped   ", display.ShortDigest(desc), desc.MediaType); err != nil {
```
qweeah, line 125: Since those descriptors are not fetched, we can no longer print `downloaded`. @shizhMSFT How about `Skipped`

qweeah:
Made draft since the oras-go change is not merged yet.

cmd/oras/pull.go
```diff
@@ -103,17 +104,38 @@ func runPull(opts pullOptions) error {
...
+			if s.MediaType == configMediaType {
+				// Add annotation for manifest config
+				if s.Annotations == nil {
+					s.Annotations = make(map[string]string)
+				}
+				s.Annotations[ocispec.AnnotationTitle] = configFileName
+			}
+			if s.Annotations[ocispec.AnnotationTitle] != "" {
+				// Only copy named blobs
+				ret = append(ret, s)
```
qweeah, line 125: @shizhMSFT Do we need to print digest and media type for skipped nodes in verbose mode?

cmd/oras/pull.go
```diff
@@ -103,17 +104,38 @@ func runPull(opts pullOptions) error {
...
 	copyOptions := oras.DefaultCopyOptions
+	configFileName, configMediaType := parseFileRef(opts.ManifestConfigRef, oras.MediaTypeUnknownConfig)
+	copyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+		successors, err := content.Successors(ctx, fetcher, desc)
+		if err != nil {
+			return nil, err
+		}
+		ret := []ocispec.Descriptor{}
+		for _, s := range successors {
+
```
shizhMSFT, line 115: nit: remove empty line.

cmd/oras/pull.go
```diff
@@ -103,17 +104,38 @@ func runPull(opts pullOptions) error {
 
 	// Copy Options
 	copyOptions := oras.DefaultCopyOptions
+	configFileName, configMediaType := parseFileRef(opts.ManifestConfigRef, oras.MediaTypeUnknownConfig)
+	copyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+		successors, err := content.Successors(ctx, fetcher, desc)
+		if err != nil {
+			return nil, err
+		}
+		ret := []ocispec.Descriptor{}
```
shizhMSFT, line 113: See https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices
```suggestion
		var ret []ocispec.Descriptor
```

cmd/oras/pull.go
```diff
@@ -103,17 +104,38 @@ func runPull(opts pullOptions) error {
...
+				// Add annotation for manifest config
+				if s.Annotations == nil {
+					s.Annotations = make(map[string]string)
+				}
+				s.Annotations[ocispec.AnnotationTitle] = configFileName
+			}
+			if s.Annotations[ocispec.AnnotationTitle] != "" {
+				// Only copy named blobs
+				ret = append(ret, s)
+			}
```
shizhMSFT, line 126: This logic is problematic. What if the current descriptor is a manifest list / index?

cmd/oras/pull.go
shizhMSFT: @qweeah Wait a second. We don't even print out the digest and media type for unamed nodes, do we? Let's not do this in this PR. We should create an issue instead.

cmd/oras/pull.go
qweeah: I was wondering if nil slice works for append, tested and looks good.

cmd/oras/pull.go
qweeah: Looks like we still need a function to tell if a media can have down edge on oras-go side?

qweeah: Added a oras-project/oras-go#193 for the manifest checking on oras-go side and applied it here.

There is a corner case we cannot fix: if there is an oras reference manifest and it's `Subject` is nil, we will still fetch its content even it may be unnamed and has no successor.

Anyway, this cost cannot be avoided since the `Subject` cannot be obtained without fetching.

cmd/oras/pull.go
qweeah: Created issue #415

shizhMSFT (approved):
LGTM

Issue #379, opened 2022-05-07
Merge features in `artifacts` branch as experimental features


shizhMSFT:
With `oras-go` v2, we can "merge" (re-implement) those features in the `artifacts` branch into `main` with those features marked as experimental.

Features:
- [x] `oras discover`.
- [x] `oras push` artifact manifest.
- [x] `oras pull` artifact manifest.

shizhMSFT:
@qweeah You may create issues according to this issue.

qweeah:
ack

qweeah:
# Implementation Gaps
## Pull
This command is already provided in the main branch. Below options are new in the `artifacts` branch:
- `media-type` and `allow-all`: Since we have already removed the metia-type filtering features in https://github.com/oras-project/oras/issues/130, those options are deprecated.
- `allow-empty-name`: File target already support in-memory caching for non-named manifests.

Basically there is no implementation effort required for the pull command.

## Discover
This is a new command, we need to support discovering artifacts linked with a specified reference, with optional artifact type filtering and different formats of output.

## Push
This command is already provided in the main branch. Below options are new in the `artifacts` branch:
- `manifestExport`: User will use this to save the packed manifest to a local file. We will refetch the manifest from the source file target and write to a file.  It would be better 

- `subject` and `artifact-type`: those two options will be removed and re-implemented in #424

shizhMSFT:
Closed by #419 and #425

PR #419, opened 2022-06-27
Add discover comand

### LINKED ISSUE #382: Discover sub-command is not using the latest ORAS Manifest Referrers API spec
The [current logic](https://github.com/oras-project/oras/blob/58fa344544b34571012d91f523cea54bf364549b/pkg/remotes/docker/discoverer.go#L119) still uses the old Manifest Referrers API spec. The API signature has been updated [here](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md).

qweeah:
This PR adds a new discover command, which can be used for searching referrer artifacts.

Resolves #382, related to #379

### DIFF
--- a/cmd/oras/discover.go
+++ b/cmd/oras/discover.go
@@ -1,0 +1,173 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras/cmd/oras/internal/option"
+
+	"github.com/containerd/containerd/reference"
+	"github.com/need-being/go-tree"
+	digest "github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
+	"github.com/spf13/cobra"
+)
+
+type discoverOptions struct {
+	option.Common
+	option.Remote
+
+	targetRef    string
+	artifactType string
+	outputType   string
+}
+
+func discoverCmd() *cobra.Command {
+	var opts discoverOptions
+	cmd := &cobra.Command{
+		Use:   "discover [options] <name:tag|name@digest>",
+		Short: "discover artifacts from remote registry",
+		Long: `discover artifacts from remote registry
+
+Example - Discover artifacts of type "" linked with the specified reference:
+  oras discover --artifact-type application/vnd.cncf.notary.v2 localhost:5000/hello:latest
+`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			return runDiscover(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
+	cmd.Flags().StringVarP(&opts.outputType, "output", "o", "table", fmt.Sprintf("Format in which to display references (%s, %s, or %s). tree format will show all references including nested", "table", "json", "tree"))
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func runDiscover(opts discoverOptions) error {
+	ctx, _ := opts.SetLoggerLevel()
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	// discover artifacts
+	root := tree.New(opts.targetRef)
+	desc, err := repo.Resolve(ctx, repo.Reference.Reference)
+	if err != nil {
+		return err
+	}
+	desc, refs, err := getAllReferences(ctx, repo, desc, opts.artifactType, root, opts.outputType == "tree")
+	if err != nil {
+		return err
+	}
+
+	switch opts.outputType {
+	case "tree":
+		tree.Print(root)
+	case "json":
+		printDiscoveredReferencesJSON(desc, *refs)
+	default:
+		fmt.Println("Discovered", len(*refs), "artifacts referencing", opts.targetRef)
+		fmt.Println("Digest:", desc.Digest)
+
+		if len(*refs) != 0 {
+			fmt.Println()
+			printDiscoveredReferencesTable(*refs, opts.Verbose)
+		}
+	}
+	return nil
+}
+
+func getAllReferences(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string, node *tree.Node, queryGraph bool) (ocispec.Descriptor, *[]artifactspec.Descriptor, error) {
+	var results []artifactspec.Descriptor
+	err := repo.Referrers(ctx, desc, func(referrers []artifactspec.Descriptor) error {
+		for _, r := range referrers {
+			if artifactType == "" || artifactType == r.ArtifactType {
+				results = append(results, r)
+				if queryGraph {
+					// Find all referrers
+					referrerNode := node.AddPath(r.ArtifactType, r.Digest)
+					_, nestedReferrers, err := getAllReferences(
+						ctx, repo,
+						ocispec.Descriptor{
+							Digest:    r.Digest,
+							Size:      r.Size,
+							MediaType: r.MediaType,
+						},
+						artifactType, referrerNode, queryGraph)
+					if err != nil {
+						return err
+					}
+					results = append(results, *nestedReferrers...)
+				}
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		if err == reference.ErrObjectRequired {
+			return ocispec.Descriptor{}, nil, fmt.Errorf("image reference format is invalid. Please specify <name:tag|name@digest>")
+		}
+		return ocispec.Descriptor{}, nil, err
+	}
+	return desc, &results, nil
+}
+
+func printDiscoveredReferencesTable(refs []artifactspec.Descriptor, verbose bool) {
+	typeNameTitle := "Artifact Type"
+	typeNameLength := len(typeNameTitle)
+	for _, ref := range refs {
+		if length := len(ref.ArtifactType); length > typeNameLength {
+			typeNameLength = length
+		}
+	}
+
+	print := func(key string, value interface{}) {
+		fmt.Println(key, strings.Repeat(" ", typeNameLength-len(key)+1), value)
+	}
+
+	print(typeNameTitle, "Digest")
+	for _, ref := range refs {
+		print(ref.ArtifactType, ref.Digest)
+		if verbose {
+			printJSON(ref)
+		}
+	}
+}
+
+func printDiscoveredReferencesJSON(desc ocispec.Descriptor, refs []artifactspec.Descriptor) {
+	type reference struct {
+		Digest   digest.Digest `json:"digest"`
+		Artifact string        `json:"artifactType"`
+	}
+	output := struct {
+		Digest     digest.Digest `json:"digest"`
+		References []reference   `json:"references"`
+	}{
+		Digest:     desc.Digest,
+		References: make([]reference, len(refs)),
+	}
+
+	for i, ref := range refs {
+		output.References[i] = reference{
+			Digest:   ref.Digest,
+			Artifact: ref.ArtifactType,
+		}
+	}
+
+	printJSON(output)
+}
+
+func printJSON(object interface{}) {
+	encoder := json.NewEncoder(os.Stdout)
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	encoder.Encode(object)
+}
--- a/cmd/oras/discover.go
+++ b/cmd/oras/discover.go
@@ -10,7 +10,6 @@ import (
 	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras/cmd/oras/internal/option"
 
-	"github.com/containerd/containerd/reference"
 	"github.com/need-being/go-tree"
 	digest "github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -112,9 +111,6 @@ func getAllReferences(ctx context.Context, repo *remote.Repository, desc ocispec
 		return nil
 	})
 	if err != nil {
-		if err == reference.ErrObjectRequired {
-			return ocispec.Descriptor{}, nil, fmt.Errorf("image reference format is invalid. Please specify <name:tag|name@digest>")
-		}
 		return ocispec.Descriptor{}, nil, err
 	}
 	return desc, &results, nil
--- a/cmd/oras/main.go
+++ b/cmd/oras/main.go
@@ -11,7 +11,7 @@ func main() {
 		Use:          "oras [command]",
 		SilenceUsage: true,
 	}
-	cmd.AddCommand(pullCmd(), pushCmd(), loginCmd(), logoutCmd(), versionCmd())
+	cmd.AddCommand(pullCmd(), pushCmd(), loginCmd(), logoutCmd(), versionCmd(), discoverCmd())
 	if err := cmd.Execute(); err != nil {
 		os.Exit(1)
 	}


cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+		PreRun: func(cmd *cobra.Command, args []string) {
+			fmt.Println("Command discover is experimental and under development.")
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			return runDiscover(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
```
qweeah, line 55: From previous implementation in the `artifacts` branch, if direct predecessor's artifact type doesn't match, we won't continue to check the predecessor's predecessor. Suppose we have three nodes: A, B and C with the below reference relation. When we call discover on `C` with artifact type as `type-1`, even though `A` matches the discover condition, the result won't show any node because the search is aborted on `B`.
```
A(type-1) -> B(type-2) -> C
```

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
+package main
```
shizhMSFT, line 1: Missing license header.

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+
+Example - Discover all the artifacts linked with the specified reference in a tree view:
+  oras discover localhost:5000/hello -o tree
+
+Example - Discover artifacts of type test-artifact test-artifact linked with the specified reference:
+  oras discover --artifact test-artifact localhost:5000/hello
+`,
+		Args: cobra.ExactArgs(1),
+		PreRun: func(cmd *cobra.Command, args []string) {
+			fmt.Println("Command discover is experimental and under development.")
```
shizhMSFT, line 47: It's better to call it `preview` than experimental.

We also want to mark the `discover` command as `preview` in the `oras --help` command. Let's discuss offline for the temporary UX.

cmd/oras/discover.go
shizhMSFT: Good catch. /cc @FeynmanZhou @yizha1

Let's keep the previous behavior for now.

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			return runDiscover(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
+	cmd.Flags().StringVarP(&opts.outputType, "output", "o", "table", fmt.Sprintf("Format in which to display references (%s, %s, or %s). tree format will show all references including nested", "table", "json", "tree"))
```
shizhMSFT, line 56:
```suggestion
	cmd.Flags().StringVarP(&opts.outputType, "output", "o", "table", "format in which to display references (table, json, or tree). tree format will show all references including nested")
```

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+
+		if len(*refs) != 0 {
+			fmt.Println()
+			printDiscoveredReferencesTable(*refs, opts.Verbose)
+		}
+	}
+	return nil
+}
+
+func getAllReferences(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string, node *tree.Node, queryGraph bool) (ocispec.Descriptor, *[]artifactspec.Descriptor, error) {
```
shizhMSFT, line 96:
```suggestion
func getAllReferences(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string, node *tree.Node, queryGraph bool) (ocispec.Descriptor, []artifactspec.Descriptor, error) {
```
```suggestion
func getAllReferences(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string, node *tree.Node, queryGraph bool) (ocispec.Descriptor, *[]artifactspec.Descriptor, error) {
```

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+	print(typeNameTitle, "Digest")
+	for _, ref := range refs {
+		print(ref.ArtifactType, ref.Digest)
+		if verbose {
+			printJSON(ref)
+		}
+	}
+}
+
+func printDiscoveredReferencesJSON(desc ocispec.Descriptor, refs []artifactspec.Descriptor) {
```
shizhMSFT, line 151: `printDiscoveredReferencesJSON` should have the same output as the [referrer API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md).

cmd/oras/discover.go
shizhMSFT, line 96: Let refactor this method to
- `fetchReferrers` for `json` and `table`.
- `fetchAllReferrers` for `tree`.

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+			fmt.Println()
+			printDiscoveredReferencesTable(*refs, opts.Verbose)
+		}
+	}
+	return nil
+}
+
+func getAllReferences(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string, node *tree.Node, queryGraph bool) (ocispec.Descriptor, *[]artifactspec.Descriptor, error) {
+	var results []artifactspec.Descriptor
+	err := repo.Referrers(ctx, desc, func(referrers []artifactspec.Descriptor) error {
```
shizhMSFT, line 98: Let's upgrade `oras-go` version and use the latest referrer implementation: https://github.com/oras-project/oras-go/pull/211

cmd/oras/discover.go
```diff
@@ -0,0 +1,179 @@
...
+func runDiscover(opts discoverOptions) error {
+	ctx, _ := opts.SetLoggerLevel()
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	// discover artifacts
+	root := tree.New(opts.targetRef)
+	desc, err := repo.Resolve(ctx, repo.Reference.Reference)
```
shizhMSFT, line 70: What if no tag or digest is provided?

cmd/oras/discover.go
qweeah: Good catch

cmd/oras/discover.go
qweeah: Will update with the evil default latest.

cmd/oras/discover.go
qweeah: Looks like there is no good way to mark a command as hidden in the `--help` outputs. I can think of two options 
- Add `preview` to the command short description
- Hide the command from oras `--help` output. (Chosen in current implementation)

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+
+func discoverCmd() *cobra.Command {
+	var opts discoverOptions
+	cmd := &cobra.Command{
+		Hidden: true,
+		Use:    "discover [options] <name:tag|name@digest>",
+		Short:  "Discover artifacts from remote registry",
+		Long: `Discover artifacts from remote registry
+
+Example - Discover all the artifacts linked with the specified reference:
```
Wwwsylvia, line 52: "artifacts linked with the specified **reference**" also looks a bit confusing🤔. We may need some rewording here.

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+			fmt.Println("Command discover is in preview and might have breaking changes coming.")
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			return runDiscover(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
+	cmd.Flags().StringVarP(&opts.outputType, "output", "o", "table", "format in which to display references (table, json, or tree). tree format will show all references including nested")
```
Wwwsylvia, line 72: "format in which to display references (table, json, or tree)"
Should we use "references" or "referrers" here? @FeynmanZhou @yizha1 can suggest.

Wwwsylvia:
Can we put a screenshot of the command demo in the comment?

cmd/oras/discover.go
qweeah: Let's use node/predecessors to replace all the deprecated naming here

cmd/oras/discover.go
qweeah: It is for oras-go, but we can extend it if there is a actual user need.

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+			printDiscoveredReferencesTable(refs, opts.Verbose)
+		}
+	}
+	return nil
+}
+
+func fetchReferrers(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string) ([]artifactspec.Descriptor, error) {
+	var results []artifactspec.Descriptor
+	err := repo.Referrers(ctx, desc, artifactType, func(referrers []artifactspec.Descriptor) error {
+		results = referrers
```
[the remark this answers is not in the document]
qweeah: Good catch, updated

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+			ocispec.Descriptor{
+				Digest:    r.Digest,
+				Size:      r.Size,
+				MediaType: r.MediaType,
+			},
+			artifactType, referrerNode)
+		if err != nil {
+			return nil, err
+		}
+		results = append(results, nestedReferrers...)
```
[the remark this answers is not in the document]
qweeah: It's in the tree view.

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+		PreRun: func(cmd *cobra.Command, args []string) {
+			fmt.Println("Command discover is in preview and might have breaking changes coming.")
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			return runDiscover(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
```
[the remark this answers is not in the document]
qweeah: It's out of this issue's scope. I can create an issue for that if you got any use case in mind.

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+
+Example - Discover all the referrers of manifest 'hello:latest' in registry 'localhost:5000' in a tree view:
+  oras discover localhost:5000/hello -o tree
+
+Example - Discover referrers with type 'test-artifact' of manifest 'hello:latest' in registry 'localhost:5000':
+  oras discover --artifact test-artifact localhost:5000/hello
+`,
+		Args: cobra.ExactArgs(1),
+		PreRun: func(cmd *cobra.Command, args []string) {
+			fmt.Println("Command discover is in preview and might have breaking changes coming.")
```
shizhMSFT, line 63: Let's review the UX offline tomorrow.

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// discover artifacts
+	root := tree.New(opts.targetRef)
+	desc, err := repo.Resolve(ctx, repo.Reference.ReferenceOrDefault())
+	if err != nil {
+		return err
+	}
+	var refs []artifactspec.Descriptor
+	if opts.outputType == "tree" {
+		refs, err = fetchAllReferrers(ctx, repo, desc, opts.artifactType, root)
```
shizhMSFT, line 92: We don't need `refs` for `tree`, do we?

shizhMSFT: I'm expecting something like

```go
desc, err := repo.Resolve(ctx, repo.Reference.ReferenceOrDefault())
if err != nil {
    return err
}

if opts.outputType == "tree" {
    root := tree.New(opts.targetRef)
    err = fetchAllReferrers(ctx, repo, desc, opts.artifactType, root)
    // check err
    return tree.Print(root)
}

refs, err = fetchReferrers(ctx, repo, desc, opts.artifactType)
// check err
if opts.outputType == "json" {
    return printDiscoveredReferencesJSON(desc, refs)
}

fmt.Println("Discovered", len(refs), "artifacts referencing", opts.targetRef)
fmt.Println("Digest:", desc.Digest)
if len(refs) > 0 {
    fmt.Println()
    printDiscoveredReferencesTable(refs, opts.Verbose)
}
```

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+
+func printDiscoveredReferencesJSON(desc ocispec.Descriptor, refs []artifactspec.Descriptor) {
+	type referrerDesc struct {
+		Digest    digest.Digest `json:"digest"`
+		MediaType string        `json:"mediaType"`
+		Artifact  string        `json:"artifactType"`
+		Size      int64         `json:"size"`
+	}
+	output := struct {
+		// https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#artifact-referrers-api-results
```
shizhMSFT, line 184:
```suggestion
		// Reference: https://github.com/oras-project/artifacts-spec/blob/v1.0.0-rc.1/manifest-referrers-api.md#artifact-referrers-api-results
```

shizhMSFT: We might want to write documentation for `printDiscoveredReferencesJSON` and move this reference there.

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+	print(typeNameTitle, "Digest")
+	for _, ref := range refs {
+		print(ref.ArtifactType, ref.Digest)
+		if verbose {
+			printJSON(ref)
+		}
+	}
+}
+
+func printDiscoveredReferencesJSON(desc ocispec.Descriptor, refs []artifactspec.Descriptor) {
```
shizhMSFT, line 176:
```suggestion
func printDiscoveredReferencesJSON(desc ocispec.Descriptor, refs []artifactspec.Descriptor) error {
```

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+func runDiscover(opts discoverOptions) error {
+	ctx, _ := opts.SetLoggerLevel()
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	// discover artifacts
+	root := tree.New(opts.targetRef)
+	desc, err := repo.Resolve(ctx, repo.Reference.ReferenceOrDefault())
```
shizhMSFT, line 86: @qweeah @Wwwsylvia @FeynmanZhou @yizha1 We need to make the default behavior consistent for all commands.

`latest` tag by default is convenient but also causes lots of problems.

At least, we need to inform the user that we are using a default tag like docker does.
```
$ docker pull registry
Using default tag: latest
...
```

This is achievable by
```go
ref := repo.Reference.ReferenceOrDefault()
if ref != repo.Reference.Reference {
    fmt.Println("Using default tag:", ref)
}
```

cmd/oras/discover.go
Wwwsylvia: Sounds good.🤔

cmd/oras/discover.go
Wwwsylvia: Yeah CLI userd won't understand `predecessor`.

cmd/oras/discover.go
qweeah: Let me update this way but I think we do need a proper discussion on that.

cmd/oras/discover.go
qweeah: Added `latest` tag a default:

- Tree
```
Using default tag: latest
localhost:5000/net-monitor:latest
├── signature/example
│   └── sha256:2437d8720c28d2bd2ad830d218ebb47942c099e21ff9cf9abbeb5e396203f6dd
│       └── signature/example
│           └── sha256:7c8fa4053dd322a3c1ea069a43c2cfba90ef4db4c9fed71f10da9d3a2071479c
└── sbom/example
    └── sha256:4b6726d63d4283fa579ddf29c4b0838025e2e27e5856e25db78fb50387461ee3
        └── signature/example
            └── sha256:673048d3beb0fada31b268b0e56a89c49fd6c0d2ec9b5ba63330e712b0c6758f
```
- Table
```
Using default tag: latest
Discovered 2 artifacts referencing localhost:5000/net-monitor:latest
Digest: sha256:4d5dd672805a7e20d19d1ae89bf4c5d371c47a018cb33bbc4dec989444602bd0

Artifact Type       Digest
signature/example   sha256:2437d8720c28d2bd2ad830d218ebb47942c099e21ff9cf9abbeb5e396203f6dd
sbom/example        sha256:4b6726d63d4283fa579ddf29c4b0838025e2e27e5856e25db78fb50387461ee3
```
- JSON
```
Using default tag: latest
{
  "referrers": [
    {
      "digest": "sha256:2437d8720c28d2bd2ad830d218ebb47942c099e21ff9cf9abbeb5e396203f6dd",
      "mediaType": "application/vnd.cncf.oras.artifact.manifest.v1+json",
      "artifactType": "signature/example",
      "size": 411
    },
    {
      "digest": "sha256:4b6726d63d4283fa579ddf29c4b0838025e2e27e5856e25db78fb50387461ee3",
      "mediaType": "application/vnd.cncf.oras.artifact.manifest.v1+json",
      "artifactType": "sbom/example",
      "size": 401
    }
  ]
}
```

cmd/oras/discover.go
```diff
@@ -0,0 +1,207 @@
...
+	outputType   string
+}
+
+func discoverCmd() *cobra.Command {
+	var opts discoverOptions
+	cmd := &cobra.Command{
+		Hidden: true,
+		Use:    "discover [options] <name:tag|name@digest>",
+		Short:  "Discover artifacts from remote registry",
+		Long: `Discover artifacts from remote registry
```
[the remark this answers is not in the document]
qweeah: Rephrased

shizhMSFT (approved):
LGTM

PR #429, opened 2022-06-30
Add digest-only option

### LINKED ISSUE #326: oras push/cp --digest-only
To support [tag update scenarios](https://github.com/notaryproject/notation/issues/45), where a new digest, and all its references should be pushed before a tag update is applied, we need a way to push a container image, or any `artifactType`, deferring the tag application.

Initially, I considered `docker push --digest-only`, and that may still be the best idea. Except, that would require a change to the docker client. 
This proposal focuses on `oras push` with a `--digest-only` flag. 

Perhaps `oras cp` with `--digest-only`

See [Docs: Add Promoting Scenario oras-project/oras#98](https://github.com/notaryproject/notation/pull/98) for more info

qweeah:
This PR adds a new option for pushing/copying/attaching without tag created in the target registry.

Resolves #326 

### DIFF
--- a/cmd/oras/internal/option/pusher.go
+++ b/cmd/oras/internal/option/pusher.go
@@ -1,0 +1,49 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


qweeah:
Closing since we can use empty tag to specify digest-only during pushing

PR #431, opened 2022-06-30
Remove push mandate on `subject`

### LINKED ISSUE #361: Support uploading an ORAS Artifact Manifest type without a subject
# Repro Steps
```bash
ACR_NAME=wabbitnetworks
REGISTRY=wabbitnetworks.azurecr.io
REPO=net-monitor
TAG=v1
IMAGE=$REGISTRY/${REPO}:$TAG
```
Assuming you have $IMAGE, the following command works:

```bash
oras push $REGISTRY/$REPO \
    --artifact-type 'signature/example' \
    --subject $IMAGE \
    ./readme.md:application/md
```

However, if I want to upload a readme file, that doesn't have a `subject`, I'm unable to exclude the `--subject` field:
```bash
oras push $REGISTRY/$REPO:readme \
    --artifact-type 'signature/example' \
    ./readme.md:application/md
```

nima:
Fixes #361

Does NOT fix oras-project/oras-go#206; since this client does not use the go library.

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -124,13 +124,14 @@ func runPush(opts pushOptions) error {
 
 	// bake artifact
 	var pushOpts []oras.PushOpt
-	if opts.artifactType != "" {
+	if opts.artifactType != "" && opts.artifactRefs != "" {
 		resolver, err = orasDocker.WithDiscover(opts.artifactRefs, resolver, orasDocker.NewOpts(ropts))
 		if err != nil {
 			return err
 		}
 
-		manifest, err := loadReference(ctx, resolver, opts.artifactRefs)
+		var manifest ocispec.Descriptor
+		manifest, err = loadReference(ctx, resolver, opts.artifactRefs)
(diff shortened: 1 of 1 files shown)


cmd/oras/push.go
```diff
@@ -125,16 +126,18 @@ func runPush(opts pushOptions) error {
...
-		}
-
-		manifest, err := loadReference(ctx, resolver, opts.artifactRefs)
-		if err != nil {
+		var manifest ocispec.Descriptor
+		var replacement remotes.Resolver
+		replacement, err = orasDocker.WithDiscover(opts.artifactRefs, resolver, orasDocker.NewOpts(ropts))
+		if err == nil {
+			resolver = replacement
+			manifest, err = loadReference(ctx, resolver2, opts.artifactRefs)
```
nima, line 134: This is wrong, not sure why this is what was pushed.  Will clean-up and also consider not calling discover at all, rather than call and rely on it failing with the hostname error.

Issue #434, opened 2022-07-04
Fix description that out of date


CharingJC:
https://github.com/oras-project/oras/blob/main/cmd/oras/pull.go

```
Example - Pull only files with the "application/vnd.oci.image.layer.v1.tar" media type (default):
  oras pull localhost:5000/hello:latest
Example - Pull only files with the custom "application/vnd.me.hi" media type:
  oras pull localhost:5000/hello:latest -t application/vnd.me.hi
Example - Pull all files, any media type:
  oras pull localhost:5000/hello:latest -a
```

first two scenarios are covered by the third one, and no more "-a" flag is required.

CharingJC:
should be

Example - Pull all files:
  oras pull localhost:5000/hello:latest

PR #437, opened 2022-07-05
fix: update pullCmd description

CharingJC:
Resolves https://github.com/oras-project/oras/issues/434

According to the latest UX, following command can pull all files, any media type:
`oras pull localhost:5000/hello:latest`


CharingJC:
Close due to DCO concerns

Issue #439, opened 2022-07-05
Return error if tag or digest is not specified


shizhMSFT:
Impacted commands:

- [x] `oras pull` #440
- [x] `oras discover` #442
- [x] `oras copy` #432

qweeah:
ack

CharingJC:
I will help with the copyCmd

FeynmanZhou:
@shizhMSFT  I assume `oras pull` has supported the error return. Is there any additional work?

```
$ ./oras pull mcr.microsoft.com/mcr/hello-world
Error: image reference format is invalid. Please specify <name:tag|name@digest>
```

qweeah:
@FeynmanZhou You must be using ORAS 0.12.0

sajayantony:
So does this mean that oras will not default to latest?

PR #440, opened 2022-07-05
If no tag or digest specified when pulling, return error

### LINKED ISSUE #428: Empty tag is not checked for `oras pull`
https://github.com/oras-project/oras/blob/bb75f50fb5c21a59beb09ce1b632291182046632/cmd/oras/pull.go#L151

qweeah:
Related to #439, resolves #428 


internal/errors/errors.go
```diff
@@ -0,0 +1,7 @@
+package errors
+
+import "errors"
+
+var (
+	ErrInvalidReference = errors.New("image reference format is invalid. Please specify <name:tag|name@digest>")
```
shizhMSFT, line 6: The error message is too general. For `oras copy`, how do we know this error message is for the source or the destination?

internal/errors/errors.go
```diff
@@ -0,0 +1,7 @@
+package errors
```
shizhMSFT, line 1: missing license header

internal/errors/errors.go
shizhMSFT, line 6: missing documentation

qweeah: Let me add registry and repository in so user can have more context on the cause of the error

internal/errors/errors.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+import "errors"
+
+// ErrInvalidReference creates a new error based on the reference string.
+func ErrInvalidReference(reference string) error {
```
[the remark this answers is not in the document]
qweeah: Originally I am thinking the same thing, but looks like this error is most used for processing user input and displaying, I didn't see any use case we wil need to do sth like `err.(type)==ErrInvalidReference`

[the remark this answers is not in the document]
Wwwsylvia: Okay

shizhMSFT (approved):
LGTM

internal/errors/errors.go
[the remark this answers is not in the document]
qweeah: Some udpates on this:
- added `new` prefix to the error generation function
- moved the error to main package since it's directly related to commandline processing/displaying

Wwwsylvia (approved):
LGTM

PR #449, opened 2022-07-18
feat: add `fetch` command to get manifest of an artifact

### LINKED ISSUE #301: ORAS get-manifest command
ORAS Artifacts, notation and other clis associate references to a digest. 
However, users think of tags.
The oras CLI and go-lib would take a tag, and return a digest, and the associated manifest.

qweeah:
This PR adds an new command `manifest` and a subcommand `fetch` for displaying the manifest of an artifact reference with below changes:

- Moves `manifest` related code into a new folder. Cobra related code should be put in the `cmd.go` and functional code should be put in the `<sub-command-name>.go`. This practice can be applied when we have more sub-commands in `blob` and `repository` in the future.

- Adds a new mocking package to mock repository level operations (e.g. fetch, fetch reference, resolve), add used it for unit testing.

Resolves #301, #463


qweeah:
Convert to draft since we still need to support `--platform` and that need some work.

cmd/oras/fetch-manifest.go
```diff
@@ -31,26 +35,28 @@ type getManifestOptions struct {
...
 
-Example - Get manifest and save to manifest.json:
-  oras get-manifest -output manifest.json localhost:5000/hello:latest
+Example - Get manifest with specified media type:
+  oras get-manifest --media-type 'application/vnd.oci.image.manifest.v1+json' localhost:5000/hello:latest
 
-Example - Get manifest with prettified json result:
-  oras get-manifest --pretty localhost:5000/hello:latest
+Example - Get manifest with raw json result:
+  oras get-manifest --raw localhost:5000/hello:latest
```
sajayantony, line 59: Is the command get-manifest or fetch-manifest. 
Lesser characters the better IMHO. 
can we just go for manifest get and put. 
this way if someone downloads a manifest and adds and annotation, they can then upload the updated manifest as well. 

/cc @FeynmanZhou

qweeah: Now in this PR I am aligning all the naming to fetch-manifest. 

I suggest we use `fetch` command and placing all the feature provided in this PR into its sub-command `manifest`. So the command would look like
```
oras fetch manifest localhost:5000/hello:latest
```

UX like `oras manifest get localhost:5000/hello:latest` is not unified from what we have already provided since current all commands are verbs (push, pull, attach, discover)

sajayantony: I would still recommend  for get and put since they are shorter and more user friendly. The CLI is about UX which can and may diverge from the go libraries. Check with @FeynmanZhou on the usability of this command. 

Similar to pull and push these are dockerism which don’t have equivalence in the libraries.

qweeah: Let me discuss with @FeynmanZhou. BTW `git` also uses `fetch` on status updating (so common developers should have muscle memory of typing this word :) ).

sajayantony: Btw docker also has a manifest command eve though they have first level verbs. Like images etc. please also make sure we have a provision to write a manifest even if we don’t mAke the command immediately.

qweeah: > Btw docker also has a manifest command eve though they have first level verbs. Like images etc. please also make sure we have a provision to write a manifest even if we don’t mAke the command immediately.

So do crane. Will discuss all that with @FeynmanZhou when @shizhMSFT is back next week.

cmd/oras/internal/option/platform.go
```diff
@@ -0,0 +1,108 @@
...
+		// OSVersion is splitted by comma
+		p.OSVersion = parts[1]
+	}
+
+	parts = strings.Split(parts[0], "/")
+	if len(parts) < 2 {
+		return ocispec.Platform{}, fmt.Errorf("failed to parse platform '%s': expected format os/arch[/variant]", opts.Platform)
+	}
+	if len(parts) > 3 {
+		return ocispec.Platform{}, fmt.Errorf("failed to parse platform '%s': too many slashes", opts.Platform)
```
binbin-li, line 53: nit: maybe the error msg can specify the maximum slashes it accepts.

cmd/oras/internal/option/platform.go
```diff
@@ -0,0 +1,108 @@
...
+	Platform string
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *Platform) ApplyFlags(fs *pflag.FlagSet) {
+	fs.StringVarP(&opts.Platform, "platform", "", "", "fetch the manifest of a specific platform if target is multi-platform capable")
+}
+
+// ParsePlatform parses the input platform flag to an oci platform type.
+func (opts *Platform) ParsePlatform() (ocispec.Platform, error) {
```
binbin-li, line 40: Does it need to be exported?

qweeah:
Will update this to `oras manifest get`

cmd/oras/fetch-manifest.go
qweeah: Updated to `manifest fetch` and using `get` as an alias

cmd/oras/internal/option/platform.go
```diff
@@ -0,0 +1,120 @@
...
+	// OS/Arch/[Variant]
+	p.OS = parts[0]
+	if p.OS == "" {
+		return ocispec.Platform{}, fmt.Errorf("invalide platform: OS cannot be empty")
+	}
+	p.Architecture = parts[1]
+	if p.Architecture == "" {
+		return ocispec.Platform{}, fmt.Errorf("invalide platform: Architecture cannot be empty")
+	}
+	if len(parts) > 2 {
```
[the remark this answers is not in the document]
qweeah: Yes it is. It will be regarded as use want a image with OS as `os` and architecture as `arch`. BTW It's actually in the `TestPlatform_parse` test function.

qweeah:
This PR is still pending on https://github.com/oras-project/oras-go/pull/244

cmd/oras/manifest/cmd.go
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+package manifest
+
+import (
+	"github.com/spf13/cobra"
+	"oras.land/oras/cmd/oras/internal/option"
+)
+
+func Cmd() *cobra.Command {
```
shizhMSFT, line 23: Should we call it `Command()`?

cmd/oras/manifest/cmd.go
```diff
@@ -0,0 +1,69 @@
...
+	return cmd
+}
+
+func fetchCmd() *cobra.Command {
+	var opts fetchOptions
+	cmd := &cobra.Command{
+		Use:   "fetch [flags] <name:tag|name@digest>",
+		Short: "[Preview] Fetch manifest of the target artifact",
+		Long: `[Preview] Fetch manifest of the target artifact
+** This command is in preview and under development. **
```
shizhMSFT, line 41: An example for `--descriptor` would be better.

shizhMSFT (approved):
LGTM

PR #454, opened 2022-07-21
Fix status tracking bug in push, attach and copy commmands

### LINKED ISSUE #443: `push` `attach` `copy` commands display partial transfer status
# Reproduce
1. Start a local registry service
```
docker run -d -p 5000:5000 ghcr.io/oras-project/registry:v0.0.3-alpha
```

2. Create files to be uploaded
```
echo "testing" > foo
echo "testing" > bar
```
3. Push the files
```
oras push localhost:5000/oras:test foo bar -v
```
The output will look like below, where `Uploading` and `Uploaded` status is missing either for `foo` or `bar`
```
Preparing foo
Preparing bar
Exists    44136fa355b3 application/vnd.unknown.config.v1+json
Uploading 36d054bb9d20 bar
Uploaded  36d054bb9d20 bar
// MISSING HERE: Uploading 36d054bb9d20 foo
// MISSING HERE: Uploaded  36d054bb9d20 foo
Uploading bad384b6cfac application/vnd.oci.image.manifest.v1+json
Uploaded  bad384b6cfac application/vnd.oci.image.manifest.v1+json
Pushed localhost:5000/oras:test
Digest: sha256:bad384b6cfac1bf1dbfbed52a45948fa2f977765c3e35563da21a4a6d15b94ec
```
`copy` and `attach` command have similar issue.

# Possible solution
## Option 1: Display transfer status for each file
Since the layer blob are identical and only the annontation title is different, CLI should
1) Maintain a list to track all the descriptors referring the the layer blob during loading file reference, add the list to a map keyed by the layer blob's digest.
2) When outputing transfer status, query the map with blob's digest. If a list is found, print one line for each entry in the list
```
               +----------+     +--------+
               | Manifest +-----> config |
               +--+----+--+     +--------+
                  |    |
Annotation Title:Foo  Annotation Title:Bar
                  |    |
              +---v----v---+
              | Layer Blob |
              +------------+
```

## Option 2: No file name in the status output
Showing media types instead of file names in the output will work around this issue.
```
Uploading 36d054bb9d20 application/vnd.oci.image.layer.v1.tar
Uploaded  36d054bb9d20 application/vnd.oci.image.layer.v1.tar
```

qweeah:
This PR fixes a bug for `push`, `attach` and `copy` commands. Resolves #443 

For `attach` and `push`, ORAS CLI collects all the file names pointing to the same digest before pushing the content. So when pushing a descriptor, if its digest is store in the pre-baked map, then tracking status of all the related file names wil be printed out.

For `copy` command, it's impossible to setup digest->file names mapping before pushing. ORAS CLI will ignore all the file names and only print status traking for media type.


qweeah:
Discarding this PR, will open another one according to https://github.com/oras-project/oras/issues/443#issuecomment-1204853421

PR #464, opened 2022-07-29
perf: add FetchReference support in the caching proxy

qweeah:
This PR adds FetchReference() for proxied cache, which will be used for optimizing fetching manifest content (when `--descriptor` is not provided) after `oras.Resolve()`.

Related to #301


shizhMSFT (approved):
LGTM

PR #465, opened 2022-08-01
feat: push an ORAS artifact manifest type without a subject

yuehaoliang-microsoft:
feat: push an ORAS artifact manifest type without a subject (#361) 

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content/file"
@@ -40,6 +41,7 @@ type pushOptions struct {
 
 	targetRef         string
 	manifestConfigRef string
+	artifactType      string
 }
 
 func pushCmd() *cobra.Command {
@@ -79,6 +81,7 @@ Example - Push file to the HTTP registry:
 	}
 
 	cmd.Flags().StringVarP(&opts.manifestConfigRef, "manifest-config", "", "", "manifest config file")
+	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
 
 	option.ApplyFlags(&opts, cmd.Flags())
 	return cmd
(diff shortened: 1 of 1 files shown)


cmd/oras/push.go
```diff
@@ -101,7 +104,13 @@ func runPush(opts pushOptions) error {
 	copyOptions.PreCopy = display.StatusPrinter("Uploading", opts.Verbose)
 	copyOptions.OnCopySkipped = display.StatusPrinter("Exists   ", opts.Verbose)
 	copyOptions.PostCopy = display.StatusPrinter("Uploaded ", opts.Verbose)
-	desc, err := packManifest(ctx, store, annotations, &opts)
+
+	var desc ocispec.Descriptor
+	if opts.artifactType == "" {
```
binbin-li, line 109: Might be out of scope, looks like `packOCIManifest` and `packORASArtifactManifest` share the same input and output, and their logics are very similar. Maybe we can apply strategy pattern to it.

cmd/oras/push.go
```diff
@@ -101,7 +104,13 @@ func runPush(opts pushOptions) error {
...
+		desc, err = packOCIManifest(ctx, store, annotations, &opts)
+	} else {
```
qweeah, line 111: Since ORAS artifact don't have config, we need to check if `manifestConfigRef` is empty. If not, output a warning tp remind user that the parameter will be ignored.

cmd/oras/push.go
```diff
@@ -154,3 +163,22 @@ func packManifest(ctx context.Context, store *file.Store, annotations map[string
...
 	return manifestDesc, nil
 }
+
+func packORASArtifactManifest(ctx context.Context, store *file.Store, annotations map[string]map[string]string, opts *pushOptions) (ocispec.Descriptor, error) {
+	ociDescs, err := loadFiles(ctx, store, annotations, opts.FileRefs, opts.Verbose)
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	orasDescs := make([]artifactspec.Descriptor, len(ociDescs))
+	for i := range ociDescs {
```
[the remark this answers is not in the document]
qweeah: Use `for _, d := range descs` to do the iteration.

cmd/oras/push.go
```diff
@@ -154,3 +163,22 @@ func packManifest(ctx context.Context, store *file.Store, annotations map[string
...
+		orasDescs[i] = ociToArtifact(ociDescs[i])
+	}
+	manifestDesc, err := oras.PackArtifact(ctx, store, opts.artifactType, orasDescs, oras.PackArtifactOptions{})
```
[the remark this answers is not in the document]
qweeah: Yes it's needed

qweeah (changes requested):

cmd/oras/push.go
qweeah: And also the code used to pack an oras artifact is `attach.go` is duplicated with`packORASArtifactManifest` here.

yuehaoliang-microsoft:
Closing as UX for uploading an ORAS artifact manifest without a subject is not supported in CLI.

Issue #466, opened 2022-08-01
`--artifact-type` support in `oras push`


shizhMSFT:
This proposal contains two changes derived from #361.

1. When `--artifact-type $type` is specified, `oras push` will change the media type of the config blob, which is equivalent to `--manifest-config <(echo {}):$type`
2. When there is no blob specified, `--artifact-type` must be specified and an ORAS artifact manifest will be pushed.

sajayantony:
Could we do a zero byte blob for this?

shizhMSFT:
> Could we do a zero byte blob for this?

We are still not sure if a zero-byte config blob is allowed for all registries.

shizhMSFT:
BTW, even if we have a zero-byte blob, we still need to upload that empty blob.

yuehaoliang-microsoft:
Hi, team! I'll start to work on resolving this issue.

sajayantony:
Yes zero byte blobs need to be uploaded but you can choose to short circuit it on the pull path. 
Also registry implementation may also choose to short circuit the writes.

sajayantony:
Thinking more have we considered using --config-media-type and --artifact-type to disambiguate between them.
This way there is no guess work on the CLI and the flags are mutually exclusive since `artifactManifest` does not have a config media type.

shizhMSFT:
> Thinking more have we considered using --config-media-type and --artifact-type to disambiguate between them. This way there is no guess work on the CLI and the flags are mutually exclusive since `artifactManifest` does not have a config media type.

I think here is to make the concept of `artifact type` transparent. Currently, we use OCI manifest to mimic ORAS artifact manifest where the config media type is used as artifact type. In the future when ORAS artifact manifest is widely adopted, we can default `oras push` to ORAS artifact manifest without interface changes.

yuehaoliang-microsoft:
I made a mistake to handle #483, as I clicked the Github UI "fetch upstream" in my repo/branch. Since that PR attempted to merge from my repo/branch, after my repo/branch is updated to upstream, the PR was automatedly closed.

shizhMSFT:
Closing this as #482 is merged and #486 is closed.

SteveLasker:
Would like to make sure we’re in sync on the UX and experience before closing.

shizhMSFT:
Closing again as https://github.com/oras-project/oras-go/issues/271 is planned.

PR #468, opened 2022-08-02
feat!: pass annotations as a string

### LINKED ISSUE #362: ORAS Push should support annotations, without having to specify a file
# Repro steps

```bash
oras push $REGISTRY/$REPO \
    --artifact-type 'signature/example' \
    --subject $IMAGE \
    --manifest-annotations io.cncf.oras.artifact.created=$CREATED_DATE \
    ./readme.md:application/md
```

Fails, as it expects `--manifest-annotations` to be a file. I was hoping the docs were wrong :(

```bash
Flags:
      --artifact-type string          artifact type
  -c, --config stringArray            auth config path
  -d, --debug                         debug mode
      --disable-path-validation       skip path validation
      --dry-run                       push to a dummy registry instead of the actual remote registry
      --export-manifest string        export the pushed manifest
  -h, --help                          help for push
      --insecure                      allow connections to SSL registry without certs
      --manifest-annotations string   manifest annotation file
```

## Desired

The default experience should enable a user to pass annotations as a string, so they don't have to create a file just to upload a single annotation
I'd suggest a file could be sent in as a stream if we really needed to support files.
We should also support multiple `--manifest-annotation` values to be passed in:

```bash
oras push $REGISTRY/$REPO \
    --artifact-type 'signature/example' \
    --subject $IMAGE \
    --manifest-annotations io.cncf.oras.artifact.created=$CREATED_DATE \
    --manifest-annotations io.cncf.oras.artifact.description="foo"\
    ./readme.md:application/md
```

junczhuMSFT:
Resolve #362 

BREAKING CHANGE: update annotation UX
Signed-off-by: Juncheng Zhu <junczhu@microsoft.com>

### DIFF
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -35,7 +35,7 @@ const (
 var (
 	errAnnotationConflict    = errors.New("annotations cannot be specified via flags and file at the same time")
 	errAnnotationFormat      = errors.New("annotation MUST be a key-value pair")
-	errAnnotationDuplication = errors.New("annotation key duplication")
+	errAnnotationDuplication = errors.New("duplicate annotation key")
 )
 
 // Packer option struct.
@@ -70,7 +70,6 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 
 // LoadManifestAnnotations loads the manifest annotation map.
 func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
-	annotations = make(map[string]map[string]string)
 	if opts.AnnotationFilePath != "" && len(opts.ManifestAnnotations) != 0 {
 		return nil, errAnnotationConflict
 	}
@@ -80,11 +79,12 @@ func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string
 		}
 	}
 	if annotationsLength := len(opts.ManifestAnnotations); annotationsLength != 0 {
+		annotations = make(map[string]map[string]string)
 		if err = parseAnnotationFlags(opts.ManifestAnnotations, annotations); err != nil {
 			return nil, err
 		}
 	}
-	return annotations, nil
+	return
 }
 
 // decodeJSON decodes a json file v to filename.
@@ -98,19 +98,18 @@ func decodeJSON(filename string, v interface{}) error {
 }
 
 // parseAnnotationFlags resharps annotationslice to k-v type and updates annotations
-func parseAnnotationFlags(ManifestAnnotations []string, annotations map[string]map[string]string) error {
+func parseAnnotationFlags(annotationsFlagParam []string, annotations map[string]map[string]string) error {
 	manifestAnnotations := make(map[string]string)
-	for _, anno := range ManifestAnnotations {
-		parts := strings.SplitN(anno, "=", 2)
-		if len(parts) != 2 {
+	for _, anno := range annotationsFlagParam {
+		key, val, success := strings.Cut(anno, "=")
+		if !success {
 			return errAnnotationFormat
 		}
-		parts[0], parts[1] = strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
-		if _, ok := manifestAnnotations[parts[0]]; ok {
-
-			return fmt.Errorf("found annotation key, %v, more than once, %w", parts[0], errAnnotationDuplication)
+		key, val = strings.TrimSpace(key), strings.TrimSpace(val)
+		if _, ok := manifestAnnotations[key]; ok {
+			return fmt.Errorf("%w: %v, ", errAnnotationDuplication, key)
 		}
-		manifestAnnotations[parts[0]] = parts[1]
+		manifestAnnotations[key] = val
 	}
 	annotations[annotationManifest] = manifestAnnotations
 	return nil
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"regexp"
 	"strings"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -34,13 +33,13 @@ const (
 )
 
 var (
-	errAnnotationConflict    = errors.New("annotation input conflict")
+	errAnnotationConflict    = errors.New("annotations cannot be specified via flags and file at the same time")
 	errAnnotationFormat      = errors.New("annotation MUST be a key-value pair")
 	errAnnotationDuplication = errors.New("annotation key duplication")
 )
 
-// Pusher option struct.
-type Pusher struct {
+// Packer option struct.
+type Packer struct {
 	ManifestExportPath     string
 	PathValidationDisabled bool
 	AnnotationsFilePath    string
@@ -50,7 +49,7 @@ type Pusher struct {
 }
 
 // ApplyFlags applies flags to a command flag set.
-func (opts *Pusher) ApplyFlags(fs *pflag.FlagSet) {
+func (opts *Packer) ApplyFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&opts.ManifestExportPath, "export-manifest", "", "", "export the pushed manifest")
 	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil, "manifest annotations")
 	fs.StringVarP(&opts.AnnotationsFilePath, "annotations-file", "", "", "path of the annotation file")
@@ -58,7 +57,7 @@ func (opts *Pusher) ApplyFlags(fs *pflag.FlagSet) {
 }
 
 // ExportManifest saves the pushed manifest to a local file.
-func (opts *Pusher) ExportManifest(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) error {
+func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) error {
 	if opts.ManifestExportPath == "" {
 		return nil
 	}
@@ -70,10 +69,9 @@ func (opts *Pusher) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 }
 
 // LoadManifestAnnotations loads the manifest annotation map.
-func (opts *Pusher) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
+func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
 	annotations = make(map[string]map[string]string)
 	if opts.AnnotationsFilePath != "" && len(opts.ManifestAnnotations) != 0 {
-		fmt.Fprintln(os.Stderr, "WARNING! --manifest--anotation and --manifest-annotation-file can not be apply at the same time.")
 		return nil, errAnnotationConflict
 	}
 	if opts.AnnotationsFilePath != "" {
@@ -82,7 +80,7 @@ func (opts *Pusher) LoadManifestAnnotations() (annotations map[string]map[string
 		}
 	}
 	if annotationsLength := len(opts.ManifestAnnotations); annotationsLength != 0 {
-		if err = getAnnotationsMap(opts.ManifestAnnotations, annotations); err != nil {
+		if err = parseAnnotationFlags(opts.ManifestAnnotations, annotations); err != nil {
 			return nil, err
 		}
 	}
@@ -99,20 +97,20 @@ func decodeJSON(filename string, v interface{}) error {
 	return json.NewDecoder(file).Decode(v)
 }
 
-// getAnnotationsMap resharps annotationslice to k-v type and updates annotations
-func getAnnotationsMap(ManifestAnnotations []string, annotations map[string]map[string]string) error {
-	re := regexp.MustCompile(`=\s*`)
+// parseAnnotationFlags resharps annotationslice to k-v type and updates annotations
+func parseAnnotationFlags(ManifestAnnotations []string, annotations map[string]map[string]string) error {
 	rawAnnotationsMap := make(map[string]string)
-	for _, rawAnnotationStr := range ManifestAnnotations {
-		rawAnnotation := re.Split(rawAnnotationStr, 2)
-		if annotationLength := len(rawAnnotation); annotationLength != 2 {
+	for _, anno := range ManifestAnnotations {
+		parts := strings.SplitN(anno, "=", 2)
+		if len(parts) != 2 {
 			return errAnnotationFormat
 		}
-		rawAnnotation[0], rawAnnotation[1] = strings.TrimSpace(rawAnnotation[0]), strings.TrimSpace(rawAnnotation[1])
-		if _, ok := rawAnnotationsMap[rawAnnotation[0]]; ok {
-			return errAnnotationDuplication
+		parts[0], parts[1] = strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
+		if _, ok := rawAnnotationsMap[parts[0]]; ok {
+
+			return fmt.Errorf("found annotation key, %v, more than once, %w", parts[0], errAnnotationDuplication)
 		}
-		rawAnnotationsMap[rawAnnotation[0]] = rawAnnotation[1]
+		rawAnnotationsMap[parts[0]] = parts[1]
 	}
 	annotations[annotationManifest] = rawAnnotationsMap
 	return nil
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -105,7 +105,6 @@ func parseAnnotationFlags(flags []string, annotations map[string]map[string]stri
 		if !success {
 			return fmt.Errorf("%w: %s", errAnnotationFormat, anno)
 		}
-		key, val = strings.TrimSpace(key), strings.TrimSpace(val)
 		if _, ok := manifestAnnotations[key]; ok {
 			return fmt.Errorf("%w: %v, ", errAnnotationDuplication, key)
 		}
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -42,7 +42,7 @@ var (
 type Packer struct {
 	ManifestExportPath     string
 	PathValidationDisabled bool
-	AnnotationsFilePath    string
+	AnnotationFilePath     string
 	ManifestAnnotations    []string
 
 	FileRefs []string
@@ -52,7 +52,7 @@ type Packer struct {
 func (opts *Packer) ApplyFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&opts.ManifestExportPath, "export-manifest", "", "", "export the pushed manifest")
 	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil, "manifest annotations")
-	fs.StringVarP(&opts.AnnotationsFilePath, "annotations-file", "", "", "path of the annotation file")
+	fs.StringVarP(&opts.AnnotationFilePath, "annotation-file", "", "", "path of the annotation file")
 	fs.BoolVarP(&opts.PathValidationDisabled, "disable-path-validation", "", false, "skip path validation")
 }
 
@@ -71,11 +71,11 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 // LoadManifestAnnotations loads the manifest annotation map.
 func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
 	annotations = make(map[string]map[string]string)
-	if opts.AnnotationsFilePath != "" && len(opts.ManifestAnnotations) != 0 {
+	if opts.AnnotationFilePath != "" && len(opts.ManifestAnnotations) != 0 {
 		return nil, errAnnotationConflict
 	}
-	if opts.AnnotationsFilePath != "" {
-		if err = decodeJSON(opts.AnnotationsFilePath, &annotations); err != nil {
+	if opts.AnnotationFilePath != "" {
+		if err = decodeJSON(opts.AnnotationFilePath, &annotations); err != nil {
 			return nil, err
 		}
 	}
@@ -99,19 +99,19 @@ func decodeJSON(filename string, v interface{}) error {
 
 // parseAnnotationFlags resharps annotationslice to k-v type and updates annotations
 func parseAnnotationFlags(ManifestAnnotations []string, annotations map[string]map[string]string) error {
-	rawAnnotationsMap := make(map[string]string)
+	manifestAnnotations := make(map[string]string)
 	for _, anno := range ManifestAnnotations {
 		parts := strings.SplitN(anno, "=", 2)
 		if len(parts) != 2 {
 			return errAnnotationFormat
 		}
 		parts[0], parts[1] = strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
-		if _, ok := rawAnnotationsMap[parts[0]]; ok {
+		if _, ok := manifestAnnotations[parts[0]]; ok {
 
 			return fmt.Errorf("found annotation key, %v, more than once, %w", parts[0], errAnnotationDuplication)
 		}
-		rawAnnotationsMap[parts[0]] = parts[1]
+		manifestAnnotations[parts[0]] = parts[1]
 	}
-	annotations[annotationManifest] = rawAnnotationsMap
+	annotations[annotationManifest] = manifestAnnotations
 	return nil
 }
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -33,8 +33,8 @@ const (
 )
 
 var (
-	errAnnotationConflict    = errors.New("annotations cannot be specified via `--annotation` and `--annotation-file` at the same time")
-	errAnnotationFormat      = errors.New("annotation MUST be a key-value pair")
+	errAnnotationConflict    = errors.New("`--annotation` and `--annotation-file` cannot be both specified")
+	errAnnotationFormat      = errors.New("missing key in `--annotation` flag")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
 )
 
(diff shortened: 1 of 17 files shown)


cmd/oras/internal/option/pusher.go
```diff
@@ -18,26 +18,36 @@ package option
...
+	ManifestAnnotationSlice []string
 
 	FileRefs []string
 }
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Pusher) ApplyFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&opts.ManifestExportPath, "export-manifest", "", "", "export the pushed manifest")
-	fs.StringVarP(&opts.ManifestAnnotations, "manifest-annotations", "", "", "manifest annotation file")
+	fs.StringArrayVarP(&opts.ManifestAnnotationSlice, "annotation", "a", []string{}, "manifest annotations")
```
qweeah, line 49: Why not default to nil

qweeah: Suggest renaming `ManifestAnnotationSlice` to `ManifestAnnotations`

cmd/oras/internal/option/pusher.go
```diff
@@ -18,26 +18,36 @@ package option
...
+	fs.StringVarP(&opts.ManifestAnnotations, "manifest-annotations-file", "", "", "manifest annotation file")
```
qweeah, line 50:
```suggestion
	fs.StringVarP(&opts.AnnotationFilePath, "annotations-file", "", "", "path of the annotation file")
```

cmd/oras/internal/option/pusher.go
```diff
@@ -55,9 +65,19 @@ func (opts *Pusher) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 
 // LoadManifestAnnotations loads the manifest annotation map.
 func (opts *Pusher) LoadManifestAnnotations() (map[string]map[string]string, error) {
-	var annotations map[string]map[string]string
+	var err error
```
qweeah, line 68:
```suggestion
func (opts *Pusher) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
```

cmd/oras/internal/option/pusher.go
```diff
@@ -55,9 +65,19 @@ func (opts *Pusher) ExportManifest(ctx context.Context, fetcher content.Fetcher,
...
+	annotations := make(map[string]map[string]string)
 	if opts.ManifestAnnotations != "" {
```
qweeah, line 70: - terminate when both flags provided
- 1) load file or 2) get annotations flag

qweeah (changes requested):

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +93,22 @@ func decodeJSON(filename string, v interface{}) error {
...
 }
+
+// getAnnotationsMap get resharp annotationslice to target type
+func getAnnotationsMap(ManifestAnnotationSlice []string, annotations map[string]map[string]string) error {
+	re := regexp.MustCompile(`=\s*`)
+	rawAnnotationsMap := make(map[string]string)
+	for _, rawAnnotation := range ManifestAnnotationSlice {
+		rawAnnotation := re.Split(rawAnnotation, 2)
+		rawAnnotation[0], rawAnnotation[1] = strings.TrimSpace(rawAnnotation[0]), strings.TrimSpace(rawAnnotation[1])
+		if annotationLength := len(rawAnnotation); annotationLength != 2 {
```
qweeah, line 104: The length check should be done before trimming

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +93,22 @@ func decodeJSON(filename string, v interface{}) error {
...
+			return errors.New("invalid annotation")
```
qweeah, line 105: Need more infomation like how the annotation is invalid, like what annontion is invalid since user may pass an array.

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +93,22 @@ func decodeJSON(filename string, v interface{}) error {
...
+		}
```
qweeah, line 106:
```suggestion
	        parts := strings.SplitN(raw, "=", 2)
	        if len(parts) < 2 {
		        // return invalid
	        }
	        key, value := strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
```

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +93,22 @@ func decodeJSON(filename string, v interface{}) error {
 	defer file.Close()
 	return json.NewDecoder(file).Decode(v)
 }
+
+// getAnnotationsMap get resharp annotationslice to target type
+func getAnnotationsMap(ManifestAnnotationSlice []string, annotations map[string]map[string]string) error {
+	re := regexp.MustCompile(`=\s*`)
+	rawAnnotationsMap := make(map[string]string)
+	for _, rawAnnotation := range ManifestAnnotationSlice {
+		rawAnnotation := re.Split(rawAnnotation, 2)
```
qweeah, line 102: Dont reuse variable name for different types(`rawAnnotation` ), it's dangerous.

qweeah (changes requested):

cmd/oras/internal/option/pusher.go
qweeah: If annotation key doesn't contain `=`, we can use https://pkg.go.dev/strings#SplitN to simplify the parsing

qweeah:
Let there be unit tests.

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +93,22 @@ func decodeJSON(filename string, v interface{}) error {
 	defer file.Close()
 	return json.NewDecoder(file).Decode(v)
 }
+
+// getAnnotationsMap get resharp annotationslice to target type
+func getAnnotationsMap(ManifestAnnotationSlice []string, annotations map[string]map[string]string) error {
+	re := regexp.MustCompile(`=\s*`)
+	rawAnnotationsMap := make(map[string]string)
```
qweeah, line 100:
```suggestion
	manifestAnnotations := make(map[string]string)
```

junczhuMSFT:
Going to update UX, as discussed with the community, exiting with error when both flags are provided is a good option for now.  @shizhMSFT @sajayantony @qweeah

qweeah:
> Going to update UX, as discussed with the community, exiting with error when both flags are provided is a good option for now. @shizhMSFT
> 
> cc. @sajayantony @qweeah

@junczhuMSFT This looks good. I think #466 can also apply this UX: if both `manifest-config` and `artifact-type` are specified for `oras push`, then exit with error /cc @yuehaoliang-microsoft

cmd/oras/internal/option/pusher.go
```diff
@@ -18,26 +18,42 @@ package option
...
 	"github.com/spf13/pflag"
 	"oras.land/oras-go/v2/content"
 )
 
+const (
+	annotationManifest = "$manifest"
+)
+
+var (
+	errAnnotationConflict    = errors.New("annotation input conflict")
```
qweeah, line 37: Need more specific error information:

```suggestion
	errAnnotationConflict    = errors.New("annotations cannot be specified via flags and file at the same time")
```

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +98,22 @@ func decodeJSON(filename string, v interface{}) error {
 	defer file.Close()
 	return json.NewDecoder(file).Decode(v)
 }
+
+// getAnnotationsMap resharps annotationslice to k-v type and updates annotations
+func getAnnotationsMap(ManifestAnnotations []string, annotations map[string]map[string]string) error {
+	re := regexp.MustCompile(`=\s*`)
```
qweeah, line 104: Is there any reason that we can only use regular expression but not string process functions like `string.SplitN`?

Also this expression don't need to get compiled every time `getAnnotationsMap` is called.

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +98,22 @@ func decodeJSON(filename string, v interface{}) error {
 	defer file.Close()
 	return json.NewDecoder(file).Decode(v)
 }
+
+// getAnnotationsMap resharps annotationslice to k-v type and updates annotations
+func getAnnotationsMap(ManifestAnnotations []string, annotations map[string]map[string]string) error {
```
qweeah, line 103:
```suggestion
func parseAnnotationFlags(ManifestAnnotations []string, annotations map[string]map[string]string) error {
```

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +98,22 @@ func decodeJSON(filename string, v interface{}) error {
...
+	re := regexp.MustCompile(`=\s*`)
+	rawAnnotationsMap := make(map[string]string)
+	for _, rawAnnotationStr := range ManifestAnnotations {
```
qweeah, line 106: short and concise names are idiomatic in golang.
```suggestion
	for _, anno:= range ManifestAnnotations {
```

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +98,22 @@ func decodeJSON(filename string, v interface{}) error {
...
+		rawAnnotation := re.Split(rawAnnotationStr, 2)
```
qweeah, line 107:
```suggestion
		parts := re.Split(rawAnnotationStr, 2)
```

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +98,22 @@ func decodeJSON(filename string, v interface{}) error {
...
+		if annotationLength := len(rawAnnotation); annotationLength != 2 {
+			return errAnnotationFormat
+		}
```
qweeah, line 110:
```suggestion
		if len(rawAnnotation) != 2 {
			return errAnnotationFormat
		}
```

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +98,22 @@ func decodeJSON(filename string, v interface{}) error {
...
+		rawAnnotation[0], rawAnnotation[1] = strings.TrimSpace(rawAnnotation[0]), strings.TrimSpace(rawAnnotation[1])
+		if _, ok := rawAnnotationsMap[rawAnnotation[0]]; ok {
+			return errAnnotationDuplication
```
qweeah, line 113: you also need tell use which key is duplicated via the `errAnnotationDuplication`, suggest to use `fmt.Errorf` and `%w` to wrap the error type.

cmd/oras/internal/option/pusher.go
```diff
@@ -18,26 +18,42 @@ package option
...
 )
 
+const (
+	annotationManifest = "$manifest"
+)
+
+var (
+	errAnnotationConflict    = errors.New("annotation input conflict")
+	errAnnotationFormat      = errors.New("annotation MUST be a key-value pair")
+	errAnnotationDuplication = errors.New("annotation key duplication")
```
qweeah, line 39: Since those errors need to be checked and thrown in a fail-fast way. I suggest 1) add `annotations map[string]map[string]string` to the `Pusher` struct; 2) Make sure `LoadManifestAnnotations` get called in `cobra.Command.PreRunE`

cmd/oras/internal/option/pusher_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"errors"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
```
qweeah, line 26: Add `TestPusher_FlagsInit` here, you may refer to other option's test case.

qweeah (changes requested):

qweeah:
@junczhuMSFT Since this changes the `pusher` options, we should also update CLI help doc for `oras push` and `oras attach`

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

junczhuMSFT:
> 

Well Noted

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: fixed

junczhuMSFT:
> We need unit tests. Let there be unit tests.

Working on it

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: fixed

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/pusher.go
junczhuMSFT: updated

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+		if err = decodeJSON(opts.AnnotationFilePath, &annotations); err != nil {
+			return nil, err
+		}
+	}
+	if annotationsLength := len(opts.ManifestAnnotations); annotationsLength != 0 {
+		if err = parseAnnotationFlags(opts.ManifestAnnotations, annotations); err != nil {
+			return nil, err
+		}
+	}
+	return annotations, nil
```
junczhuMSFT, line 87: can use a "naked" return when having named return values

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+	manifestBytes, err := content.FetchAll(ctx, fetcher, desc)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
+}
+
+// LoadManifestAnnotations loads the manifest annotation map.
+func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
+	annotations = make(map[string]map[string]string)
```
shizhMSFT, line 73: If there is no annotation, we should return a nil map instead of an empty map

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	return json.NewDecoder(file).Decode(v)
+}
+
+// parseAnnotationFlags resharps annotationslice to k-v type and updates annotations
+func parseAnnotationFlags(ManifestAnnotations []string, annotations map[string]map[string]string) error {
```
shizhMSFT, line 101: Parameters should never begin with a capital letter.

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+	manifestAnnotations := make(map[string]string)
+	for _, anno := range ManifestAnnotations {
+		parts := strings.SplitN(anno, "=", 2)
```
shizhMSFT, line 104: Try [strings.Cut()](https://pkg.go.dev/strings#Cut)

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+		if len(parts) != 2 {
+			return errAnnotationFormat
+		}
+		parts[0], parts[1] = strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
+		if _, ok := manifestAnnotations[parts[0]]; ok {
+
```
shizhMSFT, line 110: nit: remove empty line

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+			return fmt.Errorf("found annotation key, %v, more than once, %w", parts[0], errAnnotationDuplication)
```
shizhMSFT, line 111: The error message is too verbose. We can simplify this.

shizhMSFT (changes requested):

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+	ManifestAnnotations    []string
+
+	FileRefs []string
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *Packer) ApplyFlags(fs *pflag.FlagSet) {
+	fs.StringVarP(&opts.ManifestExportPath, "export-manifest", "", "", "export the pushed manifest")
+	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil, "manifest annotations")
+	fs.StringVarP(&opts.AnnotationFilePath, "annotation-file", "", "", "path of the annotation file")
```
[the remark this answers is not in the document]
junczhuMSFT: Change took place according to Steve's comment, link attached.
[HERE](https://github.com/oras-project/oras/issues/362#issuecomment-1201990395)

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+const (
+	annotationManifest = "$manifest"
+)
+
+var (
+	errAnnotationConflict    = errors.New("annotations cannot be specified via flags and file at the same time")
+	errAnnotationFormat      = errors.New("annotation MUST be a key-value pair")
+	errAnnotationDuplication = errors.New("annotation key duplication")
```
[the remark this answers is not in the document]
junczhuMSFT: done

cmd/oras/internal/option/packer.go
junczhuMSFT: done

cmd/oras/internal/option/packer.go
junczhuMSFT: done

cmd/oras/internal/option/packer.go
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/spf13/pflag"
+	"oras.land/oras-go/v2/content"
+)
+
+const (
+	annotationManifest = "$manifest"
+)
+
+var (
+	errAnnotationConflict    = errors.New("annotations cannot be specified via `--annotation` and `--annotation-file` at the same time")
```
qweeah, line 36: Nit:
```suggestion
	errAnnotationConflict    = errors.New(" `--annotation` and `--annotation-file` cannot be both specified")
```

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,116 @@
...
+func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string]string, err error) {
+	if opts.AnnotationFilePath != "" && len(opts.ManifestAnnotations) != 0 {
+		return nil, errAnnotationConflict
+	}
+	if opts.AnnotationFilePath != "" {
+		if err = decodeJSON(opts.AnnotationFilePath, &annotations); err != nil {
+			return nil, err
+		}
+	}
+	if annotationsLength := len(opts.ManifestAnnotations); annotationsLength != 0 {
```
qweeah, line 81: Nit
```suggestion
	if len(opts.ManifestAnnotations) != 0 {
```

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,116 @@
...
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	return json.NewDecoder(file).Decode(v)
+}
+
+// parseAnnotationFlags resharps annotationslice to k-v type and updates annotations
+func parseAnnotationFlags(annotationFlagParam []string, annotations map[string]map[string]string) error {
```
qweeah, line 101:
```suggestion
func parseAnnotationFlags(annoFlags []string, annotations map[string]map[string]string) error {
```

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,116 @@
...
+	manifestAnnotations := make(map[string]string)
+	for _, anno := range annotationFlagParam {
+		key, val, success := strings.Cut(anno, "=")
+		if !success {
+			return errAnnotationFormat
```
qweeah, line 106:
```suggestion
			return fmt.Errorf("%w: %s", errAnnotationFormat, anno)
```

qweeah (changes requested):

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,117 @@
...
+func decodeJSON(filename string, v interface{}) error {
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	return json.NewDecoder(file).Decode(v)
+}
+
+// parseAnnotationFlags resharps annotationslice to k-v type and updates annotations
```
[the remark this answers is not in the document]
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
junczhuMSFT: ACK

qweeah (approved):
LGTM

cmd/oras/internal/option/packer.go
```diff
@@ -97,20 +97,19 @@ func decodeJSON(filename string, v interface{}) error {
...
+	for _, anno := range flags {
+		key, val, success := strings.Cut(anno, "=")
+		if !success {
+			return fmt.Errorf("%w: %s", errAnnotationFormat, anno)
 		}
-		parts[0], parts[1] = strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
-		if _, ok := manifestAnnotations[parts[0]]; ok {
-
-			return fmt.Errorf("found annotation key, %v, more than once, %w", parts[0], errAnnotationDuplication)
+		key, val = strings.TrimSpace(key), strings.TrimSpace(val)
```
shizhMSFT, line 108: Is there reason to trim spaces? Trim space for `key` seems reasonable but what if spaces in `val` are intended?

For example, `--annotation "greet=$(echo hello world)"` in bash.

cmd/oras/internal/option/pusher.go
junczhuMSFT: ACK

cmd/oras/internal/option/pusher_test.go
junczhuMSFT: ACK

cmd/oras/internal/option/packer.go
junczhuMSFT: Well Noted

shizhMSFT (approved):
LGTM

cmd/oras/internal/option/packer.go
```diff
@@ -0,0 +1,115 @@
...
+type Packer struct {
+	ManifestExportPath     string
+	PathValidationDisabled bool
+	AnnotationFilePath     string
+	ManifestAnnotations    []string
+
+	FileRefs []string
+}
+
+// ApplyFlags applies flags to a command flag set.
```
vsoch, line 51: Just a clarifying question here! In the first implementation I have, the user could provide an annotations file, and this wasn't just for the manifest, you could have keys `$manifest` or `$config` or the fullpath/relative path of a blob to indicate that we want to add annotations there. But now the new argument `--annotation key=value` seems to be scoped just for the final manifest, and not a config or particular blob. Is this a correct interpretation?

vsoch: And in this case, wouldn't we only care if both are defined given we have `$manifest` in the file lookup plus an `--annotation`? E.g., providing this new flag with an annotation file without manifest annotations should not be an issue, but as I see here the client is just disallowing both to be defined.

junczhuMSFT: Hello @vsoch thank you for insight.

1. If users want to update annotations including keys `$manifest`, `$config` and more, they can use `--annotation-file` and provide an annotations file
2. If users only want to update manifest annotations, they can use `--annotation`. They don't have to create a file just to upload a single annotation
3. I have set up a condition that `--annotation` and `--annotation-file` cannot be both specified to make sure their implements are isolated to each other

vsoch: Yes so my question is why not to allow a file with $manifest and a one off annotation that has a key that doesn’t overlap? It could be a use case given that someone has a file with common/shared annotations and then a few custom to a container.  I think this should be allowed because there is no complexity in terms of duplicate keys - it’s just reading a file and then adding one off annotations from the command line. And you’d immediately reject/exit given duplicates only.

Issue #275, opened 2021-05-19
Pull arbitrary manifest config


shizhMSFT:
As the solution to #142, #274 enables the CLI to pull manifest config. However, the user must provide the desired media type of the config or use the default.

The advantages of specified media types are
- Config can be virtually optional although it is a required field in the manifest.
- `oras` will not pull unrecognized config files.

In this issue, we discuss we should allow pulling arbitrary manifest config or not.

If we do want to pull arbitrary manifest config, we will have more implementation complexity.
- We need to explicitly specify a hybrid store of a file store and a memory store.
- We need to add a manifest hooker to the base handler. Basically, we need to explicitly re-invent the `oras.Pull()`.

qweeah:
@shizhMSFT Please assign it to me.

lizMSFT:
Assigned

PR #479, opened 2022-08-03
fix: print digest and media type when unnamed content is skipped

### LINKED ISSUE #415: Print out the digest and media type even when a unnamed content is skipped for pulling
Print out the digest and media type even when a unnamed content is skipped for pulling and verbose option is set.

- unnamed: the to-be-pulled content doesn't have a title annotation(org.opencontainers.image.title).

- skipped: the content won't be fetch during the pull.

- For one digest, the skipped log should be only printed out once.

_Originally posted by @shizhMSFT in https://github.com/oras-project/oras/pull/412#discussion_r903937191_

wangxiaoxuan273:
Add a print statement when unnamed content is skipped for downloading. Short digest and media type will be printed.

Resolves #415 

### DIFF
--- a/cmd/oras/internal/display/print.go
+++ b/cmd/oras/internal/display/print.go
@@ -37,8 +37,8 @@ func Print(a ...any) error {
 // StatusPrinter returns a tracking function for transfer status.
 func StatusPrinter(status string, verbose bool) func(context.Context, ocispec.Descriptor) error {
 	return func(ctx context.Context, desc ocispec.Descriptor) error {
-		name, ok := desc.Annotations[ocispec.AnnotationTitle]
-		if !ok {
+		name := desc.Annotations[ocispec.AnnotationTitle]
+		if name == "" {
 			if !verbose {
 				return nil
 			}
--- a/cmd/oras/internal/display/print.go
+++ b/cmd/oras/internal/display/print.go
@@ -40,6 +40,7 @@ func StatusPrinter(status string, verbose bool) func(context.Context, ocispec.De
 		name := desc.Annotations[ocispec.AnnotationTitle]
 		if name == "" {
 			if !verbose {
+				// no status for unamed content
 				return nil
 			}
 			name = desc.MediaType
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -33,6 +33,8 @@ import (
 	"oras.land/oras/internal/docker"
 )
 
+type void struct{}
+
 type pullOptions struct {
 	option.Common
 	option.Remote
@@ -85,6 +87,8 @@ Example - Pull files with local cache:
 }
 
 func runPull(opts pullOptions) error {
+	set := make(map[string]void)
+	var printed void
 	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
 	if err != nil {
 		return err
@@ -129,9 +133,12 @@ func runPull(opts pullOptions) error {
 					return nil, err
 				}
 				if len(ss) == 0 {
-					err = display.StatusPrinter("Skipped    ", opts.Verbose)(ctx, s)
-					if err != nil {
-						return nil, err
+					if _, exists := set[string(s.Digest)]; !exists {
+						err = display.StatusPrinter("Skipped    ", opts.Verbose)(ctx, s)
+						if err != nil {
+							return nil, err
+						}
+						set[string(s.Digest)] = printed
 					}
 					continue
 				}
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -122,13 +122,17 @@ func runPull(opts pullOptions) error {
 					s.Annotations = make(map[string]string)
 				}
 				s.Annotations[ocispec.AnnotationTitle] = configPath
-			} else if s.Annotations[ocispec.AnnotationTitle] == "" {
+			}
+			if s.Annotations[ocispec.AnnotationTitle] == "" {
 				ss, err := content.Successors(ctx, fetcher, s)
 				if err != nil {
 					return nil, err
 				}
 				if len(ss) == 0 {
-					display.StatusPrinter("Skipped", opts.Verbose)
+					err = display.StatusPrinter("Skipped    ", opts.Verbose)(ctx, s)
+					if err != nil {
+						return nil, err
+					}
 					continue
 				}
 			}
@@ -141,10 +145,10 @@ func runPull(opts pullOptions) error {
 	copyOptions.PreCopy = display.StatusPrinter("Downloading", opts.Verbose)
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
 		name := desc.Annotations[ocispec.AnnotationTitle]
-		if name == "" {
+		pulledEmpty = name == ""
+		if name == "" && !opts.Verbose {
 			return nil
 		}
-		pulledEmpty = false
 		return display.Print("Downloaded ", display.ShortDigest(desc), name)
 	}
 
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -128,6 +128,7 @@ func runPull(opts pullOptions) error {
 					return nil, err
 				}
 				if len(ss) == 0 {
+					display.Print("Skipped ", display.ShortDigest(desc), s.MediaType)
 					continue
 				}
 			}
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -125,6 +125,7 @@ func runPull(opts pullOptions) error {
 				s.Annotations[ocispec.AnnotationTitle] = configPath
 			}
 			if s.Annotations[ocispec.AnnotationTitle] == "" {
+				// skip fetching unamed leaf nodes
 				ss, err := content.Successors(ctx, fetcher, s)
 				if err != nil {
 					return nil, err
@@ -155,6 +156,7 @@ func runPull(opts pullOptions) error {
 			}
 			name = desc.MediaType
 		} else {
+			// no named content pulled
 			pulledEmpty = false
 		}
 		return display.Print("Downloaded ", display.ShortDigest(desc), name)
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -156,7 +156,7 @@ func runPull(opts pullOptions) error {
 			}
 			name = desc.MediaType
 		} else {
-			// no named content pulled
+			// named content downloaded
 			pulledEmpty = false
 		}
 		return display.Print("Downloaded ", display.ShortDigest(desc), name)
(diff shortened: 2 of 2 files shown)


cmd/oras/pull.go
```diff
@@ -121,6 +121,7 @@ func runPull(opts pullOptions) error {
 					return nil, err
 				}
 				if len(ss) == 0 {
+					display.Print("Skipped ", display.ShortDigest(s), s.MediaType)
```
qweeah, line 124:
```suggestion
					display.Print("Skipped    ", display.ShortDigest(s), s.MediaType)			
```

qweeah: And it should only be printed out in verbose mode

qweeah: As is discussed offline, there is a bug here in the `else if` block and we need to remove the `else` otherwise unnamed config content will not be skipped.

Also, @lizMSFT is working on #275 and should modify the `if` block.

cmd/oras/pull.go
```diff
@@ -83,6 +85,8 @@ Example - Pull files with local cache:
 }
 
 func runPull(opts pullOptions) error {
+	set := make(map[string]void)
```
qweeah, line 88: bool defaults to false
```suggestion
	printed := make(map[string]bool)
```

cmd/oras/pull.go
```diff
@@ -115,12 +119,20 @@ func runPull(opts pullOptions) error {
...
 				if err != nil {
 					return nil, err
 				}
 				if len(ss) == 0 {
+					if _, exists := set[string(s.Digest)]; !exists {
+						err = display.StatusPrinter("Skipped    ", opts.Verbose)(ctx, s)
+						if err != nil {
+							return nil, err
+						}
+						set[string(s.Digest)] = printed
```
qweeah, line 134:
```suggestion
						printed[string(s.Digest)] = true
```

cmd/oras/pull.go
```diff
@@ -115,12 +119,20 @@ func runPull(opts pullOptions) error {
...
 				s.Annotations[ocispec.AnnotationTitle] = configPath
-			} else if s.Annotations[ocispec.AnnotationTitle] == "" {
+			}
+			if s.Annotations[ocispec.AnnotationTitle] == "" {
 				ss, err := content.Successors(ctx, fetcher, s)
 				if err != nil {
 					return nil, err
 				}
 				if len(ss) == 0 {
+					if _, exists := set[string(s.Digest)]; !exists {
```
qweeah, line 129:
```suggestion
					if !printed[string(s.Digest)] {
```

qweeah (changes requested):

cmd/oras/pull.go
```diff
@@ -115,12 +116,20 @@ func runPull(opts pullOptions) error {
...
 				if err != nil {
 					return nil, err
 				}
 				if len(ss) == 0 {
+					if !printed[s.Digest.String()] {
+						err = display.StatusPrinter("Skipped    ", opts.Verbose)(ctx, s)
+						if err != nil {
+							return nil, err
+						}
+						printed[s.Digest.String()] = true
```
qweeah, line 131: What if a content is named in Manifest `A` but unnamed in Manifest `B`?

qweeah (changes requested):

cmd/oras/internal/display/print.go
```diff
@@ -36,8 +36,8 @@ func Print(a ...any) error {
 // StatusPrinter returns a tracking function for transfer status.
 func StatusPrinter(status string, verbose bool) func(context.Context, ocispec.Descriptor) error {
 	return func(ctx context.Context, desc ocispec.Descriptor) error {
-		name, ok := desc.Annotations[ocispec.AnnotationTitle]
-		if !ok {
+		name := desc.Annotations[ocispec.AnnotationTitle]
+		if name == "" {
 			if !verbose {
 				return nil
```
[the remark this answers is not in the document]
wangxiaoxuan273: Added the content.

cmd/oras/pull.go
```diff
@@ -115,12 +116,20 @@ func runPull(opts pullOptions) error {
 					s.Annotations = make(map[string]string)
 				}
 				s.Annotations[ocispec.AnnotationTitle] = configPath
-			} else if s.Annotations[ocispec.AnnotationTitle] == "" {
+			}
+			if s.Annotations[ocispec.AnnotationTitle] == "" {
 				ss, err := content.Successors(ctx, fetcher, s)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Added the comments.

cmd/oras/pull.go
```diff
@@ -134,9 +143,13 @@ func runPull(opts pullOptions) error {
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
 		name := desc.Annotations[ocispec.AnnotationTitle]
 		if name == "" {
-			return nil
+			if !opts.Verbose {
+				return nil
+			}
+			name = desc.MediaType
+		} else {
+			pulledEmpty = false
```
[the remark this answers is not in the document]
wangxiaoxuan273: This is a confusing comment.

cmd/oras/pull.go
wangxiaoxuan273: Discussed in the issue.

cmd/oras/pull.go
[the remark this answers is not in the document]
wangxiaoxuan273: changed accordingly.

cmd/oras/pull.go
qweeah: Let's wait for confirm in the issue. Other parts look good to me.

shizhMSFT (commented):
@wangxiaoxuan273 Could you rebase and update this PR as per #481? /cc @qweeah

wangxiaoxuan273:
> @wangxiaoxuan273 Could you rebase and update this PR as per #481? /cc @qweeah

Rebased and updated.

cmd/oras/internal/display/print.go
```diff
@@ -37,9 +37,10 @@ func Print(a ...any) error {
 // StatusPrinter returns a tracking function for transfer status.
 func StatusPrinter(status string, verbose bool) func(context.Context, ocispec.Descriptor) error {
 	return func(ctx context.Context, desc ocispec.Descriptor) error {
-		name, ok := desc.Annotations[ocispec.AnnotationTitle]
-		if !ok {
+		name := desc.Annotations[ocispec.AnnotationTitle]
+		if name == "" {
 			if !verbose {
+				// no status for unamed content
```
shizhMSFT, line 43: Any reason to change the check?

cmd/oras/pull.go
```diff
@@ -85,6 +85,7 @@ Example - Pull files with local cache:
 }
 
 func runPull(opts pullOptions) error {
+	printed := make(map[string]bool)
```
shizhMSFT, line 88: It seems that `printed` is accessed in RW mode concurrently. Should we use `sync.Map` instead?

cmd/oras/pull.go
```diff
@@ -122,12 +123,21 @@ func runPull(opts pullOptions) error {
 					s.Annotations = make(map[string]string)
 				}
 				s.Annotations[ocispec.AnnotationTitle] = configPath
```
shizhMSFT, line 125: What will happen if `configPath` is empty?

cmd/oras/pull.go
```diff
@@ -122,12 +123,21 @@ func runPull(opts pullOptions) error {
...
+			}
+			if s.Annotations[ocispec.AnnotationTitle] == "" {
+				// skip fetching unamed leaf nodes
 				ss, err := content.Successors(ctx, fetcher, s)
 				if err != nil {
 					return nil, err
 				}
 				if len(ss) == 0 {
+					if !printed[s.Digest.String()] {
+						err = display.StatusPrinter("Skipped    ", opts.Verbose)(ctx, s)
```
shizhMSFT, line 135: You may want to update this with `PrintStatus()`.

cmd/oras/internal/display/print.go
wangxiaoxuan273: This is a bug that we fixed. For unnamed content, `ok` is always true but `name` is "". This causes verbose not to work.

cmd/oras/pull.go
wangxiaoxuan273: Billy said he discussed with you and you had decided that we do not need `sync.Map` in this case.

cmd/oras/pull.go
wangxiaoxuan273: Then `s` becomes unnamed.

cmd/oras/pull.go
wangxiaoxuan273: No, `PrintStatus` implementation will cause `verbose` not to work. Billy and I have checked this scenario.

cmd/oras/internal/display/print.go
shizhMSFT: If that's the case, what caused `name` is set to empty string? We need to address that in the first place instead of having a workaround here.

cmd/oras/pull.go
shizhMSFT: That's not true. `FindSuccessors` might be invoked concurrently and thus `printed` may be RW concurrently and it causes race condition.

cmd/oras/pull.go
shizhMSFT: [PrintStatus()](https://github.com/oras-project/oras/blob/9e1d11c51dabb433e904f6c840ded448f8599862/cmd/oras/internal/display/print.go#L52) works with verbose.

If that function does not work, then we need to figure the root cause and fix it.

cmd/oras/internal/display/print.go
wangxiaoxuan273: Name is empty is the definition of "unnamed content", which needs to be skipped when downloading. I think that's Billy's design.

cmd/oras/pull.go
```diff
@@ -141,9 +151,14 @@ func runPull(opts pullOptions) error {
...
 		if name == "" {
-			return nil
+			if !opts.Verbose {
+				return nil
+			}
+			name = desc.MediaType
+		} else {
+			// named content downloaded
+			pulledEmpty = false
 		}
```
shizhMSFT, line 161: It seems that this code block can be replaced with `PrintStatus`. Could you have a check?

cmd/oras/pull.go
wangxiaoxuan273: I'll change it to sync.Map.

cmd/oras/internal/display/print.go
shizhMSFT: If the name is empty by default, then the name should never be set in the first place.

cmd/oras/pull.go
shizhMSFT: That's a bit different. Let's keep it.

wangxiaoxuan273:
Close this pr as the changes are now out of control. Will open another ones to fix the issue.

PR #274, opened 2021-05-17
Enable CLI to Pull Config

shizhMSFT:
Add a new option named `--manifest-config` to `oras pull` so that `oras` can pull the config file as well even if the config is not named. Like `oras push`, file name with media type should be passed to the `--manifest-config` option. The media type is optional and its default value is `application/vnd.unknown.config.v1+json`. The resulted config will be written to the specified file name.

Examples:
```
$ oras pull --manifest-config config.json localhost:5000/hello:latest
Downloaded a948904f2f0f hi.txt
Downloaded 57f840b6073c config.json
Pulled localhost:5000/hello:latest
Digest: sha256:a5d8f401e10aa320bd7ecabbfbabb89b0e59e599c9fdc6587edbffaf232fa4f3
```

```
$ oras pull --manifest-config config.json:application/vnd.unknown.config.v1+json localhost:5000/hello:latest
Downloaded a948904f2f0f hi.txt
Downloaded 57f840b6073c config.json
Pulled localhost:5000/hello:latest
Digest: sha256:a5d8f401e10aa320bd7ecabbfbabb89b0e59e599c9fdc6587edbffaf232fa4f3
```

PR #480, opened 2022-08-04
feat: Support pulling arbitrary manifest config

lizMSFT:
#274 enables the CLI to pull manifest config. However, the user must provide the desired media type of the config or use the default. This PR allows pulling arbitrary manifest config.

If the user does not specify the desired media type, the code will get the 0th indexed node as the manifest config if its parent node is a manifest.
Note: For a manifest, the 0th indexed element is always a manifest config. ref: https://github.com/oras-project/oras-go/blob/6a09a65fcc0b96c3448e988c1727ed154c2388ea/content/graph.go#L58

Examples:
```
$ oras pull --manifest-config config.json localhost:5000/hello:latest
Downloaded  3ad08644ab5a artifact.txt
Downloaded  98df5c495df6 config.json
Pulled localhost:5000/hello:latest
Digest: sha256:8627dfe823698a566e105d961045eafcaa9115b6faf1f625edd85df558421bd5
```

Resolves: #275 

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -101,15 +101,22 @@ func runPull(opts pullOptions) error {
 
 	// Copy Options
 	copyOptions := oras.DefaultCopyOptions
-	configPath, configMediaType := parseFileReference(opts.ManifestConfigRef, oras.MediaTypeUnknownConfig)
+	configPath, configMediaType := parseFileReference(opts.ManifestConfigRef, "")
 	copyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
 		successors, err := content.Successors(ctx, fetcher, desc)
 		if err != nil {
 			return nil, err
 		}
 		var ret []ocispec.Descriptor
-		for _, s := range successors {
-			if s.MediaType == configMediaType {
+		for i, s := range successors {
+			// Check whether the node's MediaType matches the specified MediaType.
+			// If the mediaType is not specified, check whether the parent node
+			// is a manifest.
+			// Note: For a manifest, the 0th indexed element is always a manifest
+			// config.
+			if s.MediaType == configMediaType || (configMediaType == "" && i == 0 &&
+				(desc.MediaType == "application/vnd.docker.distribution.manifest.v2+json" ||
+					desc.MediaType == ocispec.MediaTypeImageManifest)) {
 				// Add annotation for manifest config
 				if s.Annotations == nil {
 					s.Annotations = make(map[string]string)
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -109,11 +109,11 @@ func runPull(opts pullOptions) error {
 		}
 		var ret []ocispec.Descriptor
 		for i, s := range successors {
-			// Check whether the node's MediaType matches the specified MediaType.
-			// If the mediaType is not specified, check whether the parent node
-			// is a manifest.
-			// Note: For a manifest, the 0th indexed element is always a manifest
-			// config.
+			// Save the config when:
+			// 1) MediaType matches, or
+			// 2) MediaType not specified and current node is config.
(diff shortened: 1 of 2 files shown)


cmd/oras/pull.go
```diff
@@ -101,15 +101,22 @@ func runPull(opts pullOptions) error {
...
 		}
 		var ret []ocispec.Descriptor
-		for _, s := range successors {
-			if s.MediaType == configMediaType {
+		for i, s := range successors {
+			// Check whether the node's MediaType matches the specified MediaType.
+			// If the mediaType is not specified, check whether the parent node
+			// is a manifest.
+			// Note: For a manifest, the 0th indexed element is always a manifest
+			// config.
```
qweeah, line 116:
```suggestion
			// Save the config when:
			// 1) MediaType matches;
			// 2) MediaType not specified but current descriptor points to a config.
			// Note: For a manifest, the 0th indexed element is always a manifest
			// config.
```

qweeah (changes requested):
Verified on my side and functionally looks good.

qweeah:
@shizhMSFT Do we need to print warning if: no config media type is specified and there is already a name(title annotation) in the config descriptor?

cmd/oras/pull.go
lizMSFT: Resolved

sajayantony:
Recommend simplifying this to to just `oras pull --config`. I don't believe there is any other object in OCI that is called config.

shizhMSFT:
`--config` is already reserved for auth config.

See
```
  -c, --config stringArray       auth config path
```

shizhMSFT:
@lizMSFT @sajayantony @qweeah There is another option that we can rename `--config` to `--auth-config` and use `--config` for `--manifest-config`.

qweeah:
> @lizMSFT @sajayantony @qweeah There is another option that we can rename `--config` to `--auth-config` and use `--config` for `--manifest-config`.

How about `--cred-file`? Previously we used `--config` to align with the naming of docker config but it's not easy to understand out of the docker context.

We can still use -c for short.

sajayantony:
This is tricky. If you have a config file that’s for auth and config file for manifest. 
Im ok with manifest config or auth config. since —config /auth-config is a global flag and if it was there before then we might not want to break previous non alpha - releases. 
Basically you folks decide. 😁

lizMSFT:
Yes, I am feeling the same. It would be better not change the name for now since it will bring breaking changes and it currently doesn't bring additional value. Besides, consumers will need to adjust accordingly, which might make them confused.

lizMSFT:
Create a new issue for rename `--manifest-config`: https://github.com/oras-project/oras/issues/495

cmd/oras/pull.go
```diff
@@ -101,15 +101,22 @@ func runPull(opts pullOptions) error {
...
-			if s.MediaType == configMediaType {
+		for i, s := range successors {
+			// Save the config when:
+			// 1) MediaType matches, or
+			// 2) MediaType not specified and current node is config.
+			// Note: For a manifest, the 0th indexed element is always a
+			// manifest config.
+			if s.MediaType == configMediaType || (configMediaType == "" && i == 0 &&
+				(desc.MediaType == "application/vnd.docker.distribution.manifest.v2+json" ||
+					desc.MediaType == ocispec.MediaTypeImageManifest)) {
```
shizhMSFT, line 119: Let's make `application/vnd.docker.distribution.manifest.v2+json` a constant and make this Boolean expression as a function.

shizhMSFT:
@sajayantony Let's continue the discussion in #495.

shizhMSFT:
> @shizhMSFT Do we need to print warning if: no config media type is specified and there is already a name(title annotation) in the config descriptor?

No, we don't.

cmd/oras/pull.go
lizMSFT: Made requested change

shizhMSFT (approved):
LGTM

PR #482, opened 2022-08-05
feat: use --artifact-type to change the media type of the config blob

yuehaoliang-microsoft:
Resolve part of #466 

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -40,6 +40,7 @@ type pushOptions struct {
 
 	targetRef         string
 	manifestConfigRef string
+	artifactTpye      string
 }
 
 func pushCmd() *cobra.Command {
@@ -57,6 +58,9 @@ Example - Push file "hi.txt" with the custom "application/vnd.me.hi" media type:
 
 Example - Push multiple files with different media types:
   oras push localhost:5000/hello:latest hi.txt:application/vnd.me.hi bye.txt:application/vnd.me.bye
+
+Example - Push file "hi.txt" with the custom "application/vnd.me.config" artifact type:
+  oras push --artifact-type application/vnd.me.config localhost:5000/hello:latest hi.txt
 
 Example - Push file "hi.txt" with the custom manifest config "config.json" of the custom "application/vnd.me.config" media type:
   oras push --manifest-config config.json:application/vnd.me.config localhost:5000/hello:latest hi.txt
@@ -79,6 +83,7 @@ Example - Push file to the HTTP registry:
 	}
 
 	cmd.Flags().StringVarP(&opts.manifestConfigRef, "manifest-config", "", "", "manifest config file")
+	cmd.Flags().StringVarP(&opts.artifactTpye, "artifact-type", "", "", "artifact type")
 
 	option.ApplyFlags(&opts, cmd.Flags())
 	return cmd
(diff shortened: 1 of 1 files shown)


cmd/oras/push.go
```diff
@@ -58,6 +59,9 @@ Example - Push file "hi.txt" with the custom "application/vnd.me.hi" media type:
 Example - Push multiple files with different media types:
   oras push localhost:5000/hello:latest hi.txt:application/vnd.me.hi bye.txt:application/vnd.me.bye
 
+Example - Push file "hi.txt" with the custom "application/vnd.me.config" artifact type:
```
qweeah, line 62:
```suggestion
Example - Push file "hi.txt" with "application/vnd.me.config" as config type:
```

cmd/oras/push.go
```diff
@@ -79,6 +83,7 @@ Example - Push file to the HTTP registry:
 	}
 
 	cmd.Flags().StringVarP(&opts.manifestConfigRef, "manifest-config", "", "", "manifest config file")
+	cmd.Flags().StringVarP(&opts.artifactTpye, "artifact-type", "", "", "artifact type")
```
qweeah, line 86:
```suggestion
	cmd.Flags().StringVarP(&opts.artifactTpye, "artifact-type", "", "", "media type of config or manifest")
```

cmd/oras/push.go
```diff
@@ -132,7 +137,13 @@ func packManifest(ctx context.Context, store *file.Store, annotations map[string
 	packOpts.ConfigAnnotations = annotations[annotationConfig]
 	packOpts.ManifestAnnotations = annotations[annotationManifest]
 
+	if opts.artifactTpye != "" {
+		packOpts.ConfigMediaType = opts.artifactTpye
+	}
 	if opts.manifestConfigRef != "" {
+		if opts.artifactTpye != "" {
+			fmt.Println("--manifest-config will override --artifact-type")
```
qweeah, line 145: According to the [discuz](https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1658886080869469), we should exit with error so user will not abuse the overriding.

qweeah (changes requested):
The subject sounds more like 'ues artifact type flag to set config media type' to me.

cmd/oras/push.go
```diff
@@ -79,12 +84,17 @@ Example - Push file to the HTTP registry:
 	}
 
 	cmd.Flags().StringVarP(&opts.manifestConfigRef, "manifest-config", "", "", "manifest config file")
+	cmd.Flags().StringVarP(&opts.artifactTpye, "artifact-type", "", "", "media type of config or manifest")
```
qweeah, line 87:
```suggestion
	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "media type of config or manifest")
```

cmd/oras/push.go
```diff
@@ -69,6 +74,9 @@ Example - Push file to the HTTP registry:
 `,
 		Args: cobra.MinimumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if opts.artifactType != "" && opts.manifestConfigRef != "" {
+				return errors.New("--artifact-type and --manifest-config cannot both be provided")
```
binbin-li, line 78: probably you can use `fmt.Errorf()` instead.

yuehaoliang-microsoft: I see in this repo, both errors.New("") and fmt.Errorf("") are used somewhere. However, fmt.Errorf("") is always used along with a formatted string. Since there is no format argument in this case, so maybe we should choose to use errors.New("")?

qweeah: If there is no string interpolation, `errors.New` is preferred.

qweeah (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #483, opened 2022-08-05
feat: push an ORAS artifact manifest type with --artifact-type when no blob is specified

### LINKED ISSUE #466: `--artifact-type` support in `oras push`

yuehaoliang-microsoft:
Resolve #466 

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -17,6 +17,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -40,6 +41,7 @@ type pushOptions struct {
 
 	targetRef         string
 	manifestConfigRef string
+	artifactType      string
 }
 
 func pushCmd() *cobra.Command {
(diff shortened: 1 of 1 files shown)


cmd/oras/push.go
```diff
@@ -127,30 +134,41 @@ func runPush(opts pushOptions) error {
...
-func packManifest(ctx context.Context, store *file.Store, annotations map[string]map[string]string, opts *pushOptions) (ocispec.Descriptor, error) {
+func packManifest(ctx context.Context, store *file.Store, annotations map[string]map[string]string, opts *pushOptions) (manifestDesc ocispec.Descriptor, err error) {
 	var packOpts oras.PackOptions
 	packOpts.ConfigAnnotations = annotations[annotationConfig]
 	packOpts.ManifestAnnotations = annotations[annotationManifest]
 
-	if opts.manifestConfigRef != "" {
-		path, mediatype := parseFileReference(opts.manifestConfigRef, oras.MediaTypeUnknownConfig)
-		desc, err := store.Add(ctx, annotationConfig, mediatype, path)
+	if len(opts.FileRefs) != 0 {
```
qweeah, line 142: Need more comment like
```suggestion
	if len(opts.FileRefs) != 0 {
	// pack OCI artifact
```

cmd/oras/push.go
```diff
@@ -127,30 +134,41 @@ func runPush(opts pushOptions) error {
...
+		descs, err := loadFiles(ctx, store, annotations, opts.FileRefs, opts.Verbose)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		manifestDesc, err = oras.Pack(ctx, store, descs, packOpts)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+	} else {
+		if opts.artifactType == "" {
```
qweeah, line 161:
```suggestion
		if opts.artifactType == "" {
		// pack ORAS artifact
```

qweeah (changes requested):

qweeah:
Which PR is expected to get merged first? This or #482

PR #486, opened 2022-08-09
feat: push an ORAS artifact manifest when no blob is specified

### LINKED ISSUE #466: `--artifact-type` support in `oras push`

yuehaoliang-microsoft:
Resolve #466 

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -158,23 +158,21 @@ func packManifest(ctx context.Context, store *file.Store, annotations map[string
 			desc.Annotations = packOpts.ConfigAnnotations
 			packOpts.ConfigDescriptor = &desc
 		}
-		descs, err := loadFiles(ctx, store, annotations, opts.FileRefs, opts.Verbose)
+		var descs []ocispec.Descriptor
+		descs, err = loadFiles(ctx, store, annotations, opts.FileRefs, opts.Verbose)
 		if err != nil {
 			return ocispec.Descriptor{}, err
 		}
 		manifestDesc, err = oras.Pack(ctx, store, descs, packOpts)
-		if err != nil {
-			return ocispec.Descriptor{}, err
-		}
 	} else {
 		// pack ORAS artifact
 		if opts.artifactType == "" {
 			return ocispec.Descriptor{}, errors.New("artifact type required when no file specified")
 		}
 		manifestDesc, err = oras.PackArtifact(ctx, store, opts.artifactType, nil, oras.PackArtifactOptions{})
-		if err != nil {
-			return ocispec.Descriptor{}, err
-		}
+	}
+	if err != nil {
+		return ocispec.Descriptor{}, err
 	}
 	if err := store.Tag(ctx, manifestDesc, tagStaged); err != nil {
 		return ocispec.Descriptor{}, err
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -167,6 +167,7 @@ func packManifest(ctx context.Context, store *file.Store, annotations map[string
 	} else {
 		// pack ORAS artifact
 		if opts.artifactType == "" {
+			// TODO: this check should be done during pre-run with mandatory annotation checking
 			return ocispec.Descriptor{}, errors.New("artifact type required when no file specified")
 		}
 		manifestDesc, err = oras.PackArtifact(ctx, store, opts.artifactType, nil, oras.PackArtifactOptions{})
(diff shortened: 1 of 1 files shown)


cmd/oras/push.go
```diff
@@ -71,6 +71,9 @@ Example - Push file to the insecure registry:
 
 Example - Push file to the HTTP registry:
   oras push localhost:5000/hello:latest hi.txt --plain-http
+
+Example - Push oras artifact manifest type with no file:
```
qweeah, line 75:
```suggestion
Example - Push an oras artifact manifest with inlined annotation:
oras push localhost:5000/hello:latest --annotation 
```

If there is no manifest annotation or blob(oras)/file layer(oci) provided , oras should exit with error. 

 @junczhuMSFT is working on the annotation part, please talk with him on this.

cmd/oras/push.go
```diff
@@ -136,33 +139,45 @@ func runPush(opts pushOptions) error {
...
+			packOpts.ConfigDescriptor = &desc
+		}
+		descs, err := loadFiles(ctx, store, annotations, opts.FileRefs, opts.Verbose)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		manifestDesc, err = oras.Pack(ctx, store, descs, packOpts)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
```
qweeah, line 168: Nit: this can be merged with the same error checking code in else block. You can put the merged error checking code right after the if-else.

qweeah (changes requested):

cmd/oras/push.go
yuehaoliang-microsoft: @junczhuMSFT Let me add a check for valid arguments.

cmd/oras/push.go
```diff
@@ -136,33 +139,43 @@ func runPush(opts pushOptions) error {
...
-	}
-	descs, err := loadFiles(ctx, store, annotations, opts.FileRefs, opts.Verbose)
-	if err != nil {
-		return ocispec.Descriptor{}, err
+		manifestDesc, err = oras.Pack(ctx, store, descs, packOpts)
+	} else {
+		// pack ORAS artifact
+		if opts.artifactType == "" {
+			return ocispec.Descriptor{}, errors.New("artifact type required when no file specified")
+		}
```
qweeah, line 171:
```suggestion
		if opts.artifactType == "" {
		        // TODO: this check should be done during pre-run with mandatory annotation checking
			return ocispec.Descriptor{}, errors.New("artifact type required when no file specified")
		}
```

qweeah (approved):
LGTM

sajayantony:
@shizhMSFT  @FeynmanZhou  - it would be good to start discussing how the ORAS CLI  and library can start thinking about using the OCI artifact manifest as well. Having it default to ORAS artifact might make it hard to switch to the OCI artifact manifest if that is approved. 

Can we explore what options we have here 
1. Maybe have a config file support in the CLI 
2. Or a flag that that says something like `oras push  {ref} --oras-artifact` 
3. Or consider switching the default and drop the oras-artifact support from the CLI at some point. 

All of these have pros and cons and starting that discussion might be useful before switching push to use the artifact spec as default.

shizhMSFT:
@sajayantony This issue / PR is based on @SteveLasker 's feedback that OCI manifest does not support empty layer array in terms of spec. That's why ORAS artifact manifest becomes the only choice.

sajayantony:
Yes I understand the decision. I’m asking how do we plan to change the default to use the OCI manifest if that would become the preferred way.

Basically consider how the we keep compatibility with ORAS manifest later.

SteveLasker:
@sajayantony , are you asking about the pending OCI manifest changes?

shizhMSFT:
Since we will support OCI manifest & OCI Artifact (https://github.com/opencontainers/image-spec/pull/934), let's postpone this change.

shizhMSFT:
Let's re-open this PR if needed later.

SteveLasker:
The CLI IX is pretty important for the annotations, without blobs scenarios.

shizhMSFT:
Closing again as https://github.com/oras-project/oras-go/issues/271 is planned.

Issue #476, opened 2022-08-03
Command to push a blob to a remote registry


qweeah:
The proposed commands is:

```
Usage:
  oras blob push REG/REPO PATH [flags]

Flags:
  --ca-file string               server certificate authority file for the remote registry
  -d, --debug                    debug mode
  -h, --help                     help for pushing a blob
      --insecure                 allow connections to SSL registry without certs
  -p, --password string          registry password or identity token
      --password-stdin           read password or identity token from stdin
      --plain-http               allow insecure connections to registry without SSL check
  -u, --username string          registry username
  -v, --verbose                  verbose output
```

Related to #459

lizMSFT:
please assign this to me

lizMSFT:
Current UX:
![image](https://user-images.githubusercontent.com/108315938/186065386-ddd9ca35-a834-48e3-98c7-38373ab35eb5.png)
![image](https://user-images.githubusercontent.com/108315938/186065444-aaf059ec-3e82-4a68-9840-d8aa95534860.png)

yizha1:
One minor comment:
- Push a blob to `a` remote registry

Some questions:

- Can we change the mediaType for the blob?
- What is the supported maximum size of a blob? What will happen if the blob size exceeds the maximum value?
- Can we cancel pushing a blob? What will happen if I `ctrl+c` while pushing a large blob?

lizMSFT:
Hi @yizha1,
> * Push a blob to `a` remote registry

Thanks for pointing out, I will change the command message to `Push a blob to a remote registry` 👌

> Can we change the mediaType for the blob?

I'd say no. The blob only has one mediaType `application/octet-stream`, ref: https://docs.docker.com/registry/spec/api/#initiate-blob-upload

> What is the supported maximum size of a blob? What will happen if the blob size exceeds the maximum value?

Currently we do not have a maximum limit on the size of a blob from our ORAS side. But a registry should have some limit on the maximum blob size that it can accept. So if the blob size exceeds the maximum value, the error will be returned from registry side.

> Can we cancel pushing a blob? What will happen if I ctrl+c while pushing a large blob?

Yes, we can cancel pushing a blob by using ctrl+c. If the push operation has not been completely finished yet, then the process will be stopped.

FeynmanZhou:
The overall UX of `oras push` looks intuitive to me. Would it be more exact to update `<name>` to `<registry>`?
[...]

PR #489, opened 2022-08-11
feat: support push a blob to a remote registry

lizMSFT:
Add a command to push a blob to a remote registry

Resolves: https://github.com/oras-project/oras/issues/476


qweeah:
Please mark the command `blob` and `blob push` preview. You may refer to other comamnds like `attach` or `copy`.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,98 @@
...
+		return err
+	}
+
+	fmt.Println("Pushed", opts.targetRef)
+	fmt.Println("Digest:", desc.Digest)
+
+	return nil
+}
+
+func packBlob(ctx context.Context, opts *pushBlobOptions) (ocispec.Descriptor, *os.File, error) {
```
qweeah, line 67: It's definitely not packing & more like `prepareBlob`

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,98 @@
...
+		return err
+	}
+	defer func() {
+		closeErr := fp.Close()
+		if err == nil {
+			err = closeErr
+		}
+	}()
+
+	// push blob
```
qweeah, line 56: use Exist() to avoid unnecessary repushing

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,98 @@
...
+	// prepare blob content
+	desc, fp, err := packBlob(ctx, &opts)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		closeErr := fp.Close()
+		if err == nil {
+			err = closeErr
+		}
```
qweeah, line 53:
```suggestion
		if closeErr := fp.Close(); err == nil {
			err = closeErr
		}
```

qweeah (changes requested):

cmd/oras/blob/push.go
yuehaoliang-microsoft, line 67: It seems ctx is not used in this func.

yuehaoliang-microsoft, line 67: Only opts.FileRef is used in this func, so maybe just pass a filename string is enough instead of pass the entire opts. I think about it because this way, I could reuse the same func for preparing manifest.

yuehaoliang-microsoft: nvm... I think I get why ctx is needed.. This comment can be resolved.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,98 @@
...
+	digest "github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras/cmd/oras/internal/option"
+)
+
+type pushBlobOptions struct {
+	option.Common
+	option.Remote
+
+	FileRef   string
```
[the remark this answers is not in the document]
lizMSFT: I noticed we have `FileRefs []string` in Pusher option struct:
```
type Pusher struct {
	ManifestExportPath     string
	PathValidationDisabled bool
	ManifestAnnotations    string

	FileRefs []string
}
```
However, the above three fields are not necessary for the blob push case. And for now we are only considering push a blob, so we don't need `FileRefs` as a slice.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// prepare blob content
+	desc, fp, err := packBlob(ctx, &opts)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		closeErr := fp.Close()
+		if err == nil {
```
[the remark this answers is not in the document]
lizMSFT: Oops, I forgot to set the returned variable name. err will be used in:
`func pushBlob(opts pushBlobOptions) (err error) {`

cmd/oras/blob/push.go
qweeah: Why, I think it's removable

cmd/oras/blob/push.go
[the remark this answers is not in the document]
qweeah: `FileRef` doesn't need to be public. It has to be public in `pusher` because the option(a bunch of flags and arguments) will be resued by `oras push` and `oras attach` for packing.

cmd/oras/blob/push.go
qweeah, line 67: As discussed offline, this function can be moved to internal package like `func loadFile(path string, mediaType string) (ocispec.Descriptor, io.ReadCloser, error) {}`

cmd/oras/blob/push.go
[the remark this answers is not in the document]
lizMSFT: Got it, changed to unexported(lowercase), Thanks for pointing out!

cmd/oras/blob/push.go
lizMSFT: Removed. thanks!

lizMSFT: I will move `func Prepare(path string, mediaType string) (ocispec.Descriptor, io.ReadCloser, error)` to the internal package so that it could be reused. thanks!

cmd/oras/blob/push.go
lizMSFT: Added. thanks!

cmd/oras/blob/push.go
lizMSFT: I am thinking of moving this function into either `cmd/oras/internal/option/pusher.go` (later this file will be renamed to packer.go) or `cmd/oras/internal/option/common.go`. What do you think?

qweeah: Let's put it in the `internal` with a proper name.

lizMSFT: Discussed offline. Will move func PrepareContent into internal/upload

internal/upload/upload.go
```diff
@@ -0,0 +1,59 @@
...
+	"io"
+	"os"
+
+	digest "github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// PrepareContent opens the given file and generates descriptor from the given
+// file.
+func PrepareContent(path string, mediaType string) (ocispec.Descriptor, io.ReadCloser, error) {
```
qweeah, line 29: Let there be unit tests.

lizMSFT: Added

internal/file/file_test.go
```diff
@@ -0,0 +1,75 @@
...
+	// generate test content
+	tempDir := t.TempDir()
+	dirPath := filepath.Join(tempDir, "testdir")
+	if err := os.MkdirAll(dirPath, 0777); err != nil {
+		t.Fatal("error calling Mkdir(), error =", err)
+	}
+	content := []byte("hello world!")
+	fileName := "test.txt"
+	path := filepath.Join(dirPath, fileName)
+	if err := ioutil.WriteFile(path, content, 0444); err != nil {
```
qweeah, line 41: `ioutil.WriteFile` is deprecated(https://pkg.go.dev/io/ioutil#WriteFile), use `os.WriteFile` instead

internal/file/file_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"oras.land/oras/internal/file"
+)
+
+func TestFile_PrepareContent(t *testing.T) {
+	// generate test content
+	tempDir := t.TempDir()
+	dirPath := filepath.Join(tempDir, "testdir")
+	if err := os.MkdirAll(dirPath, 0777); err != nil {
+		t.Fatal("error calling Mkdir(), error =", err)
+	}
```
qweeah, line 37: Temp directory is controlled by testing package and no need to enforce creation.

internal/file/file_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	blobMediaType := "application/octet-stream"
+	wantDesc := ocispec.Descriptor{
+		MediaType: blobMediaType,
+		Digest:    digest.FromBytes(content),
+		Size:      int64(len(content)),
+	}
+
+	// test PrepareContent
+	gotDesc, rc, err := file.PrepareContent(path, blobMediaType)
```
qweeah, line 53:
```suggestion
	got, rc, err := file.PrepareContent(path, blobMediaType)
```

internal/file/file_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	content := []byte("hello world!")
+	fileName := "test.txt"
+	path := filepath.Join(dirPath, fileName)
+	if err := ioutil.WriteFile(path, content, 0444); err != nil {
+		t.Fatal("error calling WriteFile(), error =", err)
+	}
+
+	blobMediaType := "application/octet-stream"
+	wantDesc := ocispec.Descriptor{
```
qweeah, line 46:
```suggestion
	want := ocispec.Descriptor{
```

internal/file/file_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if err.Error() != expected {
+		t.Fatalf("PrepareContent() error = %v, wantErr %v", err, expected)
+	}
+
+	// test PrepareContent with nonexistent file
+	_, _, err = file.PrepareContent("nonexistent.txt", blobMediaType)
+	expected = "failed to open nonexistent.txt"
+	if !strings.Contains(err.Error(), expected) {
+		t.Fatalf("PrepareContent() error = %v, wantErr %v", err, expected)
+	}
```
qweeah, line 74: Better move those 2 to `TestFile_PrepareContent_err`

qweeah (changes requested):

internal/file/file_test.go
lizMSFT: made requested change.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,97 @@
...
+
+	exists, err := repo.Exists(ctx, desc)
+	if err != nil {
+		return err
+	}
+	if exists {
+		statusPrinter := display.StatusPrinter("Exists   ", opts.Verbose)
+		if err := statusPrinter(ctx, desc); err != nil {
+			return err
+		}
```
qweeah, line 86: use display.PrintStatus, which is newly added.

internal/file/file.go
```diff
@@ -0,0 +1,58 @@
...
+	if _, err = fp.Seek(0, io.SeekStart); err != nil {
+		return ocispec.Descriptor{}, nil, err
+	}
+
+	desc := ocispec.Descriptor{
+		MediaType: mediaType,
+		Digest:    dgst,
+		Size:      fi.Size(),
+	}
+	return desc, fp, nil
```
qweeah, line 57:
```suggestion
	return ocispec.Descriptor{
		MediaType: mediaType,
		Digest:    dgst,
		Size:      fi.Size(),
	}, fp, nil
```

internal/file/file_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// generate test content
+	tempDir := t.TempDir()
+	content := []byte("hello world!")
+	fileName := "test.txt"
+	path := filepath.Join(tempDir, fileName)
+	if err := os.WriteFile(path, content, 0444); err != nil {
+		t.Fatal("error calling WriteFile(), error =", err)
+	}
+
+	blobMediaType := "application/octet-stream"
```
qweeah, line 40: Better to use a mocked  `blobMediaType` here

internal/file/file_test.go
```diff
@@ -0,0 +1,94 @@
...
+	want := ocispec.Descriptor{
+		MediaType: blobMediaType,
+		Digest:    digest.FromBytes(content),
+		Size:      int64(len(content)),
+	}
+
+	// test PrepareContent
+	got, rc, err := file.PrepareContent(path, blobMediaType)
```
qweeah, line 48: Still need to check if we can readout `content` from `rc`

internal/file/file_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestFile_PrepareContent_errOpenFile(t *testing.T) {
+	// generate test content
+	tempDir := t.TempDir()
+	content := []byte("hello world!")
+	fileName := "test.txt"
+	path := filepath.Join(tempDir, fileName)
+	if err := os.WriteFile(path, content, 0444); err != nil {
+		t.Fatal("error calling WriteFile(), error =", err)
+	}
+	blobMediaType := "application/octet-stream"
```
qweeah, line 86: Do we need to mock everything if it's not used? Same for `TestFile_PrepareContent_errMissingFileName`

qweeah (changes requested):

internal/file/file_test.go
lizMSFT: Made requested changes.

internal/file/file_test.go
lizMSFT: Changed to `const blobMediaType = "application/mock-octet-stream"`

internal/file/file_test.go
lizMSFT: Added

internal/file/file_test.go
```diff
@@ -0,0 +1,80 @@
...
+		Size:      int64(len(content)),
+	}
+
+	// test PrepareContent
+	got, rc, err := file.PrepareContent(path, blobMediaType)
+	defer rc.Close()
+	if err != nil {
+		t.Fatal("PrepareContent() error=", err)
+	}
+	_, err = io.ReadAll(rc)
```
qweeah, line 55: Returned `[]bytes` should equals to `content`

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,96 @@
...
+	cmd := &cobra.Command{
+		Use:   "push <name> file [flags]",
+		Short: "[Preview] Push a blob to remote registry",
+		Long: `[Preview] Push a blob to remote registry
+** This command is in preview and under development. **
+
+Example - Push blob "hi.txt":
+  oras blob push localhost:5000/hello hi.txt
+
+Example - Push blob to the insecure registry:
```
qweeah, line 46:
```suggestion
Example - Push blob to without TLS:
```

internal/file/file_test.go
lizMSFT: Made requested change

qweeah (approved):
LGTM

cmd/oras/internal/option/output.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Output option struct.
+type Output struct {
+	OutputDescriptor bool
+	Pretty           bool
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *Output) ApplyFlags(fs *pflag.FlagSet) {
+	fs.BoolVarP(&opts.OutputDescriptor, "descriptor", "", false, "output descriptor")
```
qweeah, line 35: Should split this into two option types: `descriptor` and `pretty`, since manifest can also be prettified.

cmd/oras/internal/option/output.go
```diff
@@ -0,0 +1,52 @@
...
+// OutputContent outputs the content to stdout
+func (opts *Output) OutputContent(content []byte) error {
+	if opts.Pretty {
+		buf := bytes.NewBuffer(nil)
+		if err := json.Indent(buf, content, "", "  "); err != nil {
+			return fmt.Errorf("failed to prettify: %w", err)
+		}
+		buf.WriteByte('\n')
+		content = buf.Bytes()
+	}
```
qweeah, line 48: This should be moved to `pretty.Format(in []byte) (out []byte)` so manifest can also reuse it.

cmd/oras/internal/option/output.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *Output) ApplyFlags(fs *pflag.FlagSet) {
+	fs.BoolVarP(&opts.OutputDescriptor, "descriptor", "", false, "output descriptor")
+	fs.BoolVarP(&opts.Pretty, "pretty", "", false, "output prettified content")
+}
+
+// OutputContent outputs the content to stdout
+func (opts *Output) OutputContent(content []byte) error {
```
qweeah, line 40: I think this func can be removed after adding the pretty option

qweeah (changes requested):

cmd/oras/internal/option/output.go
qweeah: A second thought says no: we need a func like `(opts *Output) Print(w io.Writer, desc ocispec.Descriptor) error` here which
1) checks if `OutputDescriptor` is set and skip for false
2) marshall the desc
3) w.Write

This can be resused for all push and delete commands for manifest and blob

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,125 @@
...
+			return err
+		}
+		err = opts.Output(os.Stdout, bytes)
+		return err
+	}
+
+	if exists {
+		if err := display.PrintStatus(desc, "Exists", opts.Verbose); err != nil {
+			return err
+		}
```
qweeah, line 119:
```suggestion
		return display.PrintStatus(desc, "Exists", opts.Verbose)
```

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// outputs blob's descriptor
+	if opts.OutputDescriptor {
+		bytes, err := opts.Marshal(desc)
+		if err != nil {
+			return err
+		}
+		err = opts.Output(os.Stdout, bytes)
+		return err
```
qweeah, line 113:
```suggestion
		return opts.Output(os.Stdout, bytes)
```

cmd/oras/internal/option/descriptor.go
```diff
@@ -0,0 +1,43 @@
...
+	fs.BoolVarP(&opts.OutputDescriptor, "descriptor", "", false, "output the descriptor")
+}
+
+// Marshal returns the JSON encoding of descriptor.
+func (opts *Descriptor) Marshal(desc ocispec.Descriptor) ([]byte, error) {
+	b, err := json.Marshal(desc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal descriptor: %w", err)
+	}
+	return b, err
```
qweeah, line 42: I think we don't need this function unless we need to add more info in the returned error. The price of adding this function is that you need to write more unit tests.

qweeah (changes requested):
The overall implementation looks good.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,125 @@
...
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func pushBlob(opts pushBlobOptions) (err error) {
+	ctx, _ := opts.SetLoggerLevel()
+
+	if opts.fileRef == "-" && opts.PasswordFromStdin {
+		return errors.New("`-` read file from input and `--password-stdin` read password from input cannot be both used")
+	}
```
qweeah, line 82: This need to be done before read in password

cmd/oras/blob/push.go
lizMSFT: We cannot use `return` here, still need to print some messages in the next few lines:
```
	fmt.Println("Pushed", opts.targetRef)
	fmt.Println("Digest:", desc.Digest)

	return nil
```

cmd/oras/internal/option/descriptor.go
lizMSFT: Discussed offline. `json.Marshal` only returns go routine recover related errors. We still need this function to return descriptor related error message.

internal/file/file.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// PrepareContent prepares the content descriptor from the file path or stdin.
+func PrepareContent(path string, mediaType string) (ocispec.Descriptor, io.ReadCloser, error) {
+	if path == "" {
+		return ocispec.Descriptor{}, nil, fmt.Errorf("missing file name")
+	}
+
+	if path == "-" {
+		content, err := io.ReadAll(os.Stdin)
```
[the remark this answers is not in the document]
shizhMSFT: Alternative, we can keep the functionality of pushing blob from stdin. However, this requires the caller to provide `--size`.

The usage will be something like
```
file=large.bin
size=$(size=$(stat --printf="%s" $file))
digest=sha256:$(shasum -a 256 $file | cut -d' ' -f1)
pv $file | oras blob push --size $size localhost:5000/hello@$digest -
```

internal/file/file.go
```diff
@@ -0,0 +1,72 @@
...
+			Size:      int64(len(content)),
+		}, rc, nil
+	}
+
+	fp, err := os.Open(path)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, fmt.Errorf("failed to open %s: %w", path, err)
+	}
+
+	fi, err := os.Stat(path)
```
shizhMSFT, line 53: Call `Stat()` against `fp` to avoid potential race conditions in the filesystem.
```suggestion
	fi, err := fp.Stat()
```

Alternatively, we should be able to get the file size by `f.Seek(0, os.SEEK_CUR)` after `digest.FromReader(fp)`.

/cc @qweeah for comments.

internal/file/file.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			return ocispec.Descriptor{}, nil, err
+		}
+		rc := io.NopCloser(bytes.NewReader(content))
+
+		return ocispec.Descriptor{
+			MediaType: mediaType,
+			Digest:    digest.FromBytes(content),
+			Size:      int64(len(content)),
+		}, rc, nil
```
shizhMSFT, line 45: You may also want to check @wangxiaoxuan273 's PR: https://github.com/oras-project/oras-go/pull/281

cmd/oras/internal/option/pretty.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// Pretty option struct.
+type Pretty struct {
+	pretty bool
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *Pretty) ApplyFlags(fs *pflag.FlagSet) {
+	fs.BoolVarP(&opts.pretty, "pretty", "", false, "output prettified content")
```
shizhMSFT, line 34: How about
```suggestion
	fs.BoolVarP(&opts.pretty, "pretty", "", false, "prettify JSON objects printed to stdout")
```

cmd/oras/internal/option/descriptor.go
shizhMSFT, line 42:
```suggestion
	return b, nil
```

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,123 @@
...
+func pushBlob(opts pushBlobOptions) (err error) {
+	ctx, _ := opts.SetLoggerLevel()
+
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	// prepare blob content
+	desc, rc, err := file.PrepareContent(opts.fileRef, "application/octet-stream")
```
shizhMSFT, line 89: I think `oras blob push` can take an optional `--media-type` option similar to `oras manifest push`. If `--media-type` does not present, we can set it to `application/octet-stream`.

Otherwise, the output of `--descriptor` will be less useful.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func pushBlob(opts pushBlobOptions) (err error) {
+	ctx, _ := opts.SetLoggerLevel()
+
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
```
shizhMSFT, line 83: `repo.Reference.Reference` should be empty or contains digest only. Otherwise, errors should be returned.

Think about
```
oras blob push localhost:5000/hello:v1 hi.txt
```
This does not look like a good syntax.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,123 @@
...
+	defer rc.Close()
+
+	exists, err := repo.Exists(ctx, desc)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		if err = repo.Push(ctx, desc, rc); err != nil {
+			return err
+		}
```
shizhMSFT, line 102: We need to print status of `repo.Push` if `opts.OutputDescriptor` is not specified.

shizhMSFT (changes requested):

cmd/oras/blob/push.go
lizMSFT: I'm wondering what will be the use of this optional flag `--media-type`? When calling push function, the expected media type is ignored (ref https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L659). This optional `--media-type` option will only exist in the output of `--descriptor`.

cmd/oras/blob/push.go
lizMSFT: This should be fine because the blob will be pushed to repository only. `repo.Reference.Reference` will be ignored when calling function `buildRepositoryBlobUploadURL`
(ref https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L613)
(ref https://github.com/oras-project/oras-go/blob/295b1c32b079ea18d9d49222a4a3d4d6d9c4b9ab/registry/remote/url.go#L86)

internal/file/file.go
qweeah: Should it be `f.Seek(0, os.SEEK_SET)`?

cmd/oras/blob/push.go
lizMSFT: These are the printed status of `repo.Push` we have for now:
```
	if exists {
		if err := display.PrintStatus(desc, "Exists", opts.Verbose); err != nil {
			return err
		}
	}
	fmt.Println("Pushed", opts.targetRef)
	fmt.Println("Digest:", desc.Digest)
```
Do you have any suggestions for the other status messages?

internal/file/file.go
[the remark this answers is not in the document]
lizMSFT: I also feel that reading a large blob from a stream is not safe. The blob could be extremely large and occupies a large amount of resource. I am not sure whether we have many uses cases that uses this approach for pushing blob. If not, we could probably not support it for now. We may also need PM's insight on this. cc @FeynmanZhou @yizha1

internal/file/file.go
lizMSFT: Yes, we can use `fp.Seek(0, io.SeekCurrent)` to get the file size. Thanks for pointing this out.
ref:
```
// Seek whence values.
//
// Deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.
const (
	SEEK_SET int = 0 // seek relative to the origin of the file
	SEEK_CUR int = 1 // seek relative to the current offset
	SEEK_END int = 2 // seek relative to the end
)
```

internal/file/file_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if _, err := tmpfile.Write(content); err != nil {
+		t.Fatal("error calling Write(), error =", err)
+	}
+	if _, err := tmpfile.Seek(0, 0); err != nil {
+		t.Fatal("error calling Seek(), error =", err)
+	}
+
+	oldStdin := os.Stdin
+	defer func() { os.Stdin = oldStdin }() // Restore original Stdin
```
JeyJeyGao, line 86:
```suggestion
	defer func(stdin *os.File) { os.Stdin = stdin }(os.Stdin)
```

cmd/oras/blob/push.go
lizMSFT: Made requested change. Added `--media-type` option. Default returned media type will be `"application/vnd.oci.image.layer.v1.tar"`

cmd/oras/blob/push.go
lizMSFT: Made requested change. Support push blob with specific digest.
```
Example - Push blob "hi.txt" with the specific digest:
  oras blob push localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5 hi.txt
```

cmd/oras/blob/push.go
lizMSFT: Added printed status "Uploading" and "Uploaded"

internal/file/file.go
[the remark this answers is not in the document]
FeynmanZhou: We need to find a trade-off between usability and security when designing this feature. As reading a large blob into memory might bring security concerns, let's remain this feature and support it by adding a required option `size` and specifying a digest.

We can iterate it and make these two flags optional in the next release.

internal/file/file.go
```diff
@@ -0,0 +1,68 @@
...
+
+	digest "github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// PrepareContent prepares the content descriptor from the file path or stdin.
+// Use the input digest and size if they are provided.
+func PrepareContent(path string, mediaType string, dgst digest.Digest, size int64) (ocispec.Descriptor, io.ReadCloser, error) {
+	if path == "" {
+		return ocispec.Descriptor{}, nil, fmt.Errorf("missing file name")
```
Wwwsylvia, line 31: nit: looks like `fmt.Errorf` is not necessary here.

internal/file/file_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if !reflect.DeepEqual(actualContent, content) {
+		t.Errorf("PrepareContent() = %v, want %v", actualContent, content)
+	}
+}
+
+func TestFile_PrepareContent_fromStdin(t *testing.T) {
+	// generate test content
+	content := []byte("hello world!")
+	tmpfile, err := os.CreateTemp("", "test")
```
Wwwsylvia, line 100: Just curious, is there any reason why we don't use `t.TempDir()`?

lizMSFT: Oops. Perhaps I zoned out while writing this sentence:smiling_face_with_tear: Made requested change.

internal/file/file.go
lizMSFT: Do you mean this error will be included when `os.Open()` failed? (e.g., `Error: failed to open : open : The system cannot find the file specified.`) I kinda feel like there is still a little bit different.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,148 @@
...
+
+			return opts.ReadPassword()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return pushBlob(opts)
+		},
+	}
+
+	cmd.Flags().Int64VarP(&opts.size, "size", "", 0, "provide the blob size")
+	cmd.Flags().StringVarP(&opts.mediaType, "media-type", "", ocispec.MediaTypeImageLayer, "specify the returned media type in the descriptor if `--descriptor` is used")
```
shizhMSFT, line 92: @FeynmanZhou @yizha1 What's the default media type for blobs? Currently, it is `ocispec.MediaTypeImageLayer`, which is consistent with `oras push`. We need to review it later.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,148 @@
...
+	ctx, _ := opts.SetLoggerLevel()
+
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	// prepare blob content
+	refDigest := digest.Digest(repo.Reference.Reference)
+	desc, rc, err := file.PrepareContent(opts.fileRef, "application/octet-stream", refDigest, opts.size)
```
shizhMSFT, line 107: Should we use `opts.mediaType` instead of `application/octet-stream`?

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,148 @@
...
+			return err
+		}
+		if err := display.PrintStatus(desc, "Uploaded ", verbose); err != nil {
+			return err
+		}
+	}
+
+	// outputs blob's descriptor
+	if opts.OutputDescriptor {
+		desc.MediaType = opts.mediaType
```
shizhMSFT, line 136: We may want to remove this line corresponding to line 107.

internal/file/file.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	var file *os.File
+	var err error
```
shizhMSFT, line 35: nit: this line can be moved to line 38-39.

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,148 @@
...
+func pushBlob(opts pushBlobOptions) (err error) {
+	ctx, _ := opts.SetLoggerLevel()
+
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	// prepare blob content
+	refDigest := digest.Digest(repo.Reference.Reference)
```
shizhMSFT, line 106: The validity of `refDigest` is never checked. Try `repo.Reference.Digest()`.

internal/file/file.go
```diff
@@ -0,0 +1,68 @@
...
+		dgst, err = digest.FromReader(file)
+		if err != nil {
+			return ocispec.Descriptor{}, nil, err
+		}
+		if _, err = file.Seek(0, io.SeekStart); err != nil {
+			return ocispec.Descriptor{}, nil, err
+		}
+	}
+
+	if size <= 0 {
```
shizhMSFT, line 55: How about an empty blob?

internal/file/file.go
```diff
@@ -0,0 +1,68 @@
...
+	} else {
+		file, err = os.Open(path)
+		if err != nil {
+			return ocispec.Descriptor{}, nil, fmt.Errorf("failed to open %s: %w", path, err)
+		}
+	}
+
+	if dgst == "" {
+		dgst, err = digest.FromReader(file)
+		if err != nil {
```
shizhMSFT, line 47: When `err != nil`, this function is returned without closing the file.

internal/file/file.go
```diff
@@ -0,0 +1,68 @@
...
+			return ocispec.Descriptor{}, nil, err
+		}
+		if _, err = file.Seek(0, io.SeekStart); err != nil {
```
shizhMSFT, line 50: `Seek` does not work with `stdin`. Therefore, we require both the `digest` and the `size` if content is from `stdin`.

shizhMSFT (changes requested):

cmd/oras/blob/push.go
```diff
@@ -0,0 +1,148 @@
...
+  oras blob push --insecure localhost:5000/hello hi.txt
+`,
+		Args: cobra.ExactArgs(2),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			opts.fileRef = args[1]
+			if opts.fileRef == "-" && opts.PasswordFromStdin {
+				return errors.New("`-` read file from input and `--password-stdin` read password from input cannot be both used")
+			}
+			if opts.fileRef == "-" && opts.size == 0 {
```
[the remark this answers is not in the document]
lizMSFT: It should be an empty file if a blob has size 0. And I'm wondering whether we should still allow users to push an empty blob to registry or not.

cmd/oras/blob/push.go
lizMSFT: Yes we could. But eventually the media type will be set to "application/octet-stream" when calling the `blobStore.Push()`, and this `opts.mediaType` will be only used in the returned descriptor.

internal/file/file.go
lizMSFT: Changed to `if size < 0 {}`

internal/file/file.go
lizMSFT: Made requested change:
```
	file, err := os.Open(path)
	if err != nil {
		return ocispec.Descriptor{}, nil, fmt.Errorf("failed to open %s: %w", path, err)
	}
	defer func() {
		if err != nil {
			file.Close()
		}
	}()
```

internal/file/file.go
lizMSFT: Made requested change. Moved the check of size and digest inside the `PrepareContent` function.

cmd/oras/blob/push.go
lizMSFT: Actually `repo.Reference.Reference` has already been validated when calling `repo, err := opts.NewRepository(opts.targetRef, opts.Common)`,

internal/file/file.go
```diff
@@ -0,0 +1,92 @@
...
+	"os"
+
+	digest "github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// PrepareContent prepares the content descriptor from the file path or stdin.
+// Use the input digest and size if they are provided. Will return error if the
+// content is from stdin but the content digest and size are missing.
+func PrepareContent(path string, mediaType string, dgstStr string, size int64) (_ ocispec.Descriptor, _ io.ReadCloser, prepareErr error) {
```
shizhMSFT, line 31: It is too hacky to have `_` for the return values.

internal/file/file.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+
+	if size < 0 {
+		fi, err := file.Stat()
+		if err != nil {
+			return ocispec.Descriptor{}, nil, fmt.Errorf("failed to stat %s: %w", path, err)
+		}
+		size = fi.Size()
+	}
```
shizhMSFT, line 85: If `--size` is provided, we can do size check whether the file size matches `--size` before obtaining the digest.

shizhMSFT (changes requested):

internal/file/file.go
lizMSFT: Do you mean we could validate whether the input size (if provided) matches the actual content size `fi.Size()` before obtaining the digest? Such as:
```
	fi, err := file.Stat()
	if err != nil {
		return ocispec.Descriptor{}, nil, fmt.Errorf("failed to stat %s: %w", path, err)
	}
	if size >= 0 && size != fi.Size() {
		return ocispec.Descriptor{}, nil, fmt.Errorf("input size %d does not match the actual content size %d", size, fi.Size())
	}
```

internal/file/file.go
lizMSFT: Got it. Made requested change.

internal/file/file.go
```diff
@@ -0,0 +1,93 @@
...
+		dgst, err = digest.Parse(dgstStr)
+		if err != nil {
+			return ocispec.Descriptor{}, nil, fmt.Errorf("invalid digest %s: %w", dgstStr, err)
+		}
+	}
+
+	// prepares the content descriptor from stdin
+	if path == "-" {
+		// throw err if size or digest is not provided.
+		if size < 0 || dgst == "" {
```
yuehaoliang-microsoft: I understand that pushing a manifest doesn't require users to input its digest and size, because manifest has a small size and can be loaded into memory.

Preparing the content of the manifest can share the same part of reading from the file and generate the descriptor. After an offline discussion, I'll try to write a separate `PrepareManifestContent()` with unit tests.



@lizMSFT @shizhMSFT

shizhMSFT (approved):
LGTM

PR #490, opened 2022-08-11
fix: add empty check on blobs and manifest annotation

### LINKED ISSUE #467: `attach` should allow packing nil `Blobs` in the artifact manifest
User might want to use annotation to store information in the manifest content.

If no blob and manifest annotation are provided, should exit with error.

jasminetMSFT:
Resolves #467
Signed-off-by: JasmineTang <jasminetang@microsoft.com>


qweeah (changes requested):

cmd/oras/internal/option/pusher.go
```diff
@@ -64,6 +65,16 @@ func (opts *Pusher) LoadManifestAnnotations() (map[string]map[string]string, err
 	return annotations, nil
 }
 
+// ValidateEmpty checks whether blobs or manifest annotation are empty.
+func (opts *Pusher) ValidateEmpty() error {
+	if len(opts.FileRefs) == 0 {
+		if opts.ManifestAnnotations == "" {
```
qweeah, line 71: What if `ManifestAnnotations` is provided but contains no manifest annotation?

qweeah (changes requested):
Still need unit test for `ValidateEmpty` added in the pusher struct.

cmd/oras/internal/option/pusher.go
```diff
@@ -73,3 +93,11 @@ func decodeJSON(filename string, v interface{}) error {
 	defer file.Close()
 	return json.NewDecoder(file).Decode(v)
 }
+
+// validateEmptyManifestAnnotations check whether existing ManifestAnnotations contain no value.
+func validateEmptyManifestAnnotations(annotations map[string]map[string]string) error {
```
qweeah, line 98: We don't need an extra function here. `len(annotations[annotationManifest]) == 0` is enough. No need to check config annotation.

qweeah (changes requested):

qweeah (approved):
LGTM

cmd/oras/internal/option/pusher.go
```diff
@@ -18,13 +18,20 @@ package option
...
 	"os"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/pflag"
 	"oras.land/oras-go/v2/content"
 )
 
+const (
+	annotationConfig   = "$config"
+	annotationManifest = "$manifest"
```
shizhMSFT, line 32: Let's make them public and update the references in `cmd/oras/push.go`.

cmd/oras/internal/option/pusher.go
```diff
@@ -60,10 +67,21 @@ func (opts *Pusher) LoadManifestAnnotations() (map[string]map[string]string, err
...
 	return annotations, nil
 }
 
+// ValidateEmpty checks whether blobs or manifest annotation are empty.
+func (opts *Pusher) ValidateEmpty(args []string) error {
+	if len(args) == 1 && opts.ManifestAnnotations == "" {
+		return errors.New("no blob and manifest annotation are provided")
+	}
+	return nil
+}
```
shizhMSFT, line 83: This logic is `attach` specific. You might want to take it out of the `Pusher` option.

cmd/oras/attach.go
```diff
@@ -164,3 +168,11 @@ func ociToArtifact(desc ocispec.Descriptor) artifactspec.Descriptor {
...
 	}
 }
+
+// validateEmpty checks whether blobs or manifest annotation are empty.
+func validateEmpty(args []string, opts attachOptions) error {
+	if len(args) == 0 && opts.Packer.AnnotationFilePath == "" && len(opts.Packer.ManifestAnnotations) == 0 {
+		return errors.New("no blob and manifest annotation are provided")
+	}
+	return nil
+}
```
shizhMSFT, line 178: Should we move this check to `packer.go` as it only touches packer?

cmd/oras/internal/option/packer.go
```diff
@@ -77,6 +77,9 @@ func (opts *Packer) LoadManifestAnnotations() (annotations map[string]map[string
 		if err = decodeJSON(opts.AnnotationFilePath, &annotations); err != nil {
 			return nil, err
 		}
+		if len(annotations[annotationManifest]) == 0 && len(opts.FileRefs) == 0 {
+			return nil, fmt.Errorf("no blob and manifest annotation file %s is empty", opts.AnnotationFilePath)
+		}
```
shizhMSFT, line 82: This check overlaps with `validateEmpty()` if it is put after loading all the annotations.

jasminetMSFT: `validateEmpty()` will return error only when file reference, `AnnotationFilePath` tag and `ManifestAnnotations` tag are not provided at the same time. However, here it checks for the condition when `AnnotationFilePath `  is not empty but it have a content like `{"{"$manifest":{}}"}` and no file reference provided.

shizhMSFT (approved):
LGTM

PR #499, opened 2022-08-16
fix: add new line to prettify manifest fetch result 

qweeah:
So linux prompt will not be shown right after the prettified result.


shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #502, opened 2022-08-17
refactor: add pre-defined annotation keys into option.Packer

junczhuMSFT:
Defined pre-defined annotation keys in `option.Packer`, and replaced other const defined with the same value.

Updated annotation related example description.


qweeah (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #432, opened 2022-07-01
feat: add copyCmd()

CharingJC:
Resolves https://github.com/oras-project/oras/issues/307

Signed-off-by: Juncheng Zhu <junczhu@microsoft.com>

PR #505, opened 2022-08-18
fix: fix the lookup check for registry-config option

lizMSFT:
This bug was introduced in PR https://github.com/oras-project/oras/pull/432.
`fs.Lookup("config") != nil` was intended to check whether there was an existing config to avoid repeatedly setting the config value. However, we are not able to set the config value if `fs.Lookup("config")` returns nil.
This PR changed the check to `if fs.Lookup("config") == nil {}`


shizhMSFT (approved):
LGTM

PR #508, opened 2022-08-18
docs: add doc comments for the loop of successors

wangxiaoxuan273:
Note that there are comments in line 115-119.
This pr adds additional comments before and inside the loop of successors to describe what the loop does.

Part of issue #415 

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -110,6 +110,7 @@ func runPull(opts pullOptions) error {
 			return nil, err
 		}
 		var ret []ocispec.Descriptor
+		// For each successor s, save its config to Annotations and skip unnamed content.
 		for i, s := range successors {
 			// Save the config when:
 			// 1) MediaType matches, or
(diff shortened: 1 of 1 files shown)


cmd/oras/pull.go
```diff
@@ -110,6 +110,7 @@ func runPull(opts pullOptions) error {
 			return nil, err
 		}
 		var ret []ocispec.Descriptor
+		// For each successor s, save its config to Annotations and skip unnamed content.
```
qweeah, line 113:
```suggestion
		// Iterate all the successors to
		// 1) Add name annotation if configPath is not empty
		// 2) Skip fetching unamed leaf nodes
```

wangxiaoxuan273: Changed accordingly.

wangxiaoxuan273: Changed accordingly.

shizhMSFT (approved):
LGTM

PR #509, opened 2022-08-19
fix: check if configPath is empty before writing to Annotations

wangxiaoxuan273:
This pr adds empty string check before writing configPath to Annotations, so that if a config only specifies the type but not name, oras cli should not fetch the content of this config.

Part of issue #415 

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -111,24 +111,22 @@ func runPull(opts pullOptions) error {
 		}
 		var ret []ocispec.Descriptor
 		// Iterate all the successors to
-		// 1) Add name annotation if configPath is not empty
-		// 2) Skip fetching unamed leaf nodes
+		// 1) Add name annotation to config if configPath is not empty
+		// 2) Skip fetching unnamed leaf nodes
 		for i, s := range successors {
 			// Save the config when:
 			// 1) MediaType matches, or
 			// 2) MediaType not specified and current node is config.
 			// Note: For a manifest, the 0th indexed element is always a
 			// manifest config.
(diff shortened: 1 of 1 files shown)


qweeah (approved):
LGTM

cmd/oras/pull.go
```diff
@@ -120,6 +120,9 @@ func runPull(opts pullOptions) error {
 			// Note: For a manifest, the 0th indexed element is always a
 			// manifest config.
 			if s.MediaType == configMediaType || (configMediaType == "" && i == 0 && isManifestMediaType(desc.MediaType)) {
+				if configPath == "" {
```
shizhMSFT, line 123: What if a config is named but `configPath` is empty? In that case, we should download it, shouldn’t we?

shizhMSFT (changes requested):

shizhMSFT (approved):
LGTM

lizMSFT (approved):
LGTM

qweeah:
Below command should download `demo.confg` and `bar`:

`oras pull localhost:5000/oras:test`

Manifest content of `localhost:5000/oras:test`:

```
{
  "schemaVersion": 2,
  "config": {
    "size": 2,
    "annotations": {
      "org.opencontainers.image.title": "demo.config"
    }
  },
  "layers": [
    {
      "digest": "sha256:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c",
      "size": 4
      "annotations": {
        "org.opencontainers.image.title": "foo"
      }
    },
    {
      "digest": "sha256:7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730",
      "size": 4,
      "annotations": {
        "org.opencontainers.image.title": "bar"
      }
    }
  ]
}
```

PR #510, opened 2022-08-19
feat: print digest and media type when content is skipped using sync.Map

wangxiaoxuan273:
Print digest and media type when content is skipped, and use sync.Map to ensure skipped content is printed only once.

Part of issue #415 

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
@@ -85,6 +86,7 @@ Example - Pull files with local cache:
 }
 
 func runPull(opts pullOptions) error {
+	printed := &sync.Map{}
 	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/oras/pull.go
```diff
@@ -133,6 +135,13 @@ func runPull(opts pullOptions) error {
 				}
 				// Skip s if s is unnamed and has no successors.
 				if len(ss) == 0 {
+					if _, exists := printed.Load(s.Digest.String()); !exists {
```
qweeah, line 138: Should use https://pkg.go.dev/sync#Map.LoadOrStore instead of .Load
```suggestion
					if _, existed:= printed.LoadOrStore(s.Digest.String(), true); !existed {
```

cmd/oras/pull.go
```diff
@@ -133,6 +135,13 @@ func runPull(opts pullOptions) error {
...
+						err = display.PrintStatus(s, "Skipped    ", opts.Verbose)
+						if err != nil {
+							return nil, err
+						}
+						printed.Store(s.Digest.String(), true)
```
qweeah, line 143:
```suggestion
```

qweeah (approved):
LGTM

cmd/oras/pull.go
```diff
@@ -85,6 +86,7 @@ Example - Pull files with local cache:
 }
 
 func runPull(opts pullOptions) error {
+	printed := &sync.Map{}
```
shizhMSFT, line 89: nit:
```suggestion
	var printed sync.Map
```

cmd/oras/pull.go
```diff
@@ -133,6 +135,12 @@ func runPull(opts pullOptions) error {
 				}
 				// Skip s if s is unnamed and has no successors.
 				if len(ss) == 0 {
+					if _, loaded := printed.LoadOrStore(s.Digest.String(), true); !loaded {
+						err = display.PrintStatus(s, "Skipped    ", opts.Verbose)
+						if err != nil {
+							return nil, err
+						}
```
shizhMSFT, line 142: nit:
```suggestion
						if err = display.PrintStatus(s, "Skipped    ", opts.Verbose); err != nil {
							return nil, err
						}
```

shizhMSFT (approved):
LGTM

Issue #475, opened 2022-08-03
Command to fetch a blob from a remote registry


qweeah:
The proposed commands is:

```
Usage:
  oras blob fetch REG/REPO@DIGEST [flags]

Flags:
  --ca-file string               server certificate authority file for the remote registry
  -d, --debug                    debug mode
  --descriptor                   only fetch a descriptor of the blob
  -h, --help                     help for fetching a blob
      --insecure                 allow connections to SSL registry without certs
  -p, --password string          registry password or identity token
      --password-stdin           read password or identity token from stdin
      --plain-http               allow insecure connections to registry without SSL check
  -u, --username string          registry username
  -v, --verbose                  verbose output
  -o, --output string            save the fetched blob to a local file
```

Related to #459

lizMSFT:
please assign this to me

lizMSFT:
Hi @shizhMSFT, should we enable ORAS_CACHE for this fetch blob workflow?

shizhMSFT:
Yes, should be also the same for `oras manifest fetch`. /cc @qweeah

lizMSFT:
- **crane**: fetch a blob UX
  doc: https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_blob.md
  code: https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/blob.go#L26
  
  Example: 
  - `crane blob zoeycr0707.azurecr.io/hello-world:linuxarmv5@sha256:b921b04d0447ddcd82a9220d887cd146f6ef39e20a938ee5e19a90fc3323e030 > blob.tar.gz`
  Will save the blob output to the blob.tar.gz

  - `crane blob zoeycr0707.azurecr.io/hello-world:linuxarmv5@sha256:b921b04d0447ddcd82a9220d887cd146f6ef39e20a938ee5e19a90fc3323e030`
  Will print the output. 
  If the blob is a config, output will be like:
  ```
  PS C:\Users\zoeyli\oras\oras> crane blob zoeycr0707.azurecr.io/hello-world:linuxarmv5@sha256:c0218de6585df06a66d67b25237bdda42137c727c367373a32639710c7a9fa94
[...]

PR #520, opened 2022-08-22
feat: support fetch a blob from a remote registry

lizMSFT:
support fetch a blob from a remote registry

Resolves: #475 

### DIFF
--- a/cmd/oras/blob/fetch.go
+++ b/cmd/oras/blob/fetch.go
@@ -128,36 +128,33 @@ func fetchBlob(opts fetchBlobOptions) (fetchErr error) {
 
 		// outputs blob content if "--output -" is used
 		if opts.outputPath == "-" {
-			if _, err := io.Copy(os.Stdout, rc); err != nil {
-				return err
-			}
-			return nil
-		} else {
-			// save blob content into the local file if the output path is provided
-			file, err := os.Create(opts.outputPath)
-			if err != nil {
-				return err
-			}
-			defer func() {
-				if closeErr := file.Close(); fetchErr == nil {
-					fetchErr = closeErr
-				}
-			}()
+			_, err := io.Copy(os.Stdout, rc)
+			return err
+		}
 
-			if _, err := io.Copy(file, rc); err != nil {
-				return err
+		// save blob content into the local file if the output path is provided
+		file, err := os.Create(opts.outputPath)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err := file.Close(); fetchErr == nil {
+				fetchErr = err
 			}
+		}()
+
+		if _, err := io.Copy(file, rc); err != nil {
+			return err
 		}
 	}
 
 	// outputs blob's descriptor if `--descriptor` is used
 	if opts.OutputDescriptor {
-		descBytes, err := json.Marshal(desc)
+		descJSON, err := json.Marshal(desc)
 		if err != nil {
 			return err
 		}
-		err = opts.Output(os.Stdout, descBytes)
-		if err != nil {
+		if err := opts.Output(os.Stdout, descJSON); err != nil {
 			return err
 		}
 	}
--- a/cmd/oras/blob/fetch.go
+++ b/cmd/oras/blob/fetch.go
@@ -1,0 +1,128 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package blob
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/spf13/cobra"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content/oci"
+
+	"oras.land/oras/cmd/oras/internal/errors"
+	"oras.land/oras/cmd/oras/internal/option"
+	"oras.land/oras/internal/cache"
+	"oras.land/oras/internal/cas"
+)
+
+type fetchBlobOptions struct {
+	option.Common
+	option.Remote
+
+	cacheRoot       string
+	fetchDescriptor bool
+	output          string
+	targetRef       string
+}
+
+func fetchCmd() *cobra.Command {
+	var opts fetchBlobOptions
+	cmd := &cobra.Command{
+		Use:   "fetch <name@digest> [flags]",
+		Short: "[Preview] Fetch a blob from a remote registry",
+		Long: `[Preview] Fetch a blob from a remote registry
+** This command is in preview and under development. **
+
+Example - Fetch raw blob:
+  oras blob fetch localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
+
+Example - Fetch the descriptor of a blob:
+  oras blob fetch --descriptor localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
+
+Example - Fetch blob from the insecure registry:
+  oras blob fetch localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5 --insecure
+`,
+		Args: cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			opts.cacheRoot = os.Getenv("ORAS_CACHE")
+			return opts.ReadPassword()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			if !strings.Contains(opts.targetRef, "@") {
+				return fmt.Errorf("%s: image reference not support, expecting <name@digest>", opts.targetRef)
+			}
+			return fetchBlob(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.output, "output", "o", "", "output directory")
+	cmd.Flags().BoolVarP(&opts.fetchDescriptor, "descriptor", "", false, "fetch a descriptor of the manifest")
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func fetchBlob(opts fetchBlobOptions) (err error) {
+	ctx, _ := opts.SetLoggerLevel()
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
+	}
+
+	if repo.Reference.Reference == "" {
+		return errors.NewErrInvalidReference(repo.Reference)
+	}
+
+	var src oras.Target = cas.BlobTarget{
+		BlobStore: repo.Blobs(),
+	}
+	if opts.cacheRoot != "" {
+		ociStore, err := oci.New(opts.cacheRoot)
+		if err != nil {
+			return err
+		}
+		src = cache.New(src, ociStore)
+	}
+
+	// Fetch and output
+	var content []byte
+	if opts.fetchDescriptor {
+		content, err = cas.FetchDescriptor(ctx, src, opts.targetRef, nil)
+	} else {
+		content, err = cas.FetchBlob(ctx, src, opts.targetRef)
+	}
+	if err != nil {
+		return err
+	}
+
+	if opts.output != "" {
+		if err = os.WriteFile(opts.output, content, 0666); err != nil {
+			return err
+		}
+	}
+
+	printable := utf8.Valid(content)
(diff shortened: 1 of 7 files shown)


cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,138 @@
...
+** This command is in preview and under development. **
+
+Example - Fetch raw blob:
+  oras blob fetch localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
+
+Example - Fetch the descriptor of a blob:
+  oras blob fetch --descriptor localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
+
+Example - Fetch blob from the insecure registry:
+  oras blob fetch localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5 --insecure
```
nima, line 59: nit: keep the flag in the same position (before arguments) in all examples.

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,138 @@
...
+		return errors.New("either `--output` or `--descriptor` must be provided")
+	}
+
+	if opts.outputPath == "-" && opts.OutputDescriptor {
+		return errors.New("`--output -` cannot be used with `--descriptor` at the same time")
+	}
+
+	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	if err != nil {
+		return err
```
nima, line 90: nit: *consider* wrapping this error with more context.

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if repo.Reference.Reference == "" || !strings.Contains(opts.targetRef, "@") {
+		return fmt.Errorf("%s: blob reference not support, expecting <name@digest>", opts.targetRef)
```
nima, line 94: nit: consider "blob reference must be of the form <name>@<digest>"

Don't we have regexes already defined that we can use here, I mean instead of just looking for an @ sign?

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	var src oras.Target = cas.BlobTarget(repo.Blobs())
+	if opts.cacheRoot != "" {
+		ociStore, err := oci.New(opts.cacheRoot)
+		if err != nil {
+			return err
```
nima, line 101: nit: consider adding context to error (wrapping it)

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		err = opts.Output(os.Stdout, desc)
+		if err != nil {
+			return err
+		}
+	}
+
+	return err
```
nima, line 137: nit: return nil, to save reader from having to verify intent

nima (commented):
Have only reviewed one file so far, but publishing in advance 😃

cmd/oras/internal/option/descriptor.go
```diff
@@ -0,0 +1,41 @@
...
+	fs.BoolVarP(&opts.OutputDescriptor, "descriptor", "", false, "output the descriptor")
+}
+
+// Marshal returns the JSON encoding of descriptor.
+func (opts *Descriptor) Marshal(desc ocispec.Descriptor) ([]byte, error) {
+	b, err := json.Marshal(desc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal descriptor: %w", err)
+	}
+	return b, err
```
nima, line 40: nit: `return b, nil`

internal/cas/fetch.go
```diff
@@ -69,3 +69,28 @@ func FetchManifest(ctx context.Context, target oras.Target, reference string, p
...
+func BlobTarget(b registry.BlobStore) oras.Target {
+	return blobTarget{b}
+}
+
+type blobTarget struct {
+	registry.BlobStore
+}
+
+func (blobTarget) Tag(ctx context.Context, desc ocispec.Descriptor, reference string) error {
+	return nil
```
nima, line 95: This is probably obvious to most, but to me it's confusing... maybe a comment explaining why it's a do-nothing function?

nima (commented):
Left some comments from the perspective of someone new to the project.

internal/cas/fetch.go
lizMSFT: Hi @nima, thanks for your review. 

Basically this `Tag` function is an implementation of `Target.Tag()` in order to make blobTarget as a ORAS Target. In func `FetchBlob`, we will call `FetchReference` to get the blob content. However, if we use Repository as the `Target`, it will go to the manifest workflow (ref https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L257) which will cause the wrong result. So we have to use a `BlobStore` here as the Target. But  `BlobStore` does not have a `Tag` method (ref https://github.com/oras-project/oras-go/blob/main/registry/repository.go#L68). Therefore, this no-op method is added in order to make `blobTarget` as a `Target`.

I just added some comments in the code. Thanks for pointing this out.

internal/cas/fetch.go
```diff
@@ -69,3 +69,31 @@ func FetchManifest(ctx context.Context, target oras.Target, reference string, p
...
+	if err != nil {
+		return nil, err
+	}
+	defer rc.Close()
+	return content.ReadAll(rc, desc)
+}
+
+// BlobTarget returns a ORAS Target with a no-op Tag method wrapping the
+// provided blob store b.
+func BlobTarget(b registry.BlobStore) oras.Target {
```
Wwwsylvia, line 86: You may leverage [`oras.ReadOnlyTarget`](https://github.com/oras-project/oras-go/blob/main/target.go#L34).

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,144 @@
...
+	targetRef  string
+}
+
+func fetchCmd() *cobra.Command {
+	var opts fetchBlobOptions
+	cmd := &cobra.Command{
+		Use:   "fetch <name@digest> [flags]",
+		Short: "[Preview] Fetch a blob from a remote registry",
+		Long: `[Preview] Fetch a blob from a remote registry
+** This command is in preview and under development. **
```
shizhMSFT, line 50: nit: missing an empty line above this line.

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,144 @@
...
+			opts.cacheRoot = os.Getenv("ORAS_CACHE")
+			return opts.ReadPassword()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.targetRef = args[0]
+			return fetchBlob(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "output directory")
```
shizhMSFT, line 78:
```suggestion
	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "output file path")
```

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,144 @@
...
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func fetchBlob(opts fetchBlobOptions) (err error) {
```
yuehaoliang-microsoft, line 83: How about moving flag and argument check to line 68 PreRun?

internal/cas/fetch.go
```diff
@@ -69,3 +69,31 @@ func FetchManifest(ctx context.Context, target oras.Target, reference string, p
 	defer rc.Close()
 	return content.ReadAll(rc, desc)
 }
+
+// FetchBlob fetches the blob content of reference from blob store.
+func FetchBlob(ctx context.Context, blob oras.Target, reference string) ([]byte, error) {
```
Wwwsylvia, line 74: Let's use [oras.Fetch](https://github.com/oras-project/oras-go/blob/main/content.go#L119) and [oras.FetchBytes](https://github.com/oras-project/oras-go/blob/main/content.go#L169)! 😀

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// outputs blob content if "--output -" is used
+	if opts.outputPath == "-" {
+		_, err = os.Stdout.Write(content)
+		return err
+	}
+
+	// save blob content into the local file if the output path is provided
+	if opts.outputPath != "" {
+		if err = os.WriteFile(opts.outputPath, content, 0666); err != nil {
```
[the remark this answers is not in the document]
lizMSFT: Made requested change. Thanks!

internal/cache/target.go
```diff
@@ -117,26 +117,36 @@ type referenceTarget struct {
...
-// FetchReference fetches the content identified by the reference from the
-// remote and cache the fetched content.
+// FetchReference fetches the content identified by the reference from the cache
+// if exists, otherwise fetches the content from the remote and then cache the
+// fetched content.
 // Cached content will only be read via Fetch, FetchReference will always fetch
 // From origin.
 func (t *referenceTarget) FetchReference(ctx context.Context, reference string) (ocispec.Descriptor, io.ReadCloser, error) {
-	target, rc, err := t.ReferenceFetcher.FetchReference(ctx, reference)
+	desc, err := t.Resolve(ctx, reference)
```
qweeah, line 126:
```suggestion
	desc, err := t.cache.Resolve(ctx, reference)
```

internal/cache/target.go
```diff
@@ -117,26 +117,36 @@ type referenceTarget struct {
...
 	if err != nil {
```
qweeah, line 127: And `errdef.ErrNotFound` should be regarded as not cached and fetch from origin.

cmd/oras/blob/fetch.go
lizMSFT: Made requested change.

internal/cache/target.go
lizMSFT: Discussed offline, will still use `t.ReferenceFetcher.FetchReference(ctx, reference)`

cmd/oras/blob/fetch.go
lizMSFT: Thanks for pointing out. The error message returned from previous function should be enough for now. We could wrap this error with more context later if needed

cmd/oras/blob/fetch.go
lizMSFT: same as above

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,158 @@
...
+	if opts.cacheRoot != "" {
+		ociStore, err := oci.New(opts.cacheRoot)
+		if err != nil {
+			return err
+		}
+		src = cache.New(src, ociStore)
+	}
+
+	// fetch blob
+	desc, rc, err := oras.Fetch(ctx, src, opts.targetRef, oras.FetchOptions{})
```
shizhMSFT, line 114:
```suggestion
	desc, rc, err := oras.Fetch(ctx, src, opts.targetRef, oras.DefaultFetchOptions)
```

shizhMSFT, line 114: It seems the `rc` is never verified against the `desc`.

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,158 @@
...
+	if opts.outputPath == "-" {
+		if _, err := io.Copy(os.Stdout, rc); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// save blob content into the local file if the output path is provided
+	if opts.outputPath != "" {
+		file, err := os.OpenFile(opts.outputPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.ModePerm)
```
shizhMSFT, line 130: Should we use [os.Create](https://pkg.go.dev/os#Create)?

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+		}()
+
+		if _, err := io.Copy(file, rc); err != nil {
+			return err
+		}
+	}
+
+	// outputs blob's descriptor if `--descriptor` is used
+	if opts.OutputDescriptor {
```
shizhMSFT, line 146: What if we just want the descriptor? We just need to call `oras.Resolve()` instead of fetching the full blob.

cmd/oras/blob/fetch.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// save blob content into the local file if the output path is provided
+	if opts.outputPath != "" {
+		file, err := os.OpenFile(opts.outputPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.ModePerm)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if closeErr := file.Close(); err == nil {
+				err = closeErr
```
shizhMSFT, line 136: This `err` refers to the `err` at line 130 instead of line 92.

shizhMSFT (changes requested):

cmd/oras/blob/fetch.go
lizMSFT: made requested change

cmd/oras/blob/fetch.go
lizMSFT: will create an issue to track this

shizhMSFT (approved):
LGTM

Issue #470, opened 2022-08-03
Command to list repositories in a remote registry


qweeah:
The proposed command is:

```
Usage:
  oras repository list REGISTRY [flags]

Flags:
  --ca-file string           server certificate authority file for the remote registry
  -d, --debug                    debug mode
  -h, --help                     help for list
      --insecure                 allow connections to SSL registry without certs
  -p, --password string          registry password or identity token
      --password-stdin           read password or identity token from stdin
      --plain-http               allow insecure connections to registry without SSL check
  -u, --username string          registry username
  -v, --verbose                  verbose output
```

The output should be a list of repositories with one name per line.

Related to #459

junczhuMSFT:
I am interest in this, please assign this issue to me

sajayantony:
Since other targets don't have repositories maybe cranes approach of calling it catalog makes more sense to me since it also matches with the API. Also are there any other commands under the repository group? If there are then `repository list` makes sense.

FeynmanZhou:
@sajayantony I see `repository` is more common than `catalog`, I vote for naming `repository` here. There will be an another command `list-tags` under the `repository` group.

@qweeah @junczhuMSFT 
Let me give a detailed example for this CLI UX:

``
$ oras repository list samplecr.cncf.io
acr/hello-world
library/alpine
net-monitor
sample/hello-world
 ```

junczhuMSFT:
> @sajayantony
> 
> > Since other targets don't have repositories maybe cranes approach of calling it catalog makes more sense to me since it also matches with the API
> 
> I see `repository` is more common than `catalog` on the registry side like DockerHub, ACR, ECR, etc. so I vote for naming `repository` here.
> 
> > Also are there any other commands under the repository group?
> 
> There will be another command `list-tags` under the `repository` group.
> 
> @qweeah @junczhuMSFT Let me give a detailed example for this CLI UX:
> 
> ```
> $ oras repository list samplecr.cncf.io
> acr/hello-world
> library/alpine
> net-monitor
> sample/hello-world
> ```
> 
> Shall we provide an alias `-l` for `list` in this command?

provided an alias -l for list in the latest #525

junczhuMSFT:
Shall we provide "first" and "skip" flags? @FeynmanZhou 
There is an azure cli feature, for example, users can skip the first 3,000 records and return the first 1,000 records after those records skipped

[...]

PR #525, opened 2022-08-24
feat: command to list repositories in a remote registry

junczhuMSFT:
Resolve: #470


cmd/oras/repository/list.go
```diff
@@ -0,0 +1,69 @@
...
+			return listRepository(opts)
+		},
+	}
+
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func listRepository(opts repositoryOptions) error {
+	ctx, _ := opts.SetLoggerLevel()
```
sajayantony, line 53: This might be a personal preference so its non-blocking. I think most of the lines are probably self-explanatory so maybe you can remove the comments in the function. Just makes code easier to read :)

junczhuMSFT: I will improve the piece of code and remove the comments 😊

cmd/oras/repository/cmd.go
```diff
@@ -0,0 +1,30 @@
...
+
+package repository
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "repository [list]",
```
qweeah, line 22:
```suggestion
		Use:   "repository [command]",
```
Guess you are referring to oras manifest fetch but I made a mistake there, `command` should be used instead.

cmd/oras/repository/list.go
```diff
@@ -35,8 +35,9 @@ type repositoryOptions struct {
 func listCmd() *cobra.Command {
 	var opts repositoryOptions
 	cmd := &cobra.Command{
-		Use:   "list REGISTRY [flags]",
-		Short: "[Preview] List the repositories under the registry",
+		Use:     "list REGISTRY [flags]",
+		Aliases: []string{"l"},
+		Short:   "[Preview] List the repositories under the registry",
```
sajayantony, line 40: Would have expected the alias to be ls. Never seen l used before.

junczhuMSFT: Well Noted.

cmd/oras/repository/list.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if listErr := reg.Repositories(ctx, opts.last, func(repos []string) error {
+		for _, repo := range repos {
+			fmt.Println(repo)
+		}
+		return nil
+	}); listErr != nil {
+		return listErr
+	}
+	return nil
```
shizhMSFT, line 75: This can be simplified as
```suggestion
	return reg.Repositories(ctx, opts.last, func(repos []string) error {
		for _, repo := range repos {
			fmt.Println(repo)
		}
		return nil
	}
```

@junczhuMSFT Could you also apply this suggestion to https://github.com/oras-project/oras/blob/efd765928b9adfdc0df27cabd8ef6b9138d192b9/cmd/oras/repository/show-tags.go#L64?

junczhuMSFT: Well Noted

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #469, opened 2022-08-03
Command to show tags in a repository


qweeah:
The proposed command is:

```
Usage:
  oras repository show-tags REPO [flags]

Flags:
  --ca-file string           server certificate authority file for the remote registry
  -d, --debug                    debug mode
  -h, --help                     help for showing tags
      --insecure                 allow connections to SSL registry without certs
  -p, --password string          registry password or identity token
      --password-stdin           read password or identity token from stdin
      --plain-http               allow insecure connections to registry without SSL check
  -u, --username string          registry username
  -v, --verbose                  verbose output
```

The output should be a list of tags with one tag per line.

Related to #459

junczhuMSFT:
I am interest in this, please assign this issue to me

sajayantony:
Recommend keeping this consistent and using `list-tags` instead of `show-tags` if we want to keep this command under repository. 

Why don't we move these directly under oras and have `oras list-tags` and `oras list-repositories` or `oras catalog` - 

Basically this is in continuation of the discussion for #470

FeynmanZhou:
@oras-project/oras-maintainers 

IMO, _show tags_ defaults to the repository level, which means it is not necessary to add a noun (root command) `repository` here. 

Considering we will provide a new group `tag` in ORAS CLI 0.15, I would suggest putting _show tags_ under the `tag` group directly and using `tag list` like the following sample CLI UX. This looks more intuitive to me: 

```
$ oras tag list library/alpine
3.10.0
3.10.1
3.10.2
3.10.3
```

sajayantony:
How would you differentiate the adding tag command?

junczhu:
For your information, here is the sample output of this command:

![image](https://user-images.githubusercontent.com/74894646/188592900-7d469c47-d79b-44ce-a5ea-d81d388e9dee.png)

/cc @FeynmanZhou

PR #526, opened 2022-08-24
feat: command to show tags in a repository

junczhuMSFT:
Resolve: #469


cmd/oras/repository/show-tags.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return err
+	}
+	repo.Tags(ctx, "", func(tags []string) error {
+		for _, tag := range tags {
+			fmt.Println(tag)
+		}
+		return nil
+	})
+	// list all tags
```
sajayantony, line 64: Can possibly remove comment since the function name and loop is already clear?

junczhuMSFT: Well Noted

shizhMSFT (approved):
LGTM

PR #538, opened 2022-09-07
refactor: use read only target to replace oras.Target for the target storage with caching

lizMSFT:
Use `oras.ReadOnlyTarget` instead of `oras.Target` for the target storage with caching


shizhMSFT (approved):
LGTM

PR #539, opened 2022-09-07
fix: missing status for pull

### LINKED ISSUE #527: UI bug: pull status missing if two files have identical content
Suppose we have below artifact store in the repository `localhost:5000/oras:debug`
```
               +----------+     +--------+
               | Manifest +-----> config |
               +--+----+--+     +--------+
                  |    |
Annotation Title:foo  Annotation Title:bar
                  |    |
              +---v----v---+
              | Layer Blob |
              +------------+
```
Pulling the below artifact with command `oras pull localhost:5000/oras:debug -v`, the log will be like
```
Skipped     e3b0c44298fc application/vnd.unknown.config.v1+json
Downloading b5bb9d8014a0 bar
Downloaded  b5bb9d8014a0 bar
Downloading 87dd1ad58716 application/vnd.oci.image.manifest.v1+json
Downloaded  87dd1ad58716 application/vnd.oci.image.manifest.v1+json
Pulled localhost:5000/oras:debug
Digest: sha256:87dd1ad58716fb216620b4d28ead9b7d68911e0339cca658cfc2b371c27a0d40
```
Although indeed `foo` and `bar` are both pulled, in the UI there is no hit for downloading `foo`.

wangxiaoxuan273:
Resolves #527 

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -156,7 +156,7 @@ func runPull(opts pullOptions) error {
 	pulledEmpty := true
 	copyOptions.PreCopy = display.StatusPrinter("Downloading", opts.Verbose)
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
-		// restore named successors, if applicable
+		// restore named but deduplicated successor nodes
 		successors, err := content.Successors(ctx, dst, desc)
 		if err != nil {
 			return err
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -162,7 +162,7 @@ func runPull(opts pullOptions) error {
 			return err
 		}
 		for _, s := range ss {
-			if desc.Annotations[ocispec.AnnotationTitle] != "" {
+			if _, ok := s.Annotations[ocispec.AnnotationTitle]; ok {
 				if _, loaded := printed.LoadOrStore(s.Digest.String()+s.Annotations[ocispec.AnnotationTitle], true); !loaded {
 					if err = display.PrintStatus(s, "Restored   ", opts.Verbose); err != nil {
 						return err
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -135,7 +135,7 @@ func runPull(opts pullOptions) error {
 				}
 				// Skip s if s is unnamed and has no successors.
 				if len(ss) == 0 {
-					if _, loaded := printed.LoadOrStore(display.Name(s), true); !loaded {
+					if _, loaded := printed.LoadOrStore(desc.Digest.String()+desc.Annotations[ocispec.AnnotationTitle], true); !loaded {
 						if err = display.PrintStatus(s, "Skipped    ", opts.Verbose); err != nil {
 							return nil, err
 						}
@@ -162,9 +162,8 @@ func runPull(opts pullOptions) error {
 			return err
 		}
 		for _, s := range ss {
-			exists, _ := dst.Exists(ctx, s)
-			if exists {
-				if _, loaded := printed.LoadOrStore(display.Name(s), true); !loaded {
+			if desc.Annotations[ocispec.AnnotationTitle] != "" {
+				if _, loaded := printed.LoadOrStore(s.Digest.String()+s.Annotations[ocispec.AnnotationTitle], true); !loaded {
 					if err = display.PrintStatus(s, "Restored   ", opts.Verbose); err != nil {
 						return err
 					}
@@ -181,7 +180,7 @@ func runPull(opts pullOptions) error {
 			// named content downloaded
 			pulledEmpty = false
 		}
-		printed.LoadOrStore(display.Name(desc), true)
+		printed.Store(desc.Digest.String()+desc.Annotations[ocispec.AnnotationTitle], true)
 		return display.Print("Downloaded ", display.ShortDigest(desc), name)
 	}
 
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -135,7 +135,7 @@ func runPull(opts pullOptions) error {
 				}
 				// Skip s if s is unnamed and has no successors.
 				if len(ss) == 0 {
-					if _, loaded := printed.LoadOrStore(s.Digest.String(), true); !loaded {
+					if _, loaded := printed.LoadOrStore(display.Name(s), true); !loaded {
 						if err = display.PrintStatus(s, "Skipped    ", opts.Verbose); err != nil {
 							return nil, err
 						}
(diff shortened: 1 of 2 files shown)


cmd/oras/pull.go
```diff
@@ -135,7 +135,7 @@ func runPull(opts pullOptions) error {
 				}
 				// Skip s if s is unnamed and has no successors.
 				if len(ss) == 0 {
-					if _, loaded := printed.LoadOrStore(s.Digest.String(), true); !loaded {
+					if _, loaded := printed.LoadOrStore(display.Name(s), true); !loaded {
```
qweeah, line 138: @wangxiaoxuan273 Will `desc.Digest.String() + desc.Annotations[ocispec.AnnotationTitle]` work?

cmd/oras/pull.go
```diff
@@ -161,14 +181,10 @@ func runPull(opts pullOptions) error {
 			// named content downloaded
 			pulledEmpty = false
 		}
+		printed.LoadOrStore(display.Name(desc), true)
```
qweeah, line 184: What is the purpose of this line? Why the returned values are ignored?

cmd/oras/pull.go
```diff
@@ -148,9 +148,29 @@ func runPull(opts pullOptions) error {
...
+		}
+		for _, s := range ss {
+			exists, _ := dst.Exists(ctx, s)
+			if exists {
+				if _, loaded := printed.LoadOrStore(display.Name(s), true); !loaded {
+					if err = display.PrintStatus(s, "Restored   ", opts.Verbose); err != nil {
+						return err
+					}
+				}
+			}
```
qweeah, line 172: The missing else block here means: a manifest is pulled but one of its successor is not, which is a big SDK error and should be handled.

cmd/oras/pull.go
qweeah: Should use https://pkg.go.dev/sync#Map.Store here

cmd/oras/pull.go
```diff
@@ -148,9 +148,29 @@ func runPull(opts pullOptions) error {
...
 	pulledEmpty := true
 	copyOptions.PreCopy = display.StatusPrinter("Downloading", opts.Verbose)
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
+		// restore successors, if applicable
+		ss, err := content.Successors(ctx, dst, desc)
+		if err != nil {
+			return err
+		}
+		for _, s := range ss {
+			exists, _ := dst.Exists(ctx, s)
```
qweeah, line 165: if `desc.Annotations[ocispec.AnnotationTitle]` is not presented, then `desc` won't be downloaded to `dst`. I think that should be used here instead of `.Exists`

cmd/oras/pull.go
wangxiaoxuan273: Yes it worked. Replaced with `desc.Digest.String() + desc.Annotations[ocispec.AnnotationTitle]`.

cmd/oras/pull.go
wangxiaoxuan273: Changed to `sync.Map.Store`

cmd/oras/pull.go
wangxiaoxuan273: This `exists` check is removed per the next comment, so this comment is resolved.

cmd/oras/pull.go
wangxiaoxuan273: replaced the `exists` check with `_, ok := s.Annotations[ocispec.AnnotationTitle]`.

cmd/oras/pull.go
```diff
@@ -135,7 +135,7 @@ func runPull(opts pullOptions) error {
 				}
 				// Skip s if s is unnamed and has no successors.
 				if len(ss) == 0 {
-					if _, loaded := printed.LoadOrStore(s.Digest.String(), true); !loaded {
+					if _, loaded := printed.LoadOrStore(s.Digest.String()+s.Annotations[ocispec.AnnotationTitle], true); !loaded {
```
qweeah, line 138: let's make a private function for this with `s` as input

cmd/oras/pull.go
```diff
@@ -148,9 +148,28 @@ func runPull(opts pullOptions) error {
...
+
 	pulledEmpty := true
 	copyOptions.PreCopy = display.StatusPrinter("Downloading", opts.Verbose)
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
+		// restore named successors, if applicable
+		successors, err := content.Successors(ctx, dst, desc)
+		if err != nil {
+			return err
+		}
+		for _, successor := range successors {
```
qweeah, line 164: Nit: `s` is enough
```suggestion
		for _, s:= range successors {
```

wangxiaoxuan273: renamed it to `s`.

cmd/oras/pull.go
wangxiaoxuan273: Added a function `generateKey(desc ocispec.Descriptor)`.

qweeah (approved):
LGTM

cmd/oras/pull.go
```diff
@@ -148,9 +148,28 @@ func runPull(opts pullOptions) error {
...
 
+	ctx, _ := opts.SetLoggerLevel()
+	var dst = file.New(opts.Output)
+	dst.AllowPathTraversalOnWrite = opts.PathTraversal
+	dst.DisableOverwrite = opts.KeepOldFiles
+
 	pulledEmpty := true
 	copyOptions.PreCopy = display.StatusPrinter("Downloading", opts.Verbose)
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
+		// restore named successors, if applicable
```
qweeah, line 159:
```suggestion
		// restore named but deduplicated successor nodes
```

cmd/oras/pull.go
```diff
@@ -185,3 +200,7 @@ func runPull(opts pullOptions) error {
 func isManifestMediaType(mediaType string) bool {
 	return mediaType == docker.MediaTypeManifest || mediaType == ocispec.MediaTypeImageManifest
 }
+
+func generateKey(desc ocispec.Descriptor) string {
```
shizhMSFT, line 204: Can we have a more meaningful name other than `generateKey`? Meanwhile, can we document this method so that others can understand easily?

cmd/oras/pull.go
```diff
@@ -148,9 +148,28 @@ func runPull(opts pullOptions) error {
...
 	copyOptions.PreCopy = display.StatusPrinter("Downloading", opts.Verbose)
 	copyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
+		// restore named but deduplicated successor nodes
+		successors, err := content.Successors(ctx, dst, desc)
+		if err != nil {
+			return err
+		}
+		for _, s := range successors {
+			if _, ok := s.Annotations[ocispec.AnnotationTitle]; ok {
+				if _, ok := printed.Load(generateKey(s)); !ok {
```
shizhMSFT, line 166: Curious on the reason why we use `Load` over `LoadOrStore`.

@wangxiaoxuan273 @qweeah Could you explain a bit?

qweeah: Printed is for deduplication. File target has already done the deduplication since it won't allow duplidated filename, not mentioning duplicated digest+filename.

cmd/oras/pull.go
qweeah: How about `namedDigest`?

wangxiaoxuan273: Changed to name` generateDeduplicatedMapKey` and added a doc comment.

shizhMSFT (approved):
LGTM

cmd/oras/pull.go
wangxiaoxuan273: Changed to name `generateContentKey` and added a doc comment.

cmd/oras/pull.go
wangxiaoxuan273: Changed back to `LoadOrStore`.

PR #541, opened 2022-09-07
feat: enhance manifest fetch

### LINKED ISSUE #531: Enhance `manifet fetch`
1. Refactor the implementation with `oras.Fetch` and `oras.FetchBytes` in oras-go
2. Add below UX
```
# Fetch
# add output flag
oras manifest fetch <name> --pretty > <path>                     # manifest prettified to file via stdout
oras manifest fetch <name> --output <path> --pretty              # manifest raw to file
oras manifest fetch <name> --output - --pretty > <path>          # manifest prettified to file
oras manifest fetch <name> --output <path> --pretty --descriptor # descriptor prettified
# --output - cannot be used with --descriptor at the same time
```

yuehaoliang-microsoft:
Resolve #531 


cmd/oras/manifest/fetch.go
```diff
@@ -72,46 +79,88 @@ Example - Fetch manifest with prettified json result:
...
+	if opts.cacheRoot != "" {
+		ociStore, err := oci.New(opts.cacheRoot)
+		if err != nil {
+			return err
+		}
+		src = cache.New(src, ociStore)
 	}
+
+	// fetch manifest
+	desc, content, err := oras.FetchBytes(ctx, src, opts.targetRef, oras.FetchBytesOptions{
```
shizhMSFT, line 116: What if only `descriptor` is needed?
For instance, `oras manifest fetch --descriptor localhost:5000/hello:latest`?

shizhMSFT (changes requested):

cmd/oras/manifest/fetch.go
yuehaoliang-microsoft: If only descriptor is needed, we don't need to fetch the content, which can save a lot of effort and improve the performance. I'll make a change in this way.

For `oras manifest fetch --descriptor localhost:5000/hello:latest`, only descriptor will be outputted to stdout, line 128 to 138.

cmd/oras/manifest/fetch.go
```diff
@@ -72,46 +82,93 @@ Example - Fetch manifest with prettified json result:
...
+	var src oras.ReadOnlyTarget = repo.Manifests()
+	if opts.cacheRoot != "" {
+		ociStore, err := oci.New(opts.cacheRoot)
+		if err != nil {
+			return err
 		}
-		buf.WriteByte('\n')
-		content = buf.Bytes()
+		src = cache.New(src, ociStore)
 	}
```
shizhMSFT, line 116: Should we refactor it into the `option` package?

/cc @qweeah

cmd/oras/manifest/fetch.go
```diff
@@ -72,46 +82,93 @@ Example - Fetch manifest with prettified json result:
...
+	} else {
+		// fetch manifest content
+		desc, content, err = oras.FetchBytes(ctx, src, opts.targetRef, oras.FetchBytesOptions{
+			FetchOptions: oras.FetchOptions{
+				ResolveOptions: oras.ResolveOptions{
+					TargetPlatform: targetPlatform,
+				},
+			},
+			MaxBytes: 0,
+		})
```
shizhMSFT, line 135: We can have a copy of `oras.DefaultFetchBytesOptions` and then set the `TargetPlatform`.
```suggestion
		// fetch manifest content
		fetchOpts := oras.DefaultFetchBytesOptions
		fetchOpts.TargetPlatform = targetPlatform
		desc, content, err = oras.FetchBytes(ctx, src, opts.targetRef, fetchOpts)
```

cmd/oras/manifest/fetch.go
```diff
@@ -72,46 +82,93 @@ Example - Fetch manifest with prettified json result:
...
+		}
+		defer func() {
+			if err := file.Close(); fetchErr == nil {
+				fetchErr = err
+			}
+		}()
+
+		if _, err = file.Write(content); err != nil {
+			return err
+		}
```
shizhMSFT, line 158: This can be simplified using [os.WriteFile()](https://pkg.go.dev/os#WriteFile).

shizhMSFT (changes requested):

cmd/oras/manifest/fetch.go
```diff
@@ -72,46 +82,93 @@ Example - Fetch manifest with prettified json result:
...
+	if opts.OutputDescriptor {
+		descBytes, err := json.Marshal(desc)
+		if err != nil {
+			return err
+		}
+		err = opts.Output(os.Stdout, descBytes)
+		if err != nil {
+			return err
+		}
+	}
```
qweeah, line 171: This code block appears everywhere if `--descriptor` and `--pretty` and both needed, any idea to refactor it? It doesn't necessarily need to be done in this PR /cc @lizMSFT

cmd/oras/internal/option/cache_test.go
```diff
@@ -0,0 +1,62 @@
...
+func TestCache_CachedTarget(t *testing.T) {
+	tempDir := t.TempDir()
+	os.Setenv("ORAS_CACHE", tempDir)
+	opts := Cache{}
+
+	ociStore, err := oci.New(tempDir)
+	if err != nil {
+		t.Fatal("error calling oci.New(), error =", err)
+	}
+	want := cache.New(mockTarget, ociStore)
```
yuehaoliang-microsoft, line 40: I was trying to write a better test here, but I didn't find a good way unless exporting the `cache.target` [struct](https://github.com/oras-project/oras/blob/main/internal/cache/target.go#L36).
```
want := struct {
	oras.ReadOnlyTarget
	cache content.Storage
}{
	mockTarget,
	ociStore,
}
```
If I built a struct like above, `want` and `got` will not be equal.

cmd/oras/manifest/fetch.go
yuehaoliang-microsoft: > 

We're discussing about the change offline.

cmd/oras/manifest/fetch.go
yuehaoliang-microsoft: > Should we refactor it into the `option` package?
> 
> /cc @qweeah

Created a option `Cache`.

qweeah (changes requested):

cmd/oras/manifest/fetch.go
```diff
@@ -72,46 +79,74 @@ Example - Fetch manifest with prettified json result:
...
 		return err
 	}
-	if opts.pretty {
-		buf := bytes.NewBuffer(nil)
-		if err = json.Indent(buf, content, "", "  "); err != nil {
-			return fmt.Errorf("failed to prettify: %w", err)
+
+	var desc ocispec.Descriptor
+	var content []byte
+	if opts.OutputDescriptor && opts.outputPath == "" {
```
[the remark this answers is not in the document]
yuehaoliang-microsoft: I couldn't really follow your idea here.

In the case of fetching manifest content, there still might be a need to output descriptor, eg.  `--output manifest.json --descriptor`. Since `oras.FetchBytes` returns both the content and the descriptor, so if content has already been fetched, we don't need to call `oras.Resolve` again.

Current logic:
```
if opts.OutputDescriptor && opts.outputPath == "" {
	// oras.Resolve   fetch manifest descriptor only
} else {
	// oras.FetchBytes   fetch manifest descriptor content

	if opts.outputPath == "" || opts.outputPath == "-" {
		// output manifest content

		return
	}

	// save manifest content into the local file
}

if opts.OutputDescriptor {
	// output manifest descriptor
}
```
In this way, for `--output manifest.json --descriptor`, there's a redundant `oras.Resolve`.
```
if !opts.OutputDescriptor || opt.outputPath != "" {
	// oras.FetchBytes   fetch manifest descriptor content

	if opts.outputPath == "" || opts.outputPath == "-" {
		// output manifest content

		return
	}

	// save manifest content into the local file
} 

if opts.OutputDescriptor {
	// oras.Resolve   fetch manifest descriptor only

	// output manifest descriptor
}
```

shizhMSFT (approved):
LGTM

cmd/oras/manifest/fetch.go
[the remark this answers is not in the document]
qweeah: > Since oras.FetchBytes returns both the content and the descriptor, so if content has already been fetched, we don't need to call oras.Resolve again

This can be avoided by checking if the desc is a empty struct.

PR #544, opened 2022-09-08
fix: handle push error for internal cache

### LINKED ISSUE #543: oras pull may stuck if cache enabled and pushing to cache fails
When using [cache.target](https://github.com/oras-project/oras/blob/9ea7aee84a1a3829a6506d44114424609b546f22/internal/cache/target.go#L36), we fetch data from the base storage while pushing it to the cache storage.

If any errors occur when pushing to the cache storage, there will be a deadlock. This is because the used pipe is not properly closed when there is a [push error](https://github.com/oras-project/oras/blob/9ea7aee84a1a3829a6506d44114424609b546f22/internal/cache/target.go#L80).

qweeah:
Resolves #543 

### DIFF
--- a/internal/cache/target.go
+++ b/internal/cache/target.go
@@ -78,6 +78,9 @@ func (t *target) cacheReadCloser(ctx context.Context, rc io.ReadCloser, target o
 	go func() {
 		defer wg.Done()
 		pushErr = t.cache.Push(ctx, target, pr)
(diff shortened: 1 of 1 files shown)


Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #542, opened 2022-09-07
ORAS CLI should support interacting with docker.io


FeynmanZhou:
Currently, ORAS is not able to authenticate `docker.io` and will return 401 error:
```
$ oras cp mcr.microsoft.com/mmlspark/spark2.4:1.0.0 docker.io/pengfeizhou/spark2.4:1.0.0
Error: POST "https://registry-1.docker.io/v2/pengfeizhou/spark2.4/blobs/uploads/": unexpected status code 401: unauthorized: authentication required
```
It requires users to specify `registry-1.docker.io` to authenticate with DockerHub. I think ORAS should support the most common domain `docker.io` when interacting with DockerHub.

shizhMSFT:
The root cause is the `docker.io` is a very special case. In docker CLI / daemon, the domain `docker.io` is mapped to `registry-1.docker.io`. `oras-go` does the same trick for `docker.io`.

However, when accessing the credentials, the hostname `registry-1.docker.io` is not recognized as `docker.io`.

@junczhu Could you help fixing this bug and update [internal/credential/store.go](https://github.com/oras-project/oras/blob/main/internal/credential/store.go) to match the same behavior as docker daemon and docker CLI?

junczhu:
Thanks for the insight.
@shizhMSFT Please assign this issue to me and I will start working on it.

junczhu:
@FeynmanZhou Have open an PR #549 for this fix. 
Please check it out

<img width="731" alt="image" src="https://user-images.githubusercontent.com/74894646/189294567-5c229d7a-cddb-4ac6-99b5-4193912fc8f8.png">

/cc @qweeah @shizhMSFT

junczhu:
For your information:

By default, command `docker login` and `docker login docker.io` will store following credential on line 3.
Command `docker login registry-1.docker.io` will store the credential on line 4.

<img width="380" alt="image" src="https://user-images.githubusercontent.com/74894646/189298062-97f6adca-327e-4c18-a614-fa9709247ced.png">

And when using docker cli to push an image to the docker hub, with default hostname, only the credential on line 3 can be used.

When explicitly using registry-1.docker.io as hostname, like `registry-1.docker.io/username/repositoryname`, only the second credential can be used.

In ORAS, we have mapped `'docker.io'` to `'registry-1.docker.io'` for registry reference.
And I am going to apply the same operation in the ORAS CLI to help fix this issue

junczhu:
[...]

PR #549, opened 2022-09-09
fix: provide function to support hostname mapping

junczhu:
Resolve: #542 
Signed-off-by: Juncheng Zhu <junczhu@microsoft.com>

### DIFF
--- a/cmd/oras/internal/option/remote.go
+++ b/cmd/oras/internal/option/remote.go
@@ -153,11 +153,11 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
 		// According to the the behavior of Docker CLI,
 		// credential under key "https://index.docker.io/v1/" should be provided
 		if registry == "docker.io" {
-			client.Credential = func(ctx context.Context, reg string) (auth.Credential, error) {
-				if reg == "registry-1.docker.io" {
-					reg = "https://index.docker.io/v1/"
+			client.Credential = func(ctx context.Context, hostname string) (auth.Credential, error) {
+				if hostname == "registry-1.docker.io" {
+					hostname = "https://index.docker.io/v1/"
 				}
-				return store.Credential(ctx, reg)
+				return store.Credential(ctx, hostname)
 			}
 		} else {
 			client.Credential = store.Credential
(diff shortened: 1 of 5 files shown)


qweeah:
What if username/password are passed through flags? Replacing host name in cred store only works when credential is empty,  see 
https://github.com/oras-project/oras/blob/main/cmd/oras/internal/option/remote.go#L152

internal/credential/store.go
```diff
@@ -113,3 +116,11 @@ func (s *Store) Credential(ctx context.Context, registry string) (auth.Credentia
 	}
 	return auth.EmptyCredential, nil
 }
+
+// convertHostname maps docker.io to registry-1.docker.io
+func convertHostname(registry string) string {
```
qweeah, line 121: use [registry.Host](https://github.com/oras-project/oras-go/blob/main/registry/reference.go#L177) directly

shizhMSFT:
@qweeah I've discussed offline with @junczhu. The behavior of `docker.io` in docker is special that it connects to the host of `registry-1.docker.io` using the credential with the reigstry name `https://index.docker.io/v1/`, which is really not a valid name (but it uses).

shizhMSFT (changes requested):
Please update this PR as discussed to use the credentials of `https://index.docker.io/v1/` for `docker.io` when connecting `registry-1.docker.io` as `docker.io`. Meanwhile, please well document the docker behavior in the corresponding comment block.

cmd/oras/internal/option/remote.go
```diff
@@ -110,7 +110,7 @@ func (opts *Remote) tlsConfig() (*tls.Config, error) {
 }
 
 // authClient assembles a oras auth client.
-func (opts *Remote) authClient(debug bool) (client *auth.Client, err error) {
+func (opts *Remote) authClient(host string, debug bool) (client *auth.Client, err error) {
```
qweeah, line 113: Why we need this new argument? It's not used anywhere.

qweeah (changes requested):
Can you add a unit test case for this host name replacement?

cmd/oras/login.go
```diff
@@ -116,7 +116,14 @@ func runLogin(opts loginOptions) (err error) {
 	if err != nil {
 		return err
 	}
-	if err := store.Store(opts.Hostname, opts.Credential()); err != nil {
+	// For a user case that login 'docker.io',
+	// According the the behavior of Docker CLI,
+	// credential should be added under key "https://index.docker.io/v1/"
+	hostname := opts.Hostname
+	if opts.Hostname == "docker.io" {
```
shizhMSFT, line 123:
```suggestion
	if hostname  == "docker.io" {
```

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #550, opened 2022-09-10
Missing `get` alias for `oras blob fetch`


shizhMSFT:
Since there is an alias `oras manifest get` for `oras manifest fetch`, it is better to have `oras blob get` alias to `oras blob fetch`.

shizhMSFT:
@lizMSFT Could you help on this issue?

lizMSFT:
I will take care of this issue. 👌

PR #558, opened 2022-09-13
fix: add get alias for oras blob fetch command

lizMSFT:
Add `oras blob get` alias for `oras blob fetch` command

Resolves: https://github.com/oras-project/oras/issues/550


shizhMSFT (approved):
LGTM

PR #574, opened 2022-09-16
feat: support tagging multiple tags at once to a remote registry

### LINKED ISSUE #355: Multiple tags at the once
I did not find either in the documentation of oras or somewhere else. Is there any multi-tagging support?
i.e:

`oras push myregistry/artifact:a,b,c` then in my registry following artifacs are available:
```
myregistry/artifact:a
myregistry/artifact:b
myregistry/artifact:c
```
I know I can make for each tag a `oras push` but the feature might be already there.

wju-MSFT:
Resolves #355 

Signed-off-by: Jun <wju@microsoft.com>


cmd/oras/tag/tag.go
```diff
@@ -26,14 +32,15 @@ type tagOptions struct {
 	option.Common
 	option.Remote
 
-	srcRef    string
-	targetRef string
+	defaultTagNOptions oras.TagNOptions
+	srcRef             string
+	targetRef          []string
```
Wwwsylvia, line 37: nit: targetRef**s**?

cmd/oras/tag/tag.go
```diff
@@ -42,19 +49,24 @@ func TagCmd() *cobra.Command {
 Example - Tag the manifest 'v1.0.1' in 'locahost:5000/hello' to 'v1.0.2':
   oras tag localhost:5000/hello:v1.0.1 v1.0.2
 
-Example - Tag the manifest with digest sha256:9463e0d192846bc994279417b50114606712d516aab45f4d8b31cbc6e46aad71 to 'v1.0.3'
-  oras tag localhost:5000/hello@sha256:9463e0d192846bc994279417b50114606712d516aab45f4d8b31cbc6e46aad71 v1.0.3
+Example - Tag the manifest with digest sha256:9463e0d192846bc994279417b50114606712d516aab45f4d8b31cbc6e46aad71 to 'v1.0.2'
+  oras tag localhost:5000/hello@sha256:9463e0d192846bc994279417b50114606712d516aab45f4d8b31cbc6e46aad71 v1.0.2
```
Wwwsylvia, line 53: Should we add an example for tagging multiple tags?

cmd/oras/tag/tag.go
```diff
@@ -26,14 +32,15 @@ type tagOptions struct {
 	option.Common
 	option.Remote
 
-	srcRef    string
-	targetRef string
+	defaultTagNOptions oras.TagNOptions
```
Wwwsylvia, line 35: Why is this called `default`? My understanding is that this is `tagNOptions`?

cmd/oras/tag/tag.go
```diff
@@ -70,5 +82,22 @@ func tagManifest(opts tagOptions) error {
...
+}
+
+type wrapper struct {
+	*remote.Repository
+}
+
+// Override PushReference method to print off which tag(s) were added successfully.
+func (w *wrapper) PushReference(ctx context.Context, expected ocispec.Descriptor, content io.Reader, reference string) error {
+	if err := w.Repository.PushReference(ctx, expected, content, reference); err != nil {
+		return fmt.Errorf("failed %s", reference)
```
Wwwsylvia, line 99: The `failed %s` error message looks a bit confusing. What does the error message look like if we return the `err` directly?

cmd/oras/tag/tag.go
```diff
@@ -70,5 +82,22 @@ func tagManifest(opts tagOptions) error {
...
+	}
+
+	return oras.TagN(ctx, rp, opts.srcRef, opts.targetRef, opts.defaultTagNOptions)
+}
+
+type wrapper struct {
+	*remote.Repository
+}
+
+// Override PushReference method to print off which tag(s) were added successfully.
```
Wwwsylvia, line 96: nit:  How about
> PushReference overrides Repository.PushReference ...

?

cmd/oras/tag/tag.go
```diff
@@ -70,5 +82,22 @@ func tagManifest(opts tagOptions) error {
...
 
-	return oras.Tag(ctx, repo, opts.srcRef, opts.targetRef)
+	rp := &wrapper{
+		repo,
+	}
+
+	return oras.TagN(ctx, rp, opts.srcRef, opts.targetRef, opts.defaultTagNOptions)
+}
+
+type wrapper struct {
```
Wwwsylvia, line 92: This name is kind of rough...

shizhMSFT (changes requested):
This PR is related to #355 but does not resolve it since #355 is to support multiple tags for `oras push`.

cmd/oras/tag/tag.go
shizhMSFT, line 35: Can we just call it for better readability?
```golang
concurrency int64
```

cmd/oras/tag/tag.go
```diff
@@ -42,19 +49,24 @@ func TagCmd() *cobra.Command {
...
 		},
 		RunE: func(_ *cobra.Command, args []string) error {
 			opts.srcRef = args[0]
-			opts.targetRef = args[1]
+			opts.targetRef = args[1:]
 			return tagManifest(opts)
 		},
 	}
+
+	cmd.Flags().Int64VarP(&opts.defaultTagNOptions.Concurrency, "concurrency", "", 0, "provide concurrency number, default is 5")
```
shizhMSFT, line 69: If the default is `5`, please make it `5`.
```suggestion
	cmd.Flags().Int64VarP(&opts.defaultTagNOptions.Concurrency, "concurrency", "", 5, "concurrency level")
```
so that it decouples with the `oras-go` library and the default value will be printed in `oras tag --help`.

cmd/oras/tag/tag.go
shizhMSFT, line 99: Suggesting a better UX.
```suggestion
		return fmt.Errorf("%s: %w", reference, err)
```

cmd/oras/tag/tag.go
```diff
@@ -70,5 +82,22 @@ func tagManifest(opts tagOptions) error {
...
+type wrapper struct {
+	*remote.Repository
+}
+
+// Override PushReference method to print off which tag(s) were added successfully.
+func (w *wrapper) PushReference(ctx context.Context, expected ocispec.Descriptor, content io.Reader, reference string) error {
+	if err := w.Repository.PushReference(ctx, expected, content, reference); err != nil {
+		return fmt.Errorf("failed %s", reference)
+	}
+	fmt.Println("tagged", reference)
```
shizhMSFT, line 101: Printing using `fmt.Println` is not go-routine safe. Please consider https://github.com/oras-project/oras/blob/main/cmd/oras/internal/display/print.go#L45

shizhMSFT (changes requested):

cmd/oras/tag/tag.go
wju-MSFT: done

cmd/oras/tag/tag.go
wju-MSFT: added an example for tagging multiple tags with default concurrency number.

cmd/oras/tag/tag.go
wju-MSFT: used `concurrence int64` now instead.

cmd/oras/tag/tag.go
wju-MSFT: the error itself looks like: `Error: failed to tag localhost:12900/hello:latest as tag55:1,: invalid reference: invalid tag`
changed to what Shiwei suggested `fmt.Errorf("%s: %w", reference, err)` and now error looks like: `Error: failed to tag localhost:12900/hello:latest as tag55:1,: tag55:1,: invalid reference: invalid tag`

cmd/oras/tag/tag.go
wju-MSFT: done

cmd/oras/tag/tag.go
wju-MSFT: done

cmd/oras/tag/tag.go
wju-MSFT: done

cmd/oras/tag/tag.go
wju-MSFT: in the above discussion, i think the error message itself already holds enough information for user to debug. How about we just `return err` 
stdout: `Error: failed to tag localhost:12900/hello:latest as tag55:1,: invalid reference: invalid tag`

cmd/oras/tag/tag.go
wju-MSFT: using `display.Print` now

wju-MSFT:
> This PR is related to #355 but does not resolve it since #355 is to support multiple tags for `oras push`. Please also support `oras manifest push` related to #355.

Shiwei, I will start a new PR to include changes for `oras push` and `oras manifest push`

cmd/oras/tag/tag.go
```diff
@@ -70,5 +86,24 @@ func tagManifest(opts tagOptions) error {
...
+	rp := &repository{
+		repo,
+	}
+
+	nOpts.Concurrency = opts.concurrency
+
+	return oras.TagN(ctx, rp, opts.srcRef, opts.targetRefs, nOpts)
+}
+
+type repository struct {
```
shizhMSFT, line 98: How about `tagManifestListener`?

shizhMSFT (changes requested):

cmd/oras/tag/tag.go
```diff
@@ -70,5 +86,24 @@ func tagManifest(opts tagOptions) error {
...
+	*remote.Repository
+}
+
+// PushReference overrides Repository.PushReference method to print off which tag(s) were added successfully.
+func (r *repository) PushReference(ctx context.Context, expected ocispec.Descriptor, content io.Reader, reference string) error {
+	if err := r.Repository.PushReference(ctx, expected, content, reference); err != nil {
+		return err
+	}
+	display.Print("tagged", reference)
```
shizhMSFT, line 107:
```suggestion
	display.Print("Tagged", reference)
```

shizhMSFT (approved):
LGTM

PR #589, opened 2022-09-21
fix: use VerifyReader to verify read content after oras.Fetch

### LINKED ISSUE #547: Need verify rc against the descriptor after calling oras.Fetch()
When reading from the io.ReadCloser returned by `oras.Fetch()` function, it's recommended to verify the read content against the descriptor. 
We could use the utility method from ORAS-GO to do the verification after https://github.com/oras-project/oras-go/issues/306 being resolved.

reference: 
`desc, rc, err := oras.Fetch(ctx, src, opts.targetRef, oras.FetchOptions{})`
https://github.com/oras-project/oras/pull/520#discussion_r965650328

wangxiaoxuan273:
Resolves #547 

### DIFF
--- a/cmd/oras/blob/fetch.go
+++ b/cmd/oras/blob/fetch.go
@@ -16,7 +16,6 @@ limitations under the License.
 package blob
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


cmd/oras/blob/fetch.go
```diff
@@ -114,6 +116,14 @@ func fetchBlob(opts fetchBlobOptions) (fetchErr error) {
 		// fetch blob content
 		var rc io.ReadCloser
 		desc, rc, err = oras.Fetch(ctx, src, opts.targetRef, oras.DefaultFetchOptions)
+		vr := content.NewVerifyReader(rc, desc)
+		buf := bytes.NewBuffer(nil)
+		if _, err := io.Copy(buf, vr); err != nil {
+			return err
+		}
```
shizhMSFT, line 123: Blobs are usually large and cannot be dumped into memory.

shizhMSFT (changes requested):

cmd/oras/blob/fetch.go
wangxiaoxuan273: Fixed according to our offline discussion.

shizhMSFT (approved):
LGTM

PR #596, opened 2022-09-22
fix: update usage doc for all oras commands

### LINKED ISSUE #593: unify styling for all command help documents
# Usage Notation
Reference: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key
- Text without brackets or braces: Items you must type as shown.
- `<Text inside angle brackets>`: Placeholder for which you must supply a value.
- `[Text inside square brackets]`: Optional items.
- `{Text inside braces}`: Set of required items. You must choose one.
- Vertical bar `( \| )`: Separator for mutually exclusive items. You must choose one.
- Ellipsis `(…)`: Items that can be repeated and used multiple times.

# Styling
- Flags should be put before arguments

### LINKED ISSUE #602: Improve `oras attach` help doc
Command:
```
oras attach -h
```
Output:
image

Users can't tell that `--artifact-type` is a required field from the current help doc.
Besides, the usage format `name<:tag|@digest>` is not consistent with other commands like `manifest fetch`.
image

Some improvement can be like:
1. Add `--artifact-type` into the usage doc:
```diff
- oras attach <name:tag|name@digest> [file[:type]...] [flags]
+ oras attach <name:tag|name@digest> [file[:type]...] --artifact-type <artifactType> [flags]
```
2. Call out that `--artifact-type` is a required filed in the flag description:
```diff
- artifact type
+ artifact type (required)
```

qweeah:
Resolves #593, resolves #602

### DIFF
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -44,7 +44,7 @@ type attachOptions struct {
 func attachCmd() *cobra.Command {
 	var opts attachOptions
 	cmd := &cobra.Command{
-		Use:   "attach [flags] <name>{:<tag>|@<digest>} {<file>[:<type>]}[...]",
+		Use:   "attach name<:tag|@digest> [file[:type]...]",
 		Short: "[Preview] Attach files to an existing artifact",
 		Long: `[Preview] Attach files to an existing artifact
 
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -51,13 +51,13 @@ func attachCmd() *cobra.Command {
 ** This command is in preview and under development. **
 
 Example - Attach file 'hi.txt' with type 'doc/example' to manifest 'hello:test' in registry 'localhost:5000'
-  oras attach localhost:5000/hello:test hi.txt --artifact-type doc/example
+  oras attach --artifact-type doc/example localhost:5000/hello:test hi.txt
 
 Example - Attach and update manifest annotations
-  oras attach localhost:5000/hello:latest hi.txt --artifact-type doc/example --annotation "key=val"
+  oras attach --artifact-type doc/example --annotation "key=val" localhost:5000/hello:latest hi.txt
 
 Example - Attach and update annotation from manifest annotation file
-  oras attach localhost:5000/hello:latest hi.txt --artifact-type doc/example --annotation-file annotation.json
+  oras attach --artifact-type doc/example --annotation-file annotation.json localhost:5000/hello:latest hi.txt
 `,
 		Args: cobra.MinimumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -56,8 +56,8 @@ Example - Attach file 'hi.txt' with type 'doc/example' to manifest 'hello:test' 
 Example - Attach file 'hi.txt' and add annotations from file 'annotation.json'
   oras attach --artifact-type doc/example --annotation-file annotation.json localhost:5000/hello:latest hi.txt
 
-Example - Attach an artifact with manifest annotation
-  oras attach --artifact-type doc/example --annotation "key=val" localhost:5000/hello:latest
+Example - Attach an artifact with manifest annotations
+  oras attach --artifact-type doc/example --annotation "key1=val1" --annotation "key2=val2" localhost:5000/hello:latest
 
 Example - Attach a file and add manifest annotations
   oras attach --artifact-type doc/example --annotation "key=val" localhost:5000/hello:latest hi.txt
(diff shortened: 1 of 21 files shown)


cmd/oras/attach.go
```diff
@@ -44,7 +44,7 @@ type attachOptions struct {
 func attachCmd() *cobra.Command {
 	var opts attachOptions
 	cmd := &cobra.Command{
-		Use:   "attach name<:tag|@digest> [file[:type]...]",
+		Use:   "attach [flags] <name>{:<tag>|@<digest>} [<file>[:<type>]...]",
```
shizhMSFT, line 47:
```suggestion
		Use:   "attach [flags] <name>{:tag|@digest} [file[:type]] ...",
```

qweeah: According to the syntax [guideline](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key), we should not, e.g. https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cd

cmd/oras/attach.go
```diff
@@ -44,7 +44,7 @@ type attachOptions struct {
 func attachCmd() *cobra.Command {
 	var opts attachOptions
 	cmd := &cobra.Command{
-		Use:   "attach name<:tag|@digest> [file[:type]...]",
+		Use:   "attach [flags] <name>{:<tag>|@<digest>} {<file>[:<type>]}[...]",
```
shizhMSFT, line 47: Can we close #603 and merge the change here? Also, update the description to resolve #602?
```suggestion
		Use:   "attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} [<file>[:<type>]] [...]",
```

cmd/oras/blob/fetch.go
```diff
@@ -42,7 +42,7 @@ type fetchBlobOptions struct {
 func fetchCmd() *cobra.Command {
 	var opts fetchBlobOptions
 	cmd := &cobra.Command{
-		Use:   "fetch [flags] <name@digest>",
+		Use:   "fetch [flags] {--output <file>|--descriptor} <name>@<digest>",
```
shizhMSFT, line 45: Should we add spaces around `|`? I'm just wondering.
```suggestion
		Use:   "fetch [flags] {--output <file> | --descriptor} <name>@<digest>",
```

cmd/oras/blob/push.go
```diff
@@ -42,7 +42,7 @@ type pushBlobOptions struct {
 func pushCmd() *cobra.Command {
 	var opts pushBlobOptions
 	cmd := &cobra.Command{
-		Use:   "push [flags] name[@digest] file",
+		Use:   "push [flags] <name>[@digest] <file>",
```
shizhMSFT, line 45:
```suggestion
		Use:   "push [flags] <name>[@<digest>] <file>",
```

shizhMSFT, line 45: I just found a UX inconsistency here.

We should remove the `--media-type` option. Instead, we should
```suggestion
		Use:   "push [flags] <name>[@digest] <file>[:<type>]",
```

/cc @qweeah @lizMSFT @yuehaoliang-microsoft @FeynmanZhou

cmd/oras/manifest/push.go
```diff
@@ -43,7 +43,7 @@ type pushOptions struct {
 func pushCmd() *cobra.Command {
 	var opts pushOptions
 	cmd := &cobra.Command{
-		Use:   "push [flags] name[:tag|@digest] file",
+		Use:   "push [flags] <name>[:<tag>|@<digest>] <file>",
```
shizhMSFT, line 46: I just found a UX inconsistency here.

We should remove the `--media-type` option. Instead, we should
```suggestion
		Use:   "push [flags] <name>[:<tag>|@<digest>] <file>[:<type>]",
```
/cc @qweeah @lizMSFT @yuehaoliang-microsoft @FeynmanZhou

cmd/oras/push.go
```diff
@@ -46,7 +46,7 @@ type pushOptions struct {
 func pushCmd() *cobra.Command {
 	var opts pushOptions
 	cmd := &cobra.Command{
-		Use:   "push name[:tag|@digest] file[:type] [file...]",
+		Use:   "push [flags] <name>[:<tag>|@<digest>] <file>[:type] [...]",
```
shizhMSFT, line 49:
```suggestion
		Use:   "push [flags] <name>[:<tag>|@<digest>] <file>[:<type>] [...]",
```

shizhMSFT (changes requested):

cmd/oras/blob/fetch.go
shizhMSFT: Feel free to defend it.

qweeah: We should. The reason is simple: it looks better.

shizhMSFT (approved):
LGTM

PR #605, opened 2022-09-22
fix: update help doc for blob fetch command

### LINKED ISSUE #597: Update `oras blob fetch` documentation to better explain flag usage
For `oras blob fetch`, the current Usage states:  `oras blob fetch [flags] <name@digest>`, which indicates the flags are optional, but when we try
`oras blob fetch <name@digest>`, we will get `Error: either --output or --descriptor must be provided`.
We will need to update the documentation to better explain flag usage.

qweeah:
Resolves #597 


shizhMSFT (approved):
LGTM

PR #607, opened 2022-09-22
fix: improve the behavior of user confirmation

### LINKED ISSUE #592: ORAS delete confirmation prompt directly pressing Enter
When trying to delete using ORAS CLI, after the confirmation prompt pops up, directly pressing Enter (without typing anything) would trigger "unexpected newline" error. Expect behavior is to default as 'No'. 

related code: https://github.com/oras-project/oras/blob/02274658891c7968e8b641138a23eeb52dbd13de/cmd/oras/internal/option/confirmation.go#L46

yuehaoliang-microsoft:
Resolve #592 


shizhMSFT (approved):
LGTM

PR #617, opened 2022-09-23
feat: support tagging multiple tags at once for oras copy

wju-MSFT:
Resolves: #355 
Signed-off-by: Jun <wju@microsoft.com>

PR #621, opened 2022-09-26
feat: add concurrency flag for oras pull

wju-MSFT:
resolves: #620 

Signed-off-by: Jun <wju@microsoft.com>


cmd/oras/push.go
```diff
@@ -86,7 +86,7 @@ Example - Push file "hi.txt" with multiple tags:
 
 Example - Push file "hi.txt" with multiple tags and concurrency level tuned:
   oras push --concurrency 6 localhost:5000/hello:tag1,tag2,tag3 hi.txt
-  `,
```
shizhMSFT, line 89: Can we put this change in #617?

wju-MSFT: sg.

shizhMSFT (approved):
LGTM

Issue #630, opened 2022-10-03
Manifest level annotations not being pushed when there is no blob in attach command


sajayantony:
The manifest should add the annotations passed in when there are no blobs specified 

```
➜ oras version
Version:        0.15.0+unreleased
Go version:     go1.19
Git commit:     9e12d92c3f25ffea67b9bce073c5c46b9d3c355c
Git tree state: clean
```

```
➜ oras attach --annotation "test=value" \
   sajay.azurecr.io/hello-world-attach@sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4 \
   --artifact-type example/annotation \
    --annotation test2=value2
Uploading empty artifact
Attached to sajay.azurecr.io/hello-world-attach@sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4
Digest: sha256:93c4a77febf73c80a70839bed23911dd83f79bda72e7bc96bb1e64c88981fd27
```

```
➜ oras manifest fetch --pretty sajay.azurecr.io/hello-world-attach@sha256:93c4a77febf73c80a70839bed23911dd83f79bda72e7bc96bb1e64c88981fd27
{
  "mediaType": "application/vnd.cncf.oras.artifact.manifest.v1+json",
  "artifactType": "example/annotation",
  "blobs": [],
  "subject": {
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "digest": "sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4",
    "size": 525
  },
  "annotations": {
    "io.cncf.oras.artifact.created": "2022-10-03T21:47:26Z"
  }
}
```

sajayantony:
Looks like we need to pass in the manifest Annotations - @junczhu 

https://github.com/sajayantony/oras/blob/e7a8a33b41454bb5c3c2fd85e7f29b18fc7b3d5e/cmd/oras/attach.go#L121-L126

https://github.com/oras-project/oras/blob/9e12d92c3f25ffea67b9bce073c5c46b9d3c355c/cmd/oras/attach.go#L122-L126

This came about when testing showing the annotations as a part of the OCI specification - 

<img width="618" alt="image" src="https://user-images.githubusercontent.com/1821104/193705158-6ea9ee36-e902-4d11-be56-595e6b5130df.png">

[...]

PR #632, opened 2022-10-08
fix: annotates manifest for attach command

qweeah:
Resolves: #630


shizhMSFT (approved):
LGTM

Issue #554, opened 2022-09-12
test(e2e): add execution helper


qweeah:
add execution helper so E2E specs can easily run with customized 
- stdin/stdout/stderr rediection
- matching output in stdout/stderr

Realted to #523

Issue #556, opened 2022-09-12
test(e2e): add documentation for E2E testing


qweeah:
Need a document to cover
- How to setup and run E2E tests on a local dev machine
- Best practices need to be follow when writing E2E specs

Issue #555, opened 2022-09-12
test(e2e): matching output of spec runs


qweeah:
Need to support below when running a e2e spec

- Match full content of a output
- Match keywords
- Match status logs in the output
- Match JSON output
- Use multiple matchers in one command execution

Related to #523

PR #639, opened 2022-10-11
test(e2e): add E2E testing utilities and sample specs

qweeah:
Resolves: #554, #555, #556


test/e2e/internal/utils/file.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+	defer from.Close()
+
+	_, err = io.Copy(to, from)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
shizhMSFT, line 66:
```suggestion
	return err
```

test/e2e/internal/utils/registry.go
```diff
@@ -0,0 +1,28 @@
...
+// Reference generates the reference string from given parameters.
+func Reference(registry string, repo string, tagOrDigest string) string {
+	ret := registry + "/" + repo
+
+	if err := digest.Digest(tagOrDigest).Validate(); err != nil {
+		ret += ":" + tagOrDigest
+	} else {
+		ret += "@" + tagOrDigest
+	}
+	return ret
```
shizhMSFT, line 27: We probably need to document the reason why we don't use `registry.Reference` from` oras-go` to generate the reference string.

test/e2e/internal/utils/match/content.go
```diff
@@ -0,0 +1,33 @@
...
+
+package match
+
+import (
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+)
+
+// contentMatcher provides whole matching of the output.
+type contentMatcher struct{ s string }
```
shizhMSFT, line 22: How about
```suggestion
type contentMatcher string
```

test/e2e/internal/utils/match/keywords.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/onsi/gomega/gbytes"
+)
+
+// keywordMatcher provides selective matching of the output.
+// The match will pass if all key words existed case-insensitively in the
+// output.
+type keywordMatcher []string
+
+func NewKeywordMatcher(kw []string) keywordMatcher {
```
shizhMSFT, line 28: Curious if `NewKeywordMatcher(kw ...string)` will be better or not. Just wondering as I don't know how this method will be used.

test/e2e/internal/utils/match/keywords.go
```diff
@@ -0,0 +1,47 @@
...
+	raw := string(got.Contents())
+	lowered := strings.ToLower(raw)
+	for _, w := range want {
+		if !strings.Contains(lowered, strings.ToLower(w)) {
+			missed = append(missed, w)
+		}
+	}
+
+	if len(missed) != 0 {
+		fmt.Printf("Keywords missed: %v ", missed)
```
shizhMSFT, line 44:
```suggestion
		fmt.Printf("Keywords missed: %v", missed)
```

shizhMSFT: BTW, is it `Printf` or `Println`?

test/e2e/internal/utils/match/matchable.go
```diff
@@ -0,0 +1,23 @@
...
+*/
+
+package match
+
+import (
+	"github.com/onsi/gomega/gbytes"
+)
+
+// Matchable provides matching method towards an output buffer.
+type Matchable interface {
```
shizhMSFT, line 21: In golang convention, it is called `Matcher` for `Match()`.

test/e2e/internal/utils/match/status.go
```diff
@@ -0,0 +1,171 @@
...
+type StateKey struct {
+	Digest string
+	Name   string
+}
+
+type state struct {
+	uint // padding to avoid zero-based pointer
+}
+
+type edge = struct {
```
shizhMSFT, line 37:
```suggestion
type edge struct {
```

test/e2e/internal/utils/match/status.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// stateMachine with edges named after known status.
+type stateMachine struct {
+	edges map[status][]edge
+	start *state
+	end   *state
+}
+
+func newStateMachine(cmd string) (sm *stateMachine) {
```
shizhMSFT, line 49: It seems the named parameter is not necessary.
```suggestion
func newStateMachine(cmd string) *stateMachine {
```

test/e2e/internal/utils/match/status.go
```diff
@@ -0,0 +1,171 @@
...
+
+// StateKey represents the expected value of second and third fields in status log.
+type StateKey struct {
+	Digest string
+	Name   string
+}
+
+type state struct {
+	uint // padding to avoid zero-based pointer
+}
```
shizhMSFT, line 35: How about
```go
type state uint64

var lastState uint64

func newState() uint64 {
      return atomic.AddUint64(&lastState, 1)
}

```

.github/workflows/build.yml
```diff
@@ -44,18 +44,21 @@ jobs:
...
           go install github.com/onsi/ginkgo/v2/ginkgo
+          mnt_root="$GITHUB_WORKSPACE/test/e2e/testdata/distribution/mount"
           rm -rf $mnt_root/docker
           for layer in $(ls $mnt_root/*.tar.gz); do
             tar -xvzf $layer -C $mnt_root
           done
           trap 'docker kill oras-e2e || true' ERR
-          docker run -d -p 5000:5000 --rm --name oras-e2e \
-            --env STORAGE_DELETE_ENABLED=true \
+          docker run --pull always -d -p 5000:5000 --rm --name oras-e2e \
```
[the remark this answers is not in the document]
qweeah: The default behavior is `missing`, which can be wrong if github provided an uncleaned VM. `always` is more defensive.

test/e2e/internal/utils/exec.go
```diff
@@ -0,0 +1,171 @@
...
+	stdin      io.Reader
+	stdout     []match.Matchable
+	stderr     []match.Matchable
+	shouldFail bool
+
+	text string
+}
+
+// ORAS returns default execution option for oras binary.
+func ORAS(args ...string) *execOption {
```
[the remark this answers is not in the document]
qweeah: You are right but I don't expect user to use the struct type directly.

test/e2e/internal/utils/exec.go
```diff
@@ -0,0 +1,171 @@
...
+	if opts == nil {
+		// this should be a code error but can only be caught during runtime
+		panic("Nil option for command execution")
+	}
+
+	if opts.text == "" {
+		if opts.shouldFail {
+			opts.text = "fail"
+		} else {
+			opts.text = "succeed"
```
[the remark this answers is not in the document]
qweeah: Is `should ok` sounds `ok`? User will see steps logs like: `should ok: oras manifest localhost:5000/command/images:multi`

test/e2e/internal/utils/registry.go
qweeah: Actually we should, why not.

test/e2e/internal/utils/match/keywords.go
qweeah: It's not expected to used direct by spec writer. I added 2 fluent-styled methods, `MatchKeyWords` and `MatchErrKeyWords`, both providing `...string` experience

test/e2e/internal/utils/match/keywords.go
qweeah: added newline

test/e2e/internal/utils/match/status.go
```diff
@@ -0,0 +1,176 @@
...
+// status represents the expected value of first field in the status log.
+type status = string
+
+// StateKey represents the expected value of second and third fields in status log.
+type StateKey struct {
+	Digest string
+	Name   string
+}
+
+type state uint32
```
shizhMSFT, line 34: `uint64` seems better. If it is hard to decide whether to use `uint32` or `uint64`, `uintptr` is also good choice.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #641, opened 2022-10-11
ci: run all check CIs against patch PR

### LINKED ISSUE #640: build action should also be ran against PRs targeting release branch
Otherwise we need to manually run build task before merging any patch PR

qweeah:
I have validated it in my own repo: [build](https://github.com/qweeah/oras/actions/runs/3225144165)

Resolves #640 

### DIFF
--- a/.github/workflows/license-checker.yml
+++ b/.github/workflows/license-checker.yml
@@ -17,7 +17,9 @@ on:
   push:
     branches: main
   pull_request:
-    branches: main
+    branches:
+      - main
+      - release-*
 
 permissions:
   contents: write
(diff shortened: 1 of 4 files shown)


shizhMSFT (changes requested):
How about others? like CodeQL and license checker?

qweeah:
> How about others? like CodeQL and license checker?

Good point, updated.

.github/workflows/license-checker.yml
```diff
@@ -17,7 +17,9 @@ on:
   push:
     branches: main
```
shizhMSFT, line 18: How about `push`?

qweeah: You are right! Added.

shizhMSFT (approved):
LGTM

Issue #654, opened 2022-10-19
support OCI image fallback in `oras attach` and `oras push`


qweeah:
As is described in (distribution-spec)[https://github.com/opencontainers/distribution-spec/blob/main/spec.md#push

When pushing OCI artifact manifests to a registry, if the response is 400 Bad Request, fallback to pushing OCI image manifest.

PR #665, opened 2022-10-26
feat!: support OCI image fallback for `oras push` and `oras attach`

qweeah:
Since oras-go will take care of fallback with tag schema if referrer API not supported, this PR provides OCI Image fallback when-and-only-when OCI artifact is not supported.

Resolves: #654

### DIFF
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -19,14 +19,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/content/file"
-	"oras.land/oras/cmd/oras/internal/display"
 	oerrors "oras.land/oras/cmd/oras/internal/errors"
 	"oras.land/oras/cmd/oras/internal/option"
 )
@@ -124,7 +122,6 @@ func runAttach(opts attachOptions) error {
 	}
 
 	// prepare push
-	committed := &sync.Map{}
 	graphCopyOptions := oras.DefaultCopyGraphOptions
 	graphCopyOptions.Concurrency = opts.concurrency
 	graphCopyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
@@ -134,18 +131,7 @@ func runAttach(opts attachOptions) error {
 		}
 		return content.Successors(ctx, fetcher, node)
 	}
-	graphCopyOptions.PreCopy = display.StatusPrinter("Uploading", opts.Verbose)
-	graphCopyOptions.OnCopySkipped = func(ctx context.Context, desc ocispec.Descriptor) error {
-		committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
-		return display.PrintStatus(desc, "Exists   ", opts.Verbose)
-	}
-	graphCopyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
-		committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
-		if err := display.PrintSuccessorStatus(ctx, desc, "Skipped  ", store, committed, opts.Verbose); err != nil {
-			return err
-		}
-		return display.PrintStatus(desc, "Uploaded ", opts.Verbose)
-	}
+	updateDisplayOption(&graphCopyOptions, store, opts.Verbose)
 
 	// push
 	err = oras.CopyGraph(ctx, store, dst, root, graphCopyOptions)
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -111,30 +111,33 @@ func runAttach(opts attachOptions) error {
 	if err != nil {
 		return err
 	}
-	root, err := oras.Pack(
-		ctx, store, opts.artifactType, descs,
-		oras.PackOptions{
-			Subject:             &subject,
-			ManifestAnnotations: annotations[option.AnnotationManifest],
-		})
-	if err != nil {
-		return err
-	}
 
 	// prepare push
+	packOpts := oras.PackOptions{
+		Subject:             &subject,
+		ManifestAnnotations: annotations[option.AnnotationManifest],
+	}
+	pack := func() (ocispec.Descriptor, error) {
+		return oras.Pack(ctx, store, opts.artifactType, descs, packOpts)
+	}
+
 	graphCopyOptions := oras.DefaultCopyGraphOptions
 	graphCopyOptions.Concurrency = opts.concurrency
-	graphCopyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
-		if isEqualOCIDescriptor(node, root) {
-			// skip subject
-			return descs, nil
-		}
-		return content.Successors(ctx, fetcher, node)
-	}
 	updateDisplayOption(&graphCopyOptions, store, opts.Verbose)
+	copy := func(root ocispec.Descriptor) error {
+		if root.MediaType == ocispec.MediaTypeArtifactManifest {
+			graphCopyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+				if content.Equal(node, root) {
+					// skip subject
+					return descs, nil
+				}
+				return content.Successors(ctx, fetcher, node)
+			}
+		}
+		return oras.CopyGraph(ctx, store, dst, root, graphCopyOptions)
+	}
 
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module oras.land/oras
 
+replace oras.land/oras-go/v2 => github.com/Wwwsylvia/oras-go/v2 v2.0.0-20221031082137-4be7b30a924e
```
Wwwsylvia, line 3: This is not needed anymore.

internal/oci/artifact.go
```diff
@@ -0,0 +1,57 @@
...
+		root, err = pack(opts)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		err = copy(root)
+	}
+	return root, err
+}
+
+func ociArtifactUnsupported(err error) bool {
```
Wwwsylvia, line 52: nit: how about `isArtifactUnsupportedErr` and something like that?

internal/oci/artifact.go
```diff
@@ -0,0 +1,57 @@
...
+	var errResp *errcode.ErrorResponse
+	var errCode errcode.Error
+	return errors.As(err, &errResp) && errResp.StatusCode == http.StatusBadRequest &&
```
Wwwsylvia, line 55: Do we need to check the status code? I thought the error code `MANIFEST_INVALID` is used only for 400 responses.

cmd/oras/attach.go
```diff
@@ -109,46 +110,46 @@ func runAttach(opts attachOptions) error {
...
+	var committed sync.Map
+	copyFunc := oci.CopyFunc(func(root ocispec.Descriptor) error {
+		o := display.UploadOption(store, &committed, opts.concurrency, opts.Verbose, blobs)
+		findSuccessors := o.FindSuccessors
+		o.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+			successors, err := findSuccessors(ctx, fetcher, desc)
+			if err != nil {
+				return nil, err
+			}
+			if !isEqualOCIDescriptor(desc, root) {
```
Wwwsylvia, line 133: You may want to leverage `content.Equal()`.

internal/oci/artifact.go
```diff
@@ -0,0 +1,57 @@
...
+
+		// fallback to OCI image
+		opts.PackImageManifest = true
+		root, err = pack(opts)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		err = copy(root)
+	}
+	return root, err
```
Wwwsylvia, line 49: nit: The readability of this code block might not be very good.

internal/oci/artifact.go
qweeah: Yes we do. It's used by [4XX](https://github.com/opencontainers/distribution-spec/blob/v1.0.0-rc3/spec.md#error-codes).

internal/oci/artifact.go
qweeah: Changed accordingly.

internal/oci/artifact.go
qweeah: how about now?

internal/oci/artifact.go
Wwwsylvia: Can `MANIFEST_INVALID` error code implies 400?

qweeah: Intuitively yes, but is it documented anywhere?

Considering doing one extra check only increases e2e latency, it's worth doing.

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
-	oras.land/oras-go/v2 v2.0.0-rc.3.0.20221018111647-1969551cc3c7
+	oras.land/oras-go/v2 v2.0.0-20221031092619-f701e60ea36c
```
Wwwsylvia, line 14: Why not `v2.0.0-rc.4`?

internal/oci/artifact.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	// fallback to OCI image
+	dst.SetReferrersCapability(false)
+	opts.PackImageManifest = true
+	root, err = pack(opts)
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	return root, copy(root)
```
Wwwsylvia, line 50: Do we want to return `root` when there is error?

internal/oci/artifact.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func isOciArtifactUnsupportedErr(err error) bool {
+	var errResp *errcode.ErrorResponse
+	var errCode errcode.Error
+	return errors.As(err, &errResp) && errResp.StatusCode == http.StatusBadRequest &&
+		errors.As(errResp, &errCode) && (errCode.Code == errcode.ErrorCodeManifestInvalid ||
+		// below is special-casing for ECR
+		errCode.Code == errcode.ErrorCodeUnsupported)
```
Wwwsylvia, line 59: nit: How about
```suggestion
	var errResp *errcode.ErrorResponse
	if !errors.As(err, &errResp) || errResp.StatusCode != http.StatusBadRequest {
		return false
	}
	var errCode errcode.Error
	return errors.As(errResp, &errCode) &&
		(errCode.Code == errcode.ErrorCodeManifestInvalid || errCode.Code == errcode.ErrorCodeUnsupported)
```
for better readability?
As for the ECR comment, maybe we can make it like [this](https://github.com/oras-project/oras-go/blob/main/pack.go#L136-L137)?

cmd/oras/attach.go
```diff
@@ -109,46 +110,46 @@ func runAttach(opts attachOptions) error {
...
+			// skip subject to save one HEAD towards dst
+			j := len(successors) - 1
+			for i, s := range successors {
+				if content.Equal(s, subject) {
+					// swap subject to end and slice it off
+					successors[i] = successors[j]
+					return successors[:j], nil
+				}
+			}
+			return nil, fmt.Errorf("failed to find subject %v in the packed root %v", subject, root)
```
Wwwsylvia, line 146: Can you help me to understand this?

internal/oci/artifact.go
qweeah: Done with restricted check that ErrorCodeUnsupported happens on root.

internal/oci/artifact.go
qweeah: refactored

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

cmd/oras/push.go
```diff
@@ -120,94 +120,178 @@ func runPush(opts pushOptions) error {
...
+
+	var errCode errcode.Error
+	if !errors.As(errResp, &errCode) {
+		return false
+	}
+
+	// As of November 2022, ECR is known to return UNSUPPORTED error when
+	// putting an OCI artifact manifest.
+	switch errCode.Code {
+	case errcode.ErrorCodeManifestInvalid, errcode.ErrorCodeUnsupported:
```
nima, line 293: Probably a non-issue, just curious how do we deal with the casing (upper-case/lower-case) here? (e.g., `UNSUPPORTED` vs `Unsupported`)

```
Error: PUT "https://080565210187.dkr.ecr.us-east-1.amazonaws.com/v2/utnubu/manifests/oras.unknown":
  unexpected status code 400: unsupported:
  Unsupported manifest type: "application/vnd.oci.artifact.manifest.v1+json"
```

vs

```
% git grep ErrorCodeUnsupported
registry/remote/errcode/errors.go:42:   ErrorCodeUnsupported         = "UNSUPPORTED"
```

cmd/oras/push.go
```diff
@@ -120,94 +120,178 @@ func runPush(opts pushOptions) error {
...
+		return nil, nil
+	}
+	if err = copy(root); err != nil {
 		return ocispec.Descriptor{}, err
 	}
-	return manifestDesc, nil
+	return root, nil
+}
+
+func isArtifactUnsupported(err error) bool {
```
nima, line 279: Curious why we chose names that result in double-negatives?  For example *"is it not not-supported"* vs *"is it supported"*.

cmd/oras/push.go
sajayantony: I think it aligns with the casing here - https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes

qweeah: @nima Can you help confirm if the fallback works for ECR? I notice that the in-body error message was not capitalized in ECR, shown as below:

> unexpected status code 400: unsupported: Unsupported manifest type: "application/vnd.oci.artifact.manifest.v1+json"


You can `make` from source and run a `oras attach` based on ECR

cmd/oras/push.go
qweeah: @nima Nice catch! @shizhMSFT How about renaming it to `NoFallbackToOciImage` or something indicates that we should not retry with OCI Image?

The reason is both `isArtifactUnsupported` and `isArtifactSupported` is not accurate, **we don't know if artifact is supported if the other HTTP error happens**.

qweeah: Moving the disscusion to #672

Issue #667, opened 2022-11-01
Use shorter commands by default


shizhMSFT:
Human beings are lazy to type. It is better to default to shorter command names
- `repository` --> `repo`
- `list` --> `ls`
- `show-tags` --> `tag`

Basically, the alias and command names should be swapped.

FeynmanZhou:
> Human beings are lazy to type. It is better to default to shorter command names
> 
> * `repository` --> `repo`
> * `list` --> `ls`
> * `show-tags` --> `tag`
> 
> Basically, the alias and command names should be swapped.

I think `show-tags` should be `tags`.

shizhMSFT:
> I think `show-tags` should be `tags`.

Yes, it's `tags`. My typo.

shizhMSFT:
Since `list` becomes unix style `ls`, it is better to rename `delete` to `rm`
- `oras manifest delete` --> `oras manifest rm`
- `oras blob delete` --> `oras manifest rm`

Since we rename `delete` to `rm`, it is better to rename `--yes` to `--force` to be consistent with the unix command `rm`.

Additionally, since `oras manifest fetch` has an alias `get`, we should also have an alias `get-config` for `fetch-config`.

shizhMSFT:
According to [Remove -vs- Delete](https://apps.labor.ny.gov/ux/doc/v1/design-remove-vs-delete.html),
> Remove and Delete are defined quite similarly, but the main difference between them is that delete means erase (i.e. rendered nonexistent or nonrecoverable), while remove denotes take away and set aside (but kept in existence).

manifest / blob delete are indeed delete. Hence, we should keep the `delete` for the correct wording and uses rm just for alias.

PR #668, opened 2022-11-02
feat: support excluding digest tags

### LINKED ISSUE #658: support exclude digest tags in result of `oras repo tags`
If a registry doesn't support referrer API natively, referrers will be tracked via client- maintained referrer tags following [schema](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#referrers-tag-schema). 

User should be able to hide those tags in the result of `oras repo tags` via flag `--exclude-digest-tags`.

qweeah:
Making this PR draft since I found that it's important to write good example and long description for the new flag.
Resolves #658

### DIFF
--- a/cmd/oras/repository/show-tags.go
+++ b/cmd/oras/repository/show-tags.go
@@ -45,10 +45,7 @@ Example - Show tags of the target repository:
   oras repository show-tags localhost:5000/hello
 
 Example - Show tags in the target repository with digest-like tags hidden:
-oras repository show-tags localhost:5000/hello --exclude-digest-tag
-
-Example - Show tags of the target repository that include values lexically after last:
-oras repository show-tags --last "last_tag" localhost:5000/hello
+  oras repository show-tags --exclude-digest-tag localhost:5000/hello
 
 Example - Show tags of the target repository that include values lexically after last:
   oras repository show-tags --last "last_tag" localhost:5000/hello
@@ -61,7 +58,7 @@ Example - Show tags of the target repository that include values lexically after
 		},
 	}
 	cmd.Flags().StringVar(&opts.last, "last", "", "start after the tag specified by `last`")
-	cmd.Flags().BoolVar(&opts.excludeDigestTag, "exclude-digest-tags", false, "exclude all digest-like tags")
+	cmd.Flags().BoolVar(&opts.excludeDigestTag, "exclude-digest-tags", false, "exclude all digest-like tags such as 'sha256-aaaa...'")
 	option.ApplyFlags(&opts, cmd.Flags())
 	return cmd
 }
--- a/cmd/oras/repository/show-tags.go
+++ b/cmd/oras/repository/show-tags.go
@@ -55,7 +55,7 @@ Example - Show tags of the target repository that include values lexically after
 		},
 	}
 	cmd.Flags().StringVar(&opts.last, "last", "", "start after the tag specified by `last`")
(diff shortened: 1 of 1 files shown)


cmd/oras/repository/show-tags.go
```diff
@@ -52,6 +55,7 @@ Example - Show tags of the target repository that include values lexically after
 		},
 	}
 	cmd.Flags().StringVar(&opts.last, "last", "", "start after the tag specified by `last`")
+	cmd.Flags().BoolVar(&opts.excludeDigestTag, "exclude-digest-tag", false, "exclude all digest tags created by OCI artifact tag schema")
```
Wwwsylvia, line 58: nit: maybe "referrers tag schema"?🤔

Wwwsylvia (approved):
LGTM

cmd/oras/repository/show-tags.go
qweeah: I might need to rephrase this and add example latter, maybe without mentioning `referrer` or `tag schema`, below is why:

The use case of this flag is: one user found that `oras repo tags` is listing out some unexpected tags, he/she want to hide them. The struggling point is: **how does this user know about the unwanted tags**?
- If he/she knows well of the distribution spec, then we only need to make the flag name explainative on the behavior of filtering or hidden
- If he/she knows nothing of the distribution spec(most users don't), then mentioning `referrers` or `tag schema` in the long description won't help them realize that this flag is what they need to use.

Wwwsylvia: I agree. The users don't need to know anything about the referrers implementation.

Wwwsylvia: Brainstorming: How about "metadata tags"?

cmd/oras/repository/show-tags.go
```diff
@@ -41,6 +44,12 @@ func showTagsCmd() *cobra.Command {
 Example - Show tags of the target repository:
   oras repository show-tags localhost:5000/hello
 
+Example - Show tags in the target repository with digest-like tags hidden:
+oras repository show-tags localhost:5000/hello --exclude-digest-tag
```
FeynmanZhou, line 48: Considering https://github.com/oras-project/oras/issues/667, can we replace `show-tags` with `tags` in all examples?

qweeah: Sure thing. It will be done in another PR.

shizhMSFT, line 48: The indentation is not correct.

cmd/oras/repository/show-tags.go
```diff
@@ -52,6 +61,7 @@ Example - Show tags of the target repository that include values lexically after
 		},
 	}
 	cmd.Flags().StringVar(&opts.last, "last", "", "start after the tag specified by `last`")
+	cmd.Flags().BoolVar(&opts.excludeDigestTag, "exclude-digest-tags", false, "exclude all digest-like tags")
```
shizhMSFT, line 64:
```suggestion
	cmd.Flags().BoolVar(&opts.excludeDigestTag, "exclude-digest-tags", false, "exclude all digest-like tags such as 'sha256-aaaa...'")
```

cmd/oras/repository/show-tags.go
qweeah: resolved

shizhMSFT (approved):
LGTM

PR #673, opened 2022-11-04
fix: only fallback to OCI image when current media type is OCI artifact

### LINKED ISSUE #672: Check media type of manifest before fallback to OCI image for `oras push` and `oras atatch`
Curious why we chose names that result in double-negatives?  For example *"is it not not-supported"* vs *"is it supported"*.

_Originally posted by @nima in https://github.com/oras-project/oras/pull/665#discussion_r1013118610_

qweeah:
Resolves #672 

### DIFF
--- a/cmd/oras/push.go
+++ b/cmd/oras/push.go
@@ -280,21 +280,26 @@ func skipFallbackToImageManifest(root ocispec.Descriptor, err error) bool {
 	if root.MediaType != ocispec.MediaTypeArtifactManifest {
 		return true
 	}
+
+	return !isManifestUnsupported(err)
+}
+
+func isManifestUnsupported(err error) bool {
 	var errResp *errcode.ErrorResponse
 	if !errors.As(err, &errResp) || errResp.StatusCode != http.StatusBadRequest {
-		return true
+		return false
 	}
 
 	var errCode errcode.Error
 	if !errors.As(errResp, &errCode) {
-		return true
+		return false
 	}
 
 	// As of November 2022, ECR is known to return UNSUPPORTED error when
 	// putting an OCI artifact manifest.
 	switch errCode.Code {
 	case errcode.ErrorCodeManifestInvalid, errcode.ErrorCodeUnsupported:
-		return false
+		return true
 	}
-	return true
+	return false
 }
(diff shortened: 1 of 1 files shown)


cmd/oras/push.go
```diff
@@ -276,22 +276,25 @@ func pushArtifact(dst *remote.Repository, pack packFunc, packOpts *oras.PackOpti
 	return root, nil
 }
 
-func isArtifactUnsupported(err error) bool {
+func noFallbackToOciImage(root ocispec.Descriptor, err error) bool {
+	if root.MediaType != ocispec.MediaTypeArtifactManifest {
+		return true
+	}
```
shizhMSFT, line 282: Nice catch

cmd/oras/push.go
```diff
@@ -276,22 +276,25 @@ func pushArtifact(dst *remote.Repository, pack packFunc, packOpts *oras.PackOpti
 	return root, nil
 }
 
-func isArtifactUnsupported(err error) bool {
+func noFallbackToOciImage(root ocispec.Descriptor, err error) bool {
```
shizhMSFT, line 279: Making `isArtifactUnsupported` to `noFallbackToOciImage` might improve the readability of the invokers but make the readability of the implementation itself worse and easier to have bugs in it.

cmd/oras/push.go
```diff
@@ -276,22 +276,25 @@ func pushArtifact(dst *remote.Repository, pack packFunc, packOpts *oras.PackOpti
 	return root, nil
 }
 
-func isArtifactUnsupported(err error) bool {
+func skipFallbackToImageManifest(root ocispec.Descriptor, err error) bool {
```
Wwwsylvia, line 279: Can we have a documentation for this function?

shizhMSFT (approved):
LGTM

cmd/oras/push.go
qweeah: Do we still need it now after the refactoring?

cmd/oras/push.go
qweeah: Do you still have this concern after the refactoring?

shizhMSFT (approved):
LGTM

PR #675, opened 2022-11-07
feat!: use short command for deletion & add alias for fetch config

### LINKED ISSUE #667: Use shorter commands by default

qweeah:
Resolves #667


sajayantony:
Glad that we have these aliases and it doesn't break anyone. Is there a work item for `manifest get` instead of `manifest fetch` to align with the renames?

qweeah:
> Glad that we have these aliases and it doesn't break anyone. Is there a work item for `manifest get` instead of `manifest fetch` to align with the renames?

I am okay to replace `oras manifest fetch` with `oras manifest get`. cc @shizhMSFT @FeynmanZhou

qweeah:
Closing this PR since `remove` and `delete` means different things: delete means erase (i.e. rendered nonexistent or nonrecoverable), while remove denotes take away and set aside (but kept in existence).

Since user cannot undo `oras manifest/blob delete` via oras commands, we should avoid using `remove` in the example to avoid confusion. But it's okay to add it a alias, which is done in https://github.com/oras-project/oras/pull/677

shizhMSFT:
> Glad that we have these aliases and it doesn't break anyone. Is there a work item for `manifest get` instead of `manifest fetch` to align with the renames?

Nope, the work item isn't there right now. The problem with the word `get` is to find out its opposites in terms of UX.
- `oras manifest get` -- `oras manifest set` does not sound right.
- `oras manifest get` -- `oras manifest put` sounds right but `oras blob put` is a bit strange since blobs are usually large.

For now, we have `oras manifest fetch` -- `oras manifest push`. The idea can be viewed similar to `git fetch` v.s. `git pull` where `fetch` will not change local filesystem but `pull` does. Anyway, we already have `oras manifest get` as an alias of `oras manifest fetch`. So... feel free to use `oras manifest get`.

PR #677, opened 2022-11-07
feat!: alias `remove` to `delete` and rename confirmation flag to `force`

### LINKED ISSUE #667: Use shorter commands by default

qweeah:
Resolves #667


shizhMSFT (approved):
LGTM

PR #689, opened 2022-11-14
feat!: support configurable subject pulling

### LINKED ISSUE #681: ORAS should disable pulling subject when pulling an artifact/image manifest with subject

0.16.0

Currently oras pull brings down the subject which might be GBs in size or not needed. There is no way to exclude bringing down the subject and an option to exclude the subject might be required. 

The other option would be to make sure only blobs are downloaded by default and have a reverse flag of `--include-subject` 

For e.g. - 
➜ oras pull localhost:5000/hello-world@sha256:b7209fa16c0408c4718133eeb6f38d26772aaedfd168f6b8932bc4e951c4c8a1
Downloading 0a017e570e0b sig.txt
Downloading 5891b5b522d5 hello.txt
Downloaded  5891b5b522d5 hello.txt
Downloaded  0a017e570e0b sig.txt
Pulled localhost:5000/hello-world@sha256:b7209fa16c0408c4718133eeb6f38d26772aaedfd168f6b8932bc4e951c4c8a1

oras should enable easy pulling of attached artifacts. If the subject is pulled by default, then it would pull in things like the full image when the intent of the user is only to pull the signature or a vulnerability report. 

My recommendation here is to make the default of pulling only the manifest blobs and not the subject field since its considered as a relationship link even though its technically part of the graph of the merkel tree.

qweeah:
Resolves #681 

### DIFF
--- a/cmd/oras/pull.go
+++ b/cmd/oras/pull.go
@@ -41,6 +41,7 @@ type pullOptions struct {
 	concurrency       int64
 	targetRef         string
 	KeepOldFiles      bool
+	IncludeSubject    bool
 	PathTraversal     bool
 	Output            string
 	ManifestConfigRef string
@@ -84,6 +85,7 @@ Example - Pull all files with concurrency level tuned:
 
 	cmd.Flags().BoolVarP(&opts.KeepOldFiles, "keep-old-files", "k", false, "do not replace existing files when pulling, treat them as errors")
 	cmd.Flags().BoolVarP(&opts.PathTraversal, "allow-path-traversal", "T", false, "allow storing files out of the output directory")
+	cmd.Flags().BoolVarP(&opts.IncludeSubject, "include-subject", "", false, "download subject images/artifacts")
 	cmd.Flags().StringVarP(&opts.Output, "output", "o", ".", "output directory")
 	cmd.Flags().StringVarP(&opts.ManifestConfigRef, "config", "", "", "output manifest config file")
 	cmd.Flags().Int64VarP(&opts.concurrency, "concurrency", "", 3, "concurrency level")
@@ -146,7 +148,7 @@ func runPull(opts pullOptions) error {
 		if err != nil {
 			return nil, err
 		}
-		if subject != nil {
+		if subject != nil && opts.IncludeSubject {
 			nodes = append(nodes, *subject)
 		}
 		if config != nil {


cmd/oras/pull.go
```diff
@@ -56,6 +57,10 @@ func pullCmd() *cobra.Command {
 Example - Pull all files:
   oras pull localhost:5000/hello:latest
 
+
+Example - Recursively pulling all files, including subjects of hello:latest:
+  oras pull localhost:5000/hello:latest --include-subject
```
shizhMSFT, line 62: Put flags before arguments.

cmd/oras/pull.go
```diff
@@ -84,6 +89,7 @@ Example - Pull all files with concurrency level tuned:
 
 	cmd.Flags().BoolVarP(&opts.KeepOldFiles, "keep-old-files", "k", false, "do not replace existing files when pulling, treat them as errors")
 	cmd.Flags().BoolVarP(&opts.PathTraversal, "allow-path-traversal", "T", false, "allow storing files out of the output directory")
+	cmd.Flags().BoolVarP(&opts.IncludeSubject, "include-subject", "", false, "download files recursively from subject images/artifacts")
```
shizhMSFT, line 92:
```suggestion
	cmd.Flags().BoolVarP(&opts.IncludeSubject, "include-subject", "", false, "recursively pull the subject of artifacts")
```

cmd/oras/pull.go
```diff
@@ -84,6 +89,7 @@ Example - Pull all files with concurrency level tuned:
 
 	cmd.Flags().BoolVarP(&opts.KeepOldFiles, "keep-old-files", "k", false, "do not replace existing files when pulling, treat them as errors")
 	cmd.Flags().BoolVarP(&opts.PathTraversal, "allow-path-traversal", "T", false, "allow storing files out of the output directory")
+	cmd.Flags().BoolVarP(&opts.IncludeSubject, "include-subject", "", false, "download files recursively the subject of artifacts")
```
shizhMSFT, line 92: error: invalid grammar

shizhMSFT (approved):
LGTM

shizhMSFT:
Let's merge after the release of `v0.16.1`.

sajayantony:
I was hoping this would be there in the next release. The pulling of all related content is a bit confusing to me honest.

FeynmanZhou:
> I was hoping this would be there in the next release. The pulling of all related content is a bit confusing to be honest when you want to just pull the attached SBOM.

@sajayantony Agree, I think this PR is a breaking change that has changed the default behavior of `oras pull`. So it is more reasonable to bring it in the RC.1 instead of a patch release (v0.16.1).

shizhMSFT:
Merging to `main` since we can cherry-pick fixes to `release-0.16` branch if we need to release `0.16.1`.

PR #693, opened 2022-11-17
test: add e2e specs for `oras pull`

### LINKED ISSUE #567: e2e: add test specs for `oras pull` command
Happy paths:
- should pull all files in an image to a target folder
- should also pull config file with `--config` option
- should pull specific platform with `--platform` option

should show correct pull status for all the pulling

Bad paths:
- ~~should fail if path traversal is needed but not enabled~~
- should fail if file overwrite is needed but not enabled  
- should fail directly if not logged in
- should fail if no artifact reference provided

qweeah:
Resolves: #567


test/e2e/suite/command/pull.go
```diff
@@ -0,0 +1,110 @@
...
+
+		It("should skip config if media type not matching", func() {
+			pullRoot := "pulled"
+			tempDir := GinkgoT().TempDir()
+			if err := CopyTestData(tempDir); err != nil {
+				panic(err)
+			}
+			ORAS("pull", Reference(Host, repo, tag), "-v", "--config", fmt.Sprintf("%s:%s", files[0], "???"), "-o", pullRoot).
+				MatchStatus([]match.StateKey{
+					{Digest: "fd6ed2f36b54", Name: "application/vnd.oci.image.manifest.v1+json"},
```
shizhMSFT, line 83: same here

shizhMSFT (approved):
LGTM

PR #696, opened 2022-11-21
feat: support customized DNS resolving for remote registry

### LINKED ISSUE #688: Support customized DNS resolving from a `host` to a specific `address`

_No response_

I would like to have a new flag added to all remote registry related commands:
```
--resolve <host:port:address>
```

So any outgoing request towards `host`:`port` should be resolve to `address`:`port`.

So ORAS can be 1) used to debug registry services running locally; 2) hardcode DNS resolving to reduce DNS bouncing in geo-replicated registry services.

qweeah:
Resolves #688


cmd/oras/internal/option/remote.go
```diff
@@ -109,21 +151,42 @@ func (opts *Remote) tlsConfig() (*tls.Config, error) {
...
+		if addr == fmt.Sprintf("%s:%d", r.from, r.port) {
+			matched = r
+			break
+		}
+	}
+	if matched == nil {
+		return defaultDialer.DialContext(ctx, network, addr)
+	}
+	return net.DialTCP(network, nil, &net.TCPAddr{IP: matched.to, Port: matched.port})
+}
```
shizhMSFT, line 173: The `net.DialTCP` does not accept a context. How about
```go
type resolveDialer struct {
	base   *net.Dialer
	resolve map[string]string
}

func (rd *resolveDialer) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
	if resolved, ok := rd.resolve[addr]; ok {
		addr = resolved
	}
	return rd.base.DialContext(ctx, network, addr)
}
```
Will above code work?

qweeah: It surely works, but using mapping surely reduces extensibility for supporting ipv6. Resolved accordingly.

cmd/oras/internal/option/remote.go
```diff
@@ -35,15 +36,24 @@ import (
 	"oras.land/oras/internal/version"
 )
 
+type ResolveEntry struct {
```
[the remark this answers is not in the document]
qweeah: removed and changed to map

cmd/oras/internal/option/remote.go
qweeah: Checked it's non-relavant.

cmd/oras/internal/option/remote.go
```diff
@@ -109,21 +143,37 @@ func (opts *Remote) tlsConfig() (*tls.Config, error) {
...
+// DialContext connects to the addr on the named network using
+// the provided context.
+func (opts *Remote) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	for k := range opts.resolve {
+		if k == addr {
+			addr = opts.resolve[k]
+		}
+	}
+	return defaultDialer.DialContext(ctx, network, addr)
+}
```
shizhMSFT, line 160: Can we decouple the dialer with the opts?

shizhMSFT (approved):
LGTM

PR #698, opened 2022-11-21
fix!: fix colon in media type for windows

### LINKED ISSUE #697: oras push/attach should not allow colon in media type only on Windows
Using oras 0.16.0, I can use below command to upload a layer with colon in the media type on Windows:
```
oras push localhost:5000/test:path .\KEYS:test:test:test
```

See layers.mediatype below, this command will generate a layer with `test:test:test` as an invalid media type
```
> oras manifest fetch localhost:5000/test:path --pretty
{
  "schemaVersion": 2,
  "config": {
    "size": 2
  },
  "layers": [
    {
      "mediaType": "test:test:test",
      "digest": "sha256:b79b04d85ebb8b289bde13387a5c6f3d98448d5a51937e6f94239303ffc2add8",
      "size": 15020,
      "annotations": {
        "org.opencontainers.image.title": "KEYS"
      }
    }
  ]
}
```

According to [rfc2045 ](https://datatracker.ietf.org/doc/html/rfc2045#section-5.1), colon is a special character and should not be in the media type. So the push command should show an error same to linux/darwin build:
```
> oras push localhost:5000/test:path .\KEYS:test:test:test
Error: failed to stat .KEYS:test:test: stat .KEYS:test:test: no such file or directory
```

qweeah:
Fixes #697 


cmd/oras/internal/argument/file_windows_test.go
```diff
@@ -0,0 +1,51 @@
...
+		wantMediatype string
+	}{
+		{"file name and media type in reference", args{"a:b", "c"}, "a", "b"},
+		{"media type in reference", args{":b", "c"}, "", "b"},
+		{"file name and empty media type in reference", args{"a:", "c"}, "a", ""},
+		{"file name in reference", args{"a", "c"}, "a", "c"},
+		{"file name in reference, no default", args{"a:", ""}, "a", ""},
+		{"windows file name in reference with default media type", args{`a:\b`, "c"}, `a:\b`, "c"},
+		{"windows file name and media type in reference", args{`a:\b:c`, "d"}, `a:\b`, "c"},
+		{"file name and media type in reference", args{"a:b:c", "d"}, "a:b", "c"},
```
Wwwsylvia, line 38: Do we need to test something like:

- `a:b:` (second colon at the end)
- `:a:b` (first colon at the beginning)
- `:` (single colon)
- `::` (double colons)
- `1:\` (first character is not a letter)

They are not valid windows paths though.

qweeah: Yes they are needed. I also comeup with one `ab:\c:d`

Wwwsylvia (approved):
LGTM

cmd/oras/internal/argument/file_unix.go
```diff
@@ -15,12 +15,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package argument
```
shizhMSFT, line 18: package name `argument` seems weird. How about package `file` with func `ParseReference`?

cmd/oras/internal/argument/file_windows.go
```diff
@@ -13,28 +13,24 @@ See the License for the specific language governing permissions and
...
+package argument
 
 import (
 	"strings"
 	"unicode"
 )
 
-// parseFileReference parses file reference on windows.
+// ParseFileReference parses file reference on windows.
 // Windows systems does not allow ':' in the file path except for drive letter.
```
shizhMSFT, line 24: Windows does not allow `:`. Therefore, we should return error if the reference is invalid.

cmd/oras/internal/argument/file_unix.go
qweeah: You are right. It's related to file reference, now used in argument parsing, can be applied to others like flag parsing if we do have a file reference flag.

cmd/oras/internal/argument/file_windows.go
qweeah: Checked special reserved characters described in https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

cmd/oras/internal/fileref/windows.go
```diff
@@ -0,0 +1,52 @@
...
+package fileref
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+	"unicode"
+)
+
+// Parse parses file reference on unix.
```
Wwwsylvia, line 27: Windows?

cmd/oras/internal/fileref/unix.go
```diff
@@ -15,16 +15,15 @@ See the License for the specific language governing permissions and
...
 */
 
-package main
+package fileref
 
 import "strings"
 
-// parseFileReference parses file reference on unix.
-func parseFileReference(reference string, mediaType string) (filePath, mediatype string) {
+//Parse parses file reference on unix.
```
shizhMSFT, line 22: nit:
```suggestion
// Parse parses file reference on unix.
```

cmd/oras/internal/fileref/windows_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		{"colon file name and media type", args{"az:b:c", "d"}, "az:b", "c"},
+		{"colon file name and empty media type", args{"az:b:", "c"}, "az:b", ""},
+		{"colon-prefix file name and media type", args{":az:b:c", "d"}, ":az:b", "c"},
+
+		{"pure colon file name and media type", args{"::a", "b"}, ":", "a"},
+		{"pure colon file name and empty media type", args{"::", "a"}, ":", ""},
+
+		{"windows file name1 and default type", args{`a:\b`, "c"}, `a:\b`, "c"},
+		{"windows file name2 and default type", args{`z:b`, "c"}, `Z:\b`, "c"},
```
shizhMSFT, line 53: `z:b` should be parsed as `z` and `b`.

cmd/oras/internal/fileref/windows_test.go
```diff
@@ -0,0 +1,115 @@
...
+		{"windows file name and media type", args{`a:\b:c`, "d"}, `a:\b`, "c"},
+		{"windows file name and empty media type", args{`a:\b:`, "c"}, `a:\b`, ""},
+		{"windows file name and empty media type", args{`a:\b:`, "c"}, `a:\b`, ""},
```
shizhMSFT, line 56: duplicated test cases

cmd/oras/internal/fileref/windows_test.go
qweeah: There is a workaround like `./z:b` but still `z:b` goes wrong

qweeah: reconfirmed, `c:z:b` not working since we cut the media type from the left most colon. Maybe that's why we used to cut from right?

cmd/oras/internal/fileref/windows.go
```diff
@@ -0,0 +1,51 @@
...
+	"path/filepath"
+	"strings"
+	"unicode"
+)
+
+// Parse parses file reference on windows.
+func Parse(reference string, mediaType string) (filePath, mediatype string, err error) {
+	filePath, mediatype = doParse(reference, mediaType)
+	if strings.ContainsAny(filePath, `<>:"|?*`) {
+		// https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions
```
shizhMSFT, line 31: nit: 
```suggestion
		// Reference: https://learn.microsoft.com/windows/win32/fileio/naming-a-file#naming-conventions
```

cmd/oras/internal/fileref/windows.go
```diff
@@ -0,0 +1,51 @@
...
+		return "", "", fmt.Errorf("Reserved characters found in the file path: %s", filePath)
```
shizhMSFT, line 32: nit: error style
```suggestion
		return "", "", fmt.Errorf("reserved characters found in the file path: %s", filePath)
```

cmd/oras/internal/fileref/windows_test.go
```diff
@@ -0,0 +1,114 @@
...
+		wantFilePath  string
+		wantMediatype string
+	}{
+		{"file name and media type", args{"az:b", ""}, "az", "b"},
+		{"file name and empty media type", args{"az:", ""}, "az", ""},
+		{"file name and default media type", args{"az", "c"}, "az", "c"},
+		{"file name and media type, default type ignored", args{"az:b", "c"}, "az", "b"},
+		{"file name and empty media type, default type ignored", args{"az:", "c"}, "az", ""},
+
+		{"empty file name and media type", args{":a", "b"}, "", "a"},
```
shizhMSFT, line 41: `Parse()` now returns error. Since empty file names are not valid, should we return errors instead?

cmd/oras/internal/fileref/windows.go
```diff
@@ -0,0 +1,51 @@
...
+	i := strings.LastIndex(reference, ":")
+	if i < 0 {
+		return reference, mediaType
+	}
+	if i == 1 && len(reference) > 2 && unicode.IsLetter(rune(reference[0])) && reference[2] == '\\' {
+		// In case it is C:\
+		// Relative file path with disk prefix is NOT supported, e.g. `c:file1`
+		if abs, err := filepath.Abs(reference); err == nil {
+			return abs, mediaType
+		}
```
shizhMSFT, line 47: Since relative file path is not supported, it seems we don't need to resolve absolute path as it already is.

cmd/oras/internal/fileref/windows_test.go
qweeah: this is the test for doParse, sanctity check will be done in the Parse tests。

cmd/oras/internal/fileref/unix.go
```diff
@@ -15,16 +15,21 @@ See the License for the specific language governing permissions and
...
+import (
+	"strings"
+	"fmt"
+)
+// Parse parses file reference on unix.
+func Parse(reference string, mediaType string) (filePath, mediatype string, err error) {
 	i := strings.LastIndex(reference, ":")
 	if i < 0 {
-		return reference, mediaType
+		return reference, mediaType, nil
```
shizhMSFT, line 28: here `reference` can be empty

shizhMSFT: it is not covered by unit test

cmd/oras/internal/fileref/unix.go
```diff
@@ -15,16 +15,27 @@ See the License for the specific language governing permissions and
...
-// parseFileReference parses file reference on unix.
-func parseFileReference(reference string, mediaType string) (filePath, mediatype string) {
-	i := strings.LastIndex(reference, ":")
-	if i < 0 {
-		return reference, mediaType
+// Parse parses file reference on unix.
+func Parse(reference string, defaultMediaType string) (filePath, mediaType string, err error) {
+	if reference == "" {
+		return "", "", nil
 	}
```
shizhMSFT, line 29: no input should return error, shouldn't it?

cmd/oras/internal/fileref/unix.go
```diff
@@ -15,16 +15,27 @@ See the License for the specific language governing permissions and
...
-	return reference[:i], reference[i+1:]
 
+	filePath = reference
+	mediaType = defaultMediaType
+	i := strings.LastIndex(reference, ":")
+	if i >= 0 {
+		filePath, mediaType = reference[:i], reference[i+1:]
+	}
```
shizhMSFT, line 36:
```suggestion
	i := strings.LastIndex(reference, ":")
	if i >= 0 {
		filePath, mediaType = reference[:i], reference[i+1:]
	} else {
		filePath, mediaType  = reference, defaultMediaType	
	}
```

shizhMSFT (approved):
LGTM

PR #699, opened 2022-11-22
test: add e2e specs for `oras blob fetch`

### LINKED ISSUE #573: e2e: add test specs for `oras blob fetch` command
Happy paths:
- should fetch blob content and output to a file
- should fetch blob descriptor
- should fetch prettified blob descriptor
- should fetch blob content to a file with prettified descriptor in stdout
- should fetch blob content and output to stdout
- ~~progress status should be disabled when `--output -` or `--descriptor` is provided~~
- should show preview hint in help doc

Bad paths:
- should fail if no subject reference is provided
- should fail if neither output path nor descriptor flag is provided
- should fail if `--output -` and `--descriptor` flag are both provided
- should fail directly if not logged in

qweeah:
Resolves #573 

This PR relies on #701


cmd/oras/internal/option/pretty_test.go
```diff
@@ -38,16 +35,8 @@ func TestPretty_ApplyFlags(t *testing.T) {
```
qweeah, line 35: Since it's hard to do prettified content checking in E2E test, I enforced unit tests for pretty-related unit tests.

shizhMSFT (approved):
LGTM

PR #705, opened 2022-11-25
fix: add log when failed to scan confirmation

### LINKED ISSUE #704: manifest and blob deletion might be silently cancelled without logging

`oras manifest delete` and `oras blob delete` can be cancelled without logging if the stdin stream is closed before scanning

Should output `Operation cancelled` if deletion is not confirmed via stdin or flag.

### How can we reproduce it?

Save below snippet to test.go, Replace <your-image> with a image name  and run it via `go run test.go`

```golang
package main

import (
	"fmt"
	"io"
	"os"
	"os/exec"
	"strings"
)

func main() {
	var cmd *exec.Cmd

	fmt.Println("Should exit with no logging")
	stdin := io.NopCloser(strings.NewReader(""))
	cmd = exec.Command("oras", "manifest", "delete", "<your-image>")
	cmd.Stderr = os.Stderr
	cmd.Stdout = os.Stdout
	cmd.Stdin = stdin
	stdin.Close()
	cmd.Start()
	cmd.Wait()
	fmt.Println()

	fmt.Println("Should exit with error logging")
	stdin = io.NopCloser(strings.NewReader("n"))
	cmd = exec.Command("oras", "manifest", "delete", "<your-image>")
	cmd.Stderr = os.Stderr
	cmd.Stdout = os.Stdout
	cmd.Stdin = stdin
	defer stdin.Close()
	cmd.Start()
	cmd.Wait()
	fmt.Println()
}

```

0.16.0

 Ubuntu 20.04

qweeah:
Resolves #704 


cmd/oras/internal/option/confirmation.go
```diff
@@ -45,7 +45,12 @@ func (opts *Confirmation) AskForConfirmation(r io.Reader, prompt string) (bool,
...
 	if ok := scanner.Scan(); !ok {
-		return false, scanner.Err()
+		err := scanner.Err()
+		if err == nil {
+			// EOF
+			fmt.Println("Operation cancelled.")
+		}
+		return false, err
 	}
 	response := scanner.Text()
```
shizhMSFT, line 55: How about following the examples in the doc of [bufio.Scanner](https://pkg.go.dev/bufio#Scanner)?
```go
var response string
scanner := bufio.NewScanner(r)
if ok := scanner.Scan(); ok {
	response := scanner.Text()
}
if err := scanner.Err(); err != nil {
	return false, err
}
```

qweeah: Let me check if scanning EOF breaks `strings.ToLower(response)`. If not, this is is definitely better.

qweeah: updated

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #709, opened 2022-12-01
ci: parameterize registry and add zot as another candidate

rchincha:
Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -1,0 +1,72 @@
+# Copyright The ORAS Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: ecosystem
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+  pull_request:
+    branches:
+      - main
+      - release-*
+jobs:
+  zot:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Setup
+        run: |
+          # upload images, zot can serve OCI image layouts directly like so
+          mkdir /tmp/zot
+          skopeo copy --format=oci docker://busybox:latest oci:/tmp/zot/busybox:latest
+          # start zot
+          docker pull ghcr.io/project-zot/zot-minimal-linux-amd64:latest
+          docker run -d --name zot -v /tmp/zot:/var/lib/registry -p 8000:5000 ghcr.io/project-zot/zot-minimal-linux-amd64:latest
+          # wait until service is up
+          while true; do x=0; curl -f http://$ORAS_REGISTRY_HOST/v2/ || x=1; if [ $x -eq 0 ]; then break; fi; sleep 1; done
+        env:
+          ORAS_REGISTRY_HOST: localhost:8000
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go-version }}
+          check-latest: true
+      - name: Build CLI
+        run: make build-linux-amd64
+      - name: Check Version
+        run: bin/linux/amd64/oras version
+      - name: Run Compatibility Tests
+        run: |
+          cd $GITHUB_WORKSPACE/
+          echo "{\"name\":\"foo\",\"value\":\"bar\"}" > config.json
+          echo "hello world" > artifact.txt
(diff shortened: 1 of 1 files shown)


rchincha:
Looks like there is very docker-specific env setup. Should probably standardize on that also.

qweeah:
Also specs related docker images will be covered in ORAS E2E tests, @rchincha how to run those based on ZOT?

FeynmanZhou:
@rchincha Hi Ramkumar, it seems this PR make the E2E testing fail. It has some tech limitations to adding Zot to the CI.

We will add the Zot registry to the ORAS quick start, see https://github.com/oras-project/oras-www/issues/71.

rchincha:
> @rchincha Hi Ramkumar, it seems this PR make the E2E testing fail. It has some tech limitations to adding Zot to the CI.
> 
> BTW, I would like to add the Zot registry to the ORAS quick start, see [oras-project/oras-www#71](https://github.com/oras-project/oras-www/issues/71).

Trying to understand the test setup a bit more. It appears data is manually loaded, which is unlikely to work with every registry.

rchincha:
https://github.com/oras-project/oras-www/issues/71#issuecomment-1335662117

qweeah:
> Trying to understand the test setup a bit more. It appears data is manually loaded, which is unlikely to work with every registry.

Yes, once oras CLI supports OCI layout, we can store test data in OCI layout. I created an [issue ](https://github.com/oras-project/oras/issues/710) for that.

rchincha:
> > Trying to understand the test setup a bit more. It appears data is manually loaded, which is unlikely to work with every registry.
> 
> Yes, once oras CLI supports OCI layout, we can store test data in OCI layout. I created an [issue ](https://github.com/oras-project/oras/issues/710) for that.

You can also do:

```
skopeo copy --format=oci docker://busybox:latest oci:/tmp/vol/busybox:latest

docker run -v /tmp/vol:/var/lib/registry -p 5000:5000 ghcr.io/project-zot/zot-minimal-linux-amd64:latest
```

Basically two parts:
1) registry-specific - storage, authN, etc
2) registry-agnostic - OCI conformance really

shizhMSFT:
@rchincha Thanks for creating this PR for proposing zot as a candidate registry for oras E2E testing.

In our current plan, we are using [github.com/oras-project/distribution](https://github.com/oras-project/distribution), which is a specially crafted fork of [github.com/distribution/distribution](https://github.com/distribution/distribution), for the E2E testing.

To achieve the best quality of E2E testing, we can't use `oras` to populate the data to be tested since `oras` itself can be flawed. It is also not a good idea to use other tools to populate data since the other tools may not support oras specific features and they also need to be well tested. That's why we manually prepare test data for the distribution, which is unfortunately distribution specific. Maybe in the future, we can use the older version of oras to populate data to test the latest version of oras just like [compiler bootstrapping](https://en.wikipedia.org/wiki/Bootstrapping#Computing).

`zot` is a good candidate for oras E2E testing. However, it is an OCI-only registry. If it can support docker manifests or even custom manifests, we can definitely use `zot` for oras E2E testing.

shizhMSFT:
> Otherwise oras' contract expands to say that only registries which support docker media-types work with oras.

This statement is not true. `oras` is designed to be compatible with [OCI-compliant](https://github.com/opencontainers/distribution-spec/blob/main/spec.md) registries and [Docker-compliant](https://docs.docker.com/registry/spec/api/) registries. In other words,
- If a registry is docker-only, `oras` is compatible with it.
- If a registry is OCI-only, `oras` is compatible with it.
- If a registry is compliant with docker and OCI, `oras` is compatible with it.
- If a registry is partially compliant with docker or OCI specs, `oras` may still be compatible with it. For example, if we want to experiment a new manifest type, `oras` can still work with that manifest type even if it is not defined by oci or docker specs.

That's why we made a testing registry ([github.com/oras-project/distribution](https://github.com/oras-project/distribution)) and only for testing, which supports OCI and Docker. Although it is not suitable for production, we can easily modify it to test all scenarios. Again, we cannot use Docker-only registries since they don't support OCI. Similarly, we cannot use OCI-only registries since they don't support Docker.

rchincha:
@shizhMSFT thanks for the clarification. Once this PR passes CI, maybe it can parameterized to included other registries - docker or OCI.

rchincha:
Fixed the PR. Hopefully should pass the CI checks this time.

rchincha:
https://github.com/rchincha/oras/pull/1
^ test this action here, including the attach example. Pls wait this passes.

rchincha:
https://github.com/rchincha/oras/pull/1
^ now ready, pls note oras temporarily pinned to v0.16.0 due to https://github.com/oras-project/oras/issues/719 so remove that section later.

rchincha:
PR updated after the oras-go bump.

rchincha:
https://github.com/notaryproject/notation/pull/493
^ do you plan to do the same wrt this PR?

shizhMSFT:
> [notaryproject/notation#493](https://github.com/notaryproject/notation/pull/493) ^ do you plan to do the same wrt this PR?

/cc @qweeah

.github/workflows/ecosystem.yml
```diff
@@ -0,0 +1,72 @@
...
+  zot:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Setup
+        run: |
+          # upload images, zot can serve OCI image layouts directly like so
+          mkdir /tmp/zot
+          skopeo copy --format=oci docker://busybox:latest oci:/tmp/zot/busybox:latest
```
TerryHowe, line 35: Seems like we could do this later with oras

qweeah:
Closing as ZOT is already used as E2E backend since v1.1.0

PR #734, opened 2023-01-04
fix!: allow colon and back slash in unix file name

### LINKED ISSUE #695: ORAS attach seems to not support colon in filename

I have a file named `simple-server:v0.0.1-linux-amd64` I want to attach this to an image using ORAS. I ran:

I ran:

```
oras attach --artifact-type application/spdx+json upstream.azurecr.io/public/oss/kubernetes/simple-server:v0.0.1-linux-amd64 simple-server:v0.0.1-linux-amd64
```

The error I got was:

```
Error: failed to stat simple-server: stat simple-server: no such file or directory
```

If I rename the file to remove the colon, things work as expected. 

The file is attached to the image.  Linux filenames can contain colons, so this seems like a usability issue for folks using Linux or Mac. 

### How can we reproduce it?

Try to use ORAS attach with a file that has a colon in it. 

v0.15.0 and v0.16.0

Mac OS

qweeah:
Resolves #695

To provide a file name with colon(`:`) and media-type at the same time, we can use backslash(`\`) to escape the colon in the file name, taking `oras push` as an example:
```
oras push some.reg/repo:tag "my\:file:my.media.type" # will push a file named `my:file` with mediatype `my.media.type`
```

Backslash can be used to escape itself:

```
oras push some.reg/repo:tag "my\\file:my.media.type" # will push a file named `my\file` with mediatype `my.media.type`
```

The old workaround still works after this change:
```
oras push some.reg/repo:tag my:file:my.media.type: # will push a file named `my:file` with mediatype `my.media.type`
```

### DIFF
--- a/cmd/oras/internal/fileref/unix.go
+++ b/cmd/oras/internal/fileref/unix.go
@@ -43,7 +43,6 @@ func Parse(reference string, defaultMediaType string) (filePath, mediaType strin
 	return filePath, mediaType, nil
 }
 
-// isEscaped returns if the character in path with offset is escaped by '\'.
 func isEscaped(path string, offset int) bool {
 	cnt := 0
 	for i := offset - 1; i >= 0; i-- {
--- a/cmd/oras/internal/fileref/unix.go
+++ b/cmd/oras/internal/fileref/unix.go
@@ -24,14 +24,50 @@ import (
 
 // Parse parses file reference on unix.
 func Parse(reference string, defaultMediaType string) (filePath, mediaType string, err error) {
-	i := strings.LastIndex(reference, ":")
-	if i < 0 {
-		filePath, mediaType = reference, defaultMediaType
+	i := len(reference)
+	for {
+		// found the right most colon which is not escaped
+		i = strings.LastIndex(reference[:i], ":")
+		if i < 0 || !isEscaped(reference, i) {
+			break
+		}
+	}
+	if i < 0 || isEscaped(reference, i) {
+		filePath, mediaType = unescape(reference), defaultMediaType
 	} else {
-		filePath, mediaType = reference[:i], reference[i+1:]
+		filePath, mediaType = unescape(reference[:i]), reference[i+1:]
 	}
 	if filePath == "" {
 		return "", "", fmt.Errorf("found empty file path in %q", reference)
 	}
 	return filePath, mediaType, nil
+}
+
+// isEscaped returns if the character in path with offset is escaped by '\'.
+func isEscaped(path string, offset int) bool {
+	cnt := 0
+	for i := offset - 1; i >= 0; i-- {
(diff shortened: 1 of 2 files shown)


cmd/oras/internal/fileref/unix.go
```diff
@@ -24,14 +24,49 @@ import (
...
 }
+
+func isEscaped(path string, offset int) bool {
+	cnt := 0
+	for i := offset - 1; i >= 0; i-- {
+		if path[i] != '\\' {
+			break
+		}
+	}
+	return cnt%2 != 0
```
wangxiaoxuan273, line 53: Isn't `cnt` always 0? Its value is not modified in the loop.

qweeah: Good catch! Added it with another unit test case.

cmd/oras/internal/fileref/unix_test.go
```diff
@@ -35,7 +35,9 @@ func Test_ParseFileReference(t *testing.T) {
 		{"file name and default media type", args{"az", "c"}, "az", "c"},
 		{"file name and media type, default type ignored", args{"az:b", "c"}, "az", "b"},
 		{"file name and empty media type, default type ignored", args{"az:", "c"}, "az", ""},
-		{"colon file name and media type", args{"az:b:c", "d"}, "az:b", "c"},
+		{"colon file name and media type", args{`az\:b:c`, "d"}, "az:b", "c"},
+		{"colon file name with backslash and media type1", args{`az\\\:b:c`, "d"}, `az\:b`, `c`},
+		{"colon file name with backslash and media type2", args{`az\\\\:b`, "c"}, `az\\`, `b`},
```
shizhMSFT, line 40: missing a case, which is reported by the users.
```
sha256\:xxxxx....
```

cmd/oras/internal/fileref/unix_test.go
```diff
@@ -88,3 +90,26 @@ func TestParse(t *testing.T) {
...
+		{"double backslash", `\\`, `\`},
+		{"escaping t", `\t\t`, "tt"},
+		{"not escaping 1st t", `\\t\t`, `\tt`},
+		{"not escaping 2nd t", `\t\\t`, `t\t`},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := unescape(tt.input)
+			if got != tt.want {
+				t.Errorf("Parse() gotFilePath = %v, want %v", got, tt.want)
```
shizhMSFT, line 111:
```suggestion
				t.Errorf("unescape() = %v, want %v", got, tt.want)
```

TerryHowe:
Merge conflict

qweeah:
Close as it's easier to use the workaround

PR #736, opened 2023-01-06
refactor(oci-layout): use Parser to process option input 

qweeah:
This PR adds a new option parser which can be used to process input before cmd execution. This change will be used in OCI layout support.


cmd/oras/internal/option/platform.go
```diff
@@ -26,25 +24,26 @@ import (
 
 // Platform option struct.
 type Platform struct {
-	Platform string
+	platform    string
+	OCIPlatform *ocispec.Platform
```
shizhMSFT, line 28: The name `OCIPlatform` looks strange since all `OCIxxx` are top class in oras. It is better to name it `Platform`.

cmd/oras/pull.go
```diff
@@ -125,9 +121,7 @@ func runPull(opts pullOptions) error {
 			return err
 		}
 	}
-	if targetPlatform != nil {
-		copyOptions.WithTargetPlatform(targetPlatform)
-	}
+	copyOptions.WithTargetPlatform(opts.OCIPlatform)
```
shizhMSFT, line 124: nit: Although `WithTargetPlatform` supports `nil` input, it is better to have an `if` statement just for readability.

cmd/oras/internal/option/remote.go
```diff
@@ -87,8 +87,8 @@ func (opts *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
 	}
 }
 
-// ReadPassword tries to read password with optional cmd prompt.
-func (opts *Remote) ReadPassword() (err error) {
+// Parse tries to read password with optional cmd prompt.
+func (opts *Remote) Parse() error {
```
shizhMSFT, line 91: The method name `Parse` is reasonable for `Platform` but not for `Remote`.

How about
```go
// Parse bla bla bla
func (opts *Remote) Parse() error {
    return opts.readPassword()
}

// readPassword tries to read password with optional cmd prompt.
func (opts *Remote) readPassword() error {
    // ...
}
```

cmd/oras/internal/option/parser.go
```diff
@@ -0,0 +1,41 @@
...
+			iface := f.Addr().Interface()
+			if a, ok := iface.(FlagParser); ok {
+				if err := a.Parse(); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
qweeah: The tricky part is applying flags won't return error, but we still can make function argument `fn` with type `func(T, *error)` to resolve that.

cmd/oras/pull.go
qweeah: Should we also enforce the nil check for fetch option and resolve option?

cmd/oras/internal/option/parser.go
```diff
@@ -0,0 +1,51 @@
...
+// error during parsing.
+func Parse(optsPtr interface{}) error {
+	return rangeFields(optsPtr, func(fp FlagParser, err *error) {
+		*err = fp.Parse()
+	})
+}
+
+// rangeFields goes through all fields of ptr, optionally run fn if a field is
+// public AND typed T.
+func rangeFields[T any](ptr any, fn func(T, *error)) error {
```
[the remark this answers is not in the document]
qweeah: Yes, after some thoughts you are right.

shizhMSFT (approved):
LGTM

PR #745, opened 2023-01-11
feat!: force deleting a non-existent blob/manifest

### LINKED ISSUE #722: return success when force deleting a non-existent blob/manifest
When force deleting via `oras manifest/blob delete`, if the target reference is not found in the remote registry, then user will see an error message like below:

```
$ oras blob delete localhost:5000/command/blob/delete/1671102430@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa --force
Error: localhost:5000/command/blob/delete/1671102430@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa: the specified blob does not exist
```

The returned value should indicate success if successfully resolving the reference and registry returns not found.

@FeynmanZhou

qweeah:
Resolves #722 

### DIFF
--- a/cmd/oras/internal/option/confirmation.go
+++ b/cmd/oras/internal/option/confirmation.go
@@ -26,18 +26,18 @@ import (
 
 // Confirmation option struct.
 type Confirmation struct {
-	Confirmed bool
+	Force bool
 }
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Confirmation) ApplyFlags(fs *pflag.FlagSet) {
-	fs.BoolVarP(&opts.Confirmed, "force", "f", false, "do not prompt for confirmation")
+	fs.BoolVarP(&opts.Force, "force", "f", false, "do not prompt for confirmation")
 }
 
 // AskForConfirmation prints a propmt to ask for confirmation before doing an
 // action and takes user input as response.
 func (opts *Confirmation) AskForConfirmation(r io.Reader, prompt string) (bool, error) {
-	if opts.Confirmed {
+	if opts.Force {
 		return true, nil
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/oras/internal/option/confirmation.go
```diff
@@ -26,18 +26,18 @@ import (
...
 // Confirmation option struct.
 type Confirmation struct {
-	Confirmed bool
+	Force bool
 }
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Confirmation) ApplyFlags(fs *pflag.FlagSet) {
-	fs.BoolVarP(&opts.Confirmed, "force", "f", false, "do not prompt for confirmation")
+	fs.BoolVarP(&opts.Force, "force", "f", false, "do not prompt for confirmation")
```
shizhMSFT, line 34: nit: Should we consider `ignore non-existent references, never prompt` over `do not prompt for confirmation`? /cc @FeynmanZhou

shizhMSFT (approved):
LGTM with suggestions.

cmd/oras/internal/option/confirmation.go
qweeah: Looks good. Also nonexistent [is a word](https://www.merriam-webster.com/dictionary/nonexistent) and is already used by `rm` manual.

FeynmanZhou: Agree with @shizhMSFT 's suggestion. I see @qweeah has incorporated it into the confirmation.

shizhMSFT (approved):
LGTM

Issue #378, opened 2022-05-07
Push / Pull Artifacts from OCI Image Layout


shizhMSFT:
ORAS is evolving into a generic registry client along with `oras-go` v2. With the support of `oras-go` v2, the `oras` CLI should be able to push / pull artifacts from [OCI image layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md).

By resolving this issue, the following issue should be auto resolved.
- #244
- #245 
- #453

sajayantony:
Can we discuss some candidate commands that would pull to tar and push the tar files?

qweeah:
@sajayantony I added a hackmd doc, please help review the session[ Pushing and Pulling via oras copy](https://hackmd.io/ogglvwhLTVW2-T6b2JC_Gg#Pushing-and-Pulling-via-oras-copy)

qweeah:
@FeynmanZhou Thanks for reminding. It's shared it, could you also help review it.

qweeah:
Closing as it's already supportted in following commands:
- oras cp
- oras push
- oras pull
- oras attach
- oras manifest push
- oras blob push
- oras manifest fetch
- oras manifest fetch-config
- oras blob fetch

PR #750, opened 2023-01-16
feat(oci-layout): support in `oras push` and `oras attach`

qweeah:
Related to https://github.com/oras-project/oras/issues/378

### DIFF
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -25,18 +25,18 @@ import (
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/content/file"
+	"oras.land/oras-go/v2/registry/remote"
 	oerrors "oras.land/oras/cmd/oras/internal/errors"
 	"oras.land/oras/cmd/oras/internal/option"
 )
 
 type attachOptions struct {
 	option.Common
-	option.Remote
 	option.Packer
 	option.ImageSpec
 	option.DistributionSpec
+	option.Target
 
-	targetRef    string
 	artifactType string
 	concurrency  int
 }
@@ -75,11 +75,17 @@ Example - Attach file 'hi.txt' and export the pushed manifest to 'manifest.json'
 `,
 		Args: cobra.MinimumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return option.Parse(&opts)
+			opts.RawReference = args[0]
+			opts.FileRefs = args[1:]
+			if err := option.Parse(&opts); err != nil {
+				return err
+			}
+			if opts.Type == option.TargetTypeOCILayout && opts.DistributionSpec.ReferrersAPI != nil && *opts.DistributionSpec.ReferrersAPI == true {
+				return errors.New("cannot enforce referrers API for image layout target")
+			}
+			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts.targetRef = args[0]
-			opts.FileRefs = args[1:]
 			return runAttach(opts)
 		},
 	}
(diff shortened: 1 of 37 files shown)


FeynmanZhou:
The examples in CLI help look good to me.

FeynmanZhou (commented):
The examples in CLI help look good to me!

cmd/oras/attach.go
```diff
@@ -72,14 +73,23 @@ Example - Attach file 'hi.txt' and add manifest annotations:
...
 		Args: cobra.MinimumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return option.Parse(&opts)
+			opts.RawReference = args[0]
+			opts.FileRefs = args[1:]
+			if err := option.Parse(&opts); err != nil {
+				return err
+			}
+			if opts.Type == option.TargetTypeOCILayout && opts.DistributionSpec.ReferrersAPI != nil && *opts.DistributionSpec.ReferrersAPI {
+				return errors.New("cannot enforce referrers API for image layout target")
```
Wwwsylvia, line 88: What does it mean if a user uses OCI layout and sets `opts.DistributionSpec.ReferrersAPI` to false? Does it mean they want to use tag schema?

cmd/oras/push.go
```diff
@@ -262,7 +261,9 @@ func pushArtifact(dst *remote.Repository, pack packFunc, packOpts *oras.PackOpti
 	if err := display.PrintStatus(root, "Fallback ", verbose); err != nil {
 		return ocispec.Descriptor{}, err
 	}
-	dst.SetReferrersCapability(false)
+	if repo, ok := dst.(*remote.Repository); ok {
+		repo.SetReferrersCapability(false)
```
Wwwsylvia, line 265: I may have missed something... Can you help me to understand why do we set the referrers capability here?

cmd/oras/attach.go
```diff
@@ -19,24 +19,25 @@ import (
...
 	oerrors "oras.land/oras/cmd/oras/internal/errors"
 	"oras.land/oras/cmd/oras/internal/option"
 )
 
 type attachOptions struct {
 	option.Common
-	option.Remote
 	option.Packer
 	option.ImageSpec
 	option.DistributionSpec
```
shizhMSFT, line 38: I just noticed that the `DistributionSpec` should be stick to the `option.Remote`. This should also be applied to `oras cp` and `oras manifest push`.

cmd/oras/attach.go
qweeah: Referrers are supported via tag scheme in OCI layout, right? Here we safely ignored the ONLY if the `DistributionSpec.ReferrersAPI` is set to false

cmd/oras/push.go
qweeah: Since it's fallbacking, we assumes **Referrers API is not supported** based on **artifact media type is not supported**. Let me add doc in code.

cmd/oras/attach.go
```diff
@@ -156,9 +166,18 @@ func runAttach(opts attachOptions) error {
...
+	digest := subject.Digest.String()
+	if !strings.HasSuffix(opts.RawReference, digest) {
+		// Reassemble a reference with subject digest
+		if repo, ok := dst.(*remote.Repository); ok {
+			ref := repo.Reference
+			ref.Reference = subject.Digest.String()
+			opts.RawReference = ref.String()
+		} else if opts.Type == option.TargetTypeOCILayout {
+			opts.RawReference = fmt.Sprintf("%s@%s", opts.Path, subject.Digest)
+		}
```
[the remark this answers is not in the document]
qweeah: But `repo.Reference` (including Host name and repository name) need `*remote.Repository` type assertion anyway.

cmd/oras/attach.go
qweeah: No it should not. If we do that, `DistributionSpec` will apply flag to all commands related to `option.Remote`, whichi is wrong considering the following cases:
1. Since `oras push` is not related to referrers maintanance, the execution is always the same no matter what value is provided to `--distribution-spec`. So we should NEVER add this flag to those commands.
2. If user specify `--distribution-spec v1.1-referrers-tag` to `oras discover`, should the command looking for referrers via tag scheme? We should design the expected behaviour and then add this flag to those commands.

cmd/oras/push.go
```diff
@@ -95,22 +94,22 @@ Example - Push file "hi.txt" with multiple tags:
...
+			opts.FileRefs = args[1:]
 			if opts.artifactType != "" && opts.manifestConfigRef != "" {
 				return errors.New("--artifact-type and --config cannot both be provided")
 			}
 			return option.Parse(&opts)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if opts.ManifestMediaType == ocispec.MediaTypeArtifactManifest && opts.manifestConfigRef != "" {
-				return errors.New("cannot build an OCI artifact with manifest config")
-			}
```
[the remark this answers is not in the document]
qweeah: This is lost by mistake.

cmd/oras/attach.go
Wwwsylvia: No, the OCI store does not use tag schema for Referrers.

qweeah: > No, the OCI store does not use tag schema for Referrers.

If so, we should definitely return error.

shizhMSFT (approved):
LGTM

PR #752, opened 2023-01-17
feat(oci-layout): support in fetch and push for manifest and blob commands

qweeah:
Related to #378

### DIFF
--- a/cmd/oras/blob/fetch.go
+++ b/cmd/oras/blob/fetch.go
@@ -21,10 +21,12 @@ import (
 	"io"
 	"os"
 
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras/cmd/oras/internal/option"
 )
 
@@ -33,10 +35,9 @@ type fetchBlobOptions struct {
 	option.Common
 	option.Descriptor
 	option.Pretty
-	option.Remote
+	option.Target
 
 	outputPath string
-	targetRef  string
 }
 
 func fetchCmd() *cobra.Command {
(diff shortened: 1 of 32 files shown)


cmd/oras/blob/push.go
```diff
@@ -48,30 +47,33 @@ func pushCmd() *cobra.Command {
 
 ** This command is in preview and under development. **
 
-Example - Push blob "hi.txt":
+Example - Push blob 'hi.txt' into a registry:
```
Wwwsylvia, line 50: doc: "into" or "to"? 🤔

cmd/oras/manifest/fetch.go
```diff
@@ -87,17 +93,18 @@ Example - Fetch manifest with prettified json result:
...
-	repo, err := opts.NewRepository(opts.targetRef, opts.Common)
+	target, err := opts.NewReadonlyTarget(ctx, opts.Common)
 	if err != nil {
 		return err
 	}
-
-	if repo.Reference.Reference == "" {
-		return oerrors.NewErrInvalidReference(repo.Reference)
+	if opts.Reference == "" {
+		return oerrors.NewErrInvalidReferenceStr(opts.Reference)
```
Wwwsylvia, line 101: Just a question, what do the `opts.Reference` and `opts.RawReference` look like?

qweeah: `opts.RawReference` is the full user input, e.g. `localhost:5000/test@sha256:xxxx`. `opts.Reference` is the tag or digest part, e.g. `sha256:XXXX`. Here should use RawReference.

qweeah: Maybe I should move this error-checking function into target so it's less error-prone.

cmd/oras/blob/fetch.go
```diff
@@ -48,17 +49,23 @@ func fetchCmd() *cobra.Command {
...
   oras blob fetch --output - localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
 
 Example - Fetch and print the descriptor of a blob:
   oras blob fetch --descriptor localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
 
-Example - Fetch the blob, save it to a local file and print the descriptor:
+Example - Fetch a blob, save it to a local file and print the descriptor:
   oras blob fetch --output blob.tar.gz --descriptor localhost:5000/hello@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
+
+Example - Fetch and print a blob from OCI image layout folder 'layout-dir':
```
FeynmanZhou, line 64: Is it exact to use `layout-dir@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5` to represent an OCI layout folder?

cmd/oras/manifest/push.go
```diff
@@ -75,19 +74,22 @@ Example - Push a manifest to repository 'locahost:5000/hello' and tag with 'tag1
 
 Example - Push a manifest to repository 'locahost:5000/hello' and tag with 'tag1', 'tag2', 'tag3' and concurrency level tuned:
   oras manifest push --concurrency 6 localhost:5000/hello:tag1,tag2,tag3 manifest.json
+
+Example - Push a manifest to an OCI layout folder 'layout-dir' and with tag 'latest':
```
FeynmanZhou, line 78: Should we use `tag with 'latest'` to keep consistent with other examples?

FeynmanZhou (commented):
The examples look good to me. Only two simple questions left.

shizhMSFT (approved):
LGTM

cmd/oras/blob/fetch.go
qweeah: `layout-dir@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5` means fetch a blob with digest `sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5` from an OCI image layout folder 'layout-dir'

cmd/oras/manifest/push.go
qweeah: I will update all tags to 'latest' to make it consistent.

qweeah: Sorry I guess I misunderstand your meaning, will rephrase to `tag with`

Wwwsylvia (approved):
LGTM

FeynmanZhou (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #765, opened 2023-01-28
feat: support artifact discovering in OCI image layout

### LINKED ISSUE #760: support discovering referrers in OCI image layouts
Currently `oras discover` doesn't support OCI image layout target. Artifact users should be able to use it to view referrers of a subject in their OCI layout folders or tar archives.

qweeah:
This PR adds OCI image layout support for `oras discover` command and resolves #760. The UX is consistent with remote registry and looks like below:

```
# 0. Copy the test DAG to local OCI image layout folder
$ oras cp jinzha1.azurecr.io/oci-root:foo -r --to-oci-layout oci-root:foo

# 1. Discovering referrers
$ oras discover --oci-layout oci-root:foo
Discovered 2 artifacts referencing jinzha1.azurecr.io/oci-root:foo
Digest: sha256:d415a6dda5872dedf2e080fded1dc40276fbfb71ce054f70bb57e2431d6a8b23

Artifact Type   Digest
type.artifact   sha256:3d050f891f461ce021dcf5b1bc6586c4d6d7e3087867cf859ee613e8b304dfb3
type.image      sha256:cd13e0cb5a8d5f42c3fbcfe53171ad25fc430879a190dc71121ec7b7f7b83386

# 2. Discovering referrers and filtering
$ oras discover --oci-layout oci-root:foo  --artifact-type type.image
Discovered 1 artifact referencing foo
Digest: sha256:d415a6dda5872dedf2e080fded1dc40276fbfb71ce054f70bb57e2431d6a8b23

Artifact Type   Digest
type.image      sha256:cd13e0cb5a8d5f42c3fbcfe53171ad25fc430879a190dc71121ec7b7f7b83386
```

### DIFF
--- a/cmd/oras/discover.go
+++ b/cmd/oras/discover.go
@@ -18,7 +18,6 @@ package main
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -77,9 +76,6 @@ Example - Discover referrers of the manifest tagged 'v1' in an OCI layout folder
 `,
 		Args: cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			if opts.ReferrersAPI != nil && opts.Type == option.TargetTypeOCILayout {
-				return errors.New("cannot specify --distribution-spec flag on an image layout target")
-			}
 			opts.RawReference = args[0]
 			return option.Parse(&opts)
 		},
--- a/cmd/oras/discover.go
+++ b/cmd/oras/discover.go
@@ -175,7 +175,7 @@ func fetchReferrers(ctx context.Context, target oras.ReadOnlyGraphTarget, desc o
 			case ocispec.MediaTypeImageManifest:
 				var image ocispec.Manifest
 				json.Unmarshal(fetched, &image)
-				matchSubject(image.Subject, desc, &node, image.Config.ArtifactType)
+				matchSubject(image.Subject, desc, &node, image.Config.MediaType)
 			}
 			if node.ArtifactType != "" && (artifactType == "" || artifactType == node.ArtifactType) {
 				results = append(results, node)
--- a/cmd/oras/discover.go
+++ b/cmd/oras/discover.go
@@ -69,6 +69,10 @@ Example - Discover all the referrers of manifest with annotations, displayed in 
 
 Example - Discover referrers with type 'test-artifact' of manifest 'hello:v1' in registry 'localhost:5000':
   oras discover --artifact-type test-artifact localhost:5000/hello:v1
+
+Example - Discover referrers of the manifest tagged 'v1' in an OCI layout folder 'layout-dir':
+  oras discover layout-dir:v1
+  oras discover -v -o tree layout-dir:v1
 `,
 		Args: cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/oras/discover.go
+++ b/cmd/oras/discover.go
@@ -25,7 +25,7 @@ import (
 	"gopkg.in/yaml.v3"
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
-	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry"
 	"oras.land/oras/cmd/oras/internal/option"
 
 	"github.com/need-being/go-tree"
@@ -71,8 +71,8 @@ Example - Discover referrers with type 'test-artifact' of manifest 'hello:v1' in
   oras discover --artifact-type test-artifact localhost:5000/hello:v1
 
 Example - Discover referrers of the manifest tagged 'v1' in an OCI layout folder 'layout-dir':
-  oras discover layout-dir:v1
-  oras discover -v -o tree layout-dir:v1
+  oras discover --oci-layout layout-dir:v1
+  oras discover --oci-layout -v -o tree layout-dir:v1
 `,
 		Args: cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
@@ -141,7 +141,7 @@ func runDiscover(opts discoverOptions) error {
 
 func fetchReferrers(ctx context.Context, target oras.ReadOnlyGraphTarget, desc ocispec.Descriptor, artifactType string) ([]ocispec.Descriptor, error) {
 	var results []ocispec.Descriptor
-	if repo, ok := target.(*remote.Repository); ok {
+	if repo, ok := target.(registry.ReferrerLister); ok {
 		// get referrers directly
 		err := repo.Referrers(ctx, desc, artifactType, func(referrers []ocispec.Descriptor) error {
 			results = append(results, referrers...)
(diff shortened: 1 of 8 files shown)


cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
 
-func fetchReferrers(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor, artifactType string) ([]ocispec.Descriptor, error) {
-	results := []ocispec.Descriptor{}
-	err := repo.Referrers(ctx, desc, artifactType, func(referrers []ocispec.Descriptor) error {
-		results = append(results, referrers...)
-		return nil
-	})
-	if err != nil {
-		return nil, err
+func fetchReferrers(ctx context.Context, target oras.ReadOnlyGraphTarget, desc ocispec.Descriptor, artifactType string) ([]ocispec.Descriptor, error) {
```
shizhMSFT, line 142: Can we put this function into a package so that we can have unit tests?

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+	var results []ocispec.Descriptor
+	if repo, ok := target.(*remote.Repository); ok {
```
shizhMSFT, line 144: You may consider asserting [registry.ReferrerLister](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0/registry#ReferrerLister).

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+		// find matched referrers in all predecessors
+		predecessors, err := target.Predecessors(ctx, desc)
+		if err != nil {
+			return nil, err
+		}
+		for _, node := range predecessors {
+			var fetched []byte
+			if rc, err := target.Fetch(ctx, node); err != nil {
+				return nil, err
+			} else {
```
wangxiaoxuan273, line 163: The else enclosing can be deleted.

qweeah: If else is deleted, then `rc` will be not reachable.

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+				if isMatch(image.Subject, desc) {
+					node.ArtifactType = image.Config.MediaType
+				}
+			}
+			if node.ArtifactType != "" && (artifactType == "" || artifactType == node.ArtifactType) {
+				results = append(results, node)
+			}
+		}
 	}
 	return results, nil
```
shizhMSFT, line 191: nit: code style: it would be more readable and less indents to have
```go
var result T
if ... {
    ...
    return result
}

...
return result
```
instead of
```go
if ... {
    ...
} else {
    ...
}

return result
```

cmd/oras/discover.go
```diff
@@ -70,13 +69,17 @@ Example - Discover all the referrers of manifest with annotations, displayed in
 
 Example - Discover referrers with type 'test-artifact' of manifest 'hello:v1' in registry 'localhost:5000':
   oras discover --artifact-type test-artifact localhost:5000/hello:v1
+
+Example - Discover referrers of the manifest tagged 'v1' in an OCI layout folder 'layout-dir':
+  oras discover layout-dir:v1
+  oras discover -v -o tree layout-dir:v1
```
shizhMSFT, line 75: Missing `--oci-layout`?

cmd/oras/discover.go
qweeah: It reminds me that I can actually use https://pkg.go.dev/oras.land/oras-go/v2#FetchBytes to avoid reading from `rc` myself

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+		err := repo.Referrers(ctx, desc, artifactType, func(referrers []ocispec.Descriptor) error {
+			results = append(results, referrers...)
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// find matched referrers in all predecessors
+		predecessors, err := target.Predecessors(ctx, desc)
```
shizhMSFT, line 155: We need to check whether `desc` is a manifest or not. Proceed only if `desc` describes a manifest.

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+		if err != nil {
+			return nil, err
+		}
+		for _, node := range predecessors {
+			var fetched []byte
+			if rc, err := target.Fetch(ctx, node); err != nil {
```
shizhMSFT, line 161: We may check the node media type before fetching it. It saves disk I/Os.

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+				return nil, err
+			} else {
+				fetched, err = content.ReadAll(rc, node)
+				if err != nil {
+					return nil, err
+				}
+			}
```
shizhMSFT, line 168: Instead of using `target.Fetch`, try [oras.FetchBytes](https://pkg.go.dev/oras.land/oras-go/v2#FetchBytes). Or make your own `FetchBytes` function, which takes a descriptor instead of a reference string.

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+			var isMatch = func(got *ocispec.Descriptor, want ocispec.Descriptor) bool {
+				return got != nil && content.Equal(*got, want)
+			}
```
shizhMSFT, line 171: It seems redundant.

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+			switch node.MediaType {
+			case ocispec.MediaTypeArtifactManifest:
+				var artifact ocispec.Artifact
+				json.Unmarshal(fetched, &artifact)
```
shizhMSFT, line 175: error is not checked

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+				if isMatch(artifact.Subject, desc) {
+					node.ArtifactType = artifact.ArtifactType
+				}
+			case ocispec.MediaTypeImageManifest:
+				var image ocispec.Manifest
+				json.Unmarshal(fetched, &image)
```
shizhMSFT, line 181: error is not checked.

cmd/oras/discover.go
```diff
@@ -136,19 +139,59 @@ func runDiscover(opts discoverOptions) error {
...
+				if isMatch(image.Subject, desc) {
+					node.ArtifactType = image.Config.MediaType
+				}
```
shizhMSFT, line 184: Annotations are not extracted as required by Referrers API.

cmd/oras/discover.go
shizhMSFT: `rc` is not deferred to be closed.

cmd/oras/discover.go
qweeah: Will use oras.FetchBytes. Since oci target reference resolving is done in memory, it should incur minor impact on the performance.

internal/graph/graph.go
```diff
@@ -59,3 +62,72 @@ func Successors(ctx context.Context, fetcher content.Fetcher, node ocispec.Descr
...
+			results = append(results, referrers...)
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+		return results, nil
+	}
+
+	if !descriptor.IsImageManifest(desc) && desc.MediaType != ocispec.MediaTypeArtifactManifest {
```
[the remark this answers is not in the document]
shizhMSFT: Let’s remove this check and relax a bit as we can always add later.

shizhMSFT (approved):
LGTM

PR #768, opened 2023-01-28
feat: suppport repo ls sub-namespace

### LINKED ISSUE #733: oras repo ls to support sub namespaces

My registry contains multiple repos under nested namespaces
``` console
base-images/dotnet/core/sdk
base-images/node
hello-world
samples/helm/wordpress
samples/text
samples/thingthang
samples/txt
```

I'd like to find the repos under the samples/ namespace

calling `oras repo ls demo42.azurecr.io` gives an appropriate listing
However, calling `oras repo ls demo42.azurecr.io/samples` gives:
`Error: invalid reference: invalid registry`

list the repos under the `/samples` namespace. This would also be the case for something like docker that lists repos under an org namespace, although docker hub doesn't support the `_catalog` API :(

### How can we reproduce it?

```bash
oras push $REGISTRY/samples/artifact:1.0 \
    --config /dev/null:readme/example\
    ./readme.md:application/markdown

oras repo ls $REGISTRY/samples
```

0.16.0

wsl2

wangxiaoxuan273:
Resolves #733 

Test output on terminal:
image

### DIFF
--- a/cmd/oras/repository/ls.go
+++ b/cmd/oras/repository/ls.go
@@ -17,16 +17,19 @@ package repository
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
+	"oras.land/oras-go/v2/registry"
 	"oras.land/oras/cmd/oras/internal/option"
 )
 
 type repositoryOptions struct {
 	option.Remote
 	option.Common
-	hostname string
-	last     string
+	hostname  string
+	namespace string
+	last      string
 }
 
 func listCmd() *cobra.Command {
@@ -41,6 +44,9 @@ func listCmd() *cobra.Command {
 Example - List the repositories under the registry:
   oras repo ls localhost:5000
 
+Example - List the repositories under a namespace in the registry:
+  oras repo ls localhost:5000/example-namespace
+
 Example - List the repositories under the registry that include values lexically after last:
   oras repo ls --last "last_repo" localhost:5000
 `,
@@ -50,7 +56,9 @@ Example - List the repositories under the registry that include values lexically
 			return option.Parse(&opts)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts.hostname = args[0]
+			if err := parseRepoPath(&opts, args[0]); err != nil {
+				return fmt.Errorf("could not parse repository path: %w", err)
+			}
 			return listRepository(opts)
 		},
 	}
@@ -60,6 +68,21 @@ Example - List the repositories under the registry that include values lexically
 	return cmd
 }
 
+func parseRepoPath(opts *repositoryOptions, arg string) error {
+	path := strings.TrimSuffix(arg, "/")
+	if strings.Contains(path, "/") {
+		reference, err := registry.ParseReference(path)
+		if err != nil || reference.Reference != "" {
+			return err
+		}
+		opts.hostname = reference.Registry
+		opts.namespace = reference.Repository
+	} else {
+		opts.hostname = path
+	}
+	return nil
+}
+
 func listRepository(opts repositoryOptions) error {
 	ctx, _ := opts.SetLoggerLevel()
 	reg, err := opts.Remote.NewRegistry(opts.hostname, opts.Common)
@@ -67,8 +90,12 @@ func listRepository(opts repositoryOptions) error {
 		return err
 	}
 	return reg.Repositories(ctx, opts.last, func(repos []string) error {
-		for _, repo := range repos {
-			fmt.Println(repo)
+		for _, repopath := range repos {
+			if opts.namespace == "" {
+				fmt.Println(repopath)
+			} else if strings.HasPrefix(repopath, opts.namespace+"/") {
+				fmt.Println(repopath[len(opts.namespace)+1:])
+			}
 		}
 		return nil
 	})
--- a/cmd/oras/repository/ls.go
+++ b/cmd/oras/repository/ls.go
@@ -72,11 +72,14 @@ func parseRepoPath(opts *repositoryOptions, arg string) error {
 	path := strings.TrimSuffix(arg, "/")
 	if strings.Contains(path, "/") {
 		reference, err := registry.ParseReference(path)
-		if err != nil || reference.Reference != "" {
+		if err != nil {
 			return err
 		}
+		if reference.Reference != "" {
+			return fmt.Errorf("tags or digests should not be provided")
+		}
 		opts.hostname = reference.Registry
-		opts.namespace = reference.Repository
+		opts.namespace = reference.Repository + "/"
 	} else {
 		opts.hostname = path
 	}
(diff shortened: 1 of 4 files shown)


wangxiaoxuan273:
@qweeah Please take a look at the failed e2e CI test. It's likely that the failed test need to be changed for this feature.

wangxiaoxuan273:
I have two design questions:

1. When `engineering` is a repository and a namespace, should `example-registry/engineering` return the `engineering` repository?

i.e. :
```
$ oras repo ls example-registry
employeehandbook
engineering
engineering/bestpractices
engineering/newhires/onboarding
engineering/newhires/teamoverview
```

currently the `engineering` repository is not returned given `example-registry/engineering`.
```
$ oras repo ls example-registry/engineering
engineering/bestpractices
engineering/newhires/onboarding
engineering/newhires/teamoverview
```

2. Which display is better?

qweeah:
> @qweeah Please take a look at the failed e2e CI test. It's likely that the failed test need to be changed for this feature.

Yes you are right. One [assertion ](https://github.com/oras-project/oras/blob/main/test/e2e/suite/command/repo.go#L37) can be removed since `oras repo ls $REG/$REPO_SUFFIX` is now a valid input.

qweeah:
> When engineering is a repository and a namespace, should example-registry/engineering return the engineering repository?

- If a registry need to explicitly define a separate namespace concept from repository(e.g. ghcr, dockerhub), I don't think `engineering` can be a repository and a namespace at the same time.
- If repository is only a slash separated path for a regsitry (e.g. ACR, MCR), `engineering` SHOULD be included in the result of `oras repo ls example.registry/engineering`. There is no need to hide the result of a full match.

qweeah:
> Which display is better?

I think the latter is out-of-option. Consider the scenario: How can the result of `oras repo ls some.registry/my/repo` be displayed if `my/repo` is a repository?

cmd/oras/repository/ls.go
```diff
@@ -50,7 +55,13 @@ Example - List the repositories under the registry that include values lexically
...
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts.hostname = args[0]
+			hostname, namespace, _ := strings.Cut(args[0], "/")
+			opts.hostname = hostname
+			opts.namespace = namespace
+			// namespace is invalid if a tag is provided
+			if strings.Contains(namespace, ":") {
+				return fmt.Errorf("invalid namespace")
+			}
```
shizhMSFT, line 64: Can we refactor this part into a function so that we can unit test it?

cmd/oras/repository/ls.go
```diff
@@ -25,8 +26,9 @@ import (
 type repositoryOptions struct {
 	option.Remote
 	option.Common
-	hostname string
-	last     string
+	hostname  string
+	namespace string
```
shizhMSFT, line 30: A reference can have many [namespaces](https://learn.microsoft.com/azure/container-registry/container-registry-best-practices#repository-namespaces).

cmd/oras/repository/ls.go
shizhMSFT: Parsing can be simplified in an easier way.

If `args[0]` contains `/`, we can parse it using [ParseReference()](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0/registry#ParseReference). Otherwise, it is a hostname.

cmd/oras/repository/ls.go
shizhMSFT: How is `namespace` typically stored?

cmd/oras/repository/ls.go
```diff
@@ -67,9 +78,22 @@ func listRepository(opts repositoryOptions) error {
...
+
+func inNamespace(repopath string, requestedNamespace string) bool {
+	if requestedNamespace == "" {
+		return true
+	}
+	before, after, found := strings.Cut(repopath, requestedNamespace)
+	if !found || after == "" || before != "" {
+		return false
+	}
+	return after[0] == '/'
```
shizhMSFT, line 98: It partially works and is not readable. Can we choose a simpler implementation?

The reason why it partially works is that the `inNamespace("foo", "foo")` returns `false`, which should be `true`. You need unit tests to ensure code quality.

shizhMSFT: It seems to be a UX issue other than a bug.

shizhMSFT:
> I have two design questions:
> 
> * When `engineering` is a repository and a namespace, should `example-registry/engineering` return the `engineering` repository?
> 
> i.e. :
> 
> ```
> $ oras repo ls example-registry
> employeehandbook
> engineering
> engineering/bestpractices
> engineering/newhires/onboarding
> engineering/newhires/teamoverview
> ```
> 
> currently the `engineering` repository is not returned given `example-registry/engineering`.
> 
> ```
> $ oras repo ls example-registry/engineering
> engineering/bestpractices
> engineering/newhires/onboarding
> engineering/newhires/teamoverview
> ```
> 
> * Which display is better?
> 
> currently:
> 
> ```
> $ oras repo ls example-registry/engineering
> engineering/bestpractices
> engineering/newhires/onboarding
> engineering/newhires/teamoverview
> ```
> 
> or
> 
> ```
> $ oras repo ls example-registry/engineering
> bestpractices
> newhires/onboarding
> newhires/teamoverview
> ```

This is a good question. I would suggest the latter one:
```console
$ oras repo ls example-registry/engineering
bestpractices
newhires/onboarding
newhires/teamoverview
```
In this way, it is more consistent with the UNIX `ls` command. In addition, it is more script-friendly that you can do
```sh
namespace=example-registry/engineering
oras repo ls $namespace | while read repo; do
    repo=$namespace/$repo
    echo Tags in $repo:
    oras repo tags $repo
    echo
done
```

wangxiaoxuan273:
Need to handle the case when user inputs "foo/bar/extra/", the last slash should be trimmed off.

wangxiaoxuan273:
> > Which display is better?
> 
> I think the latter is out-of-option. Consider the scenario: How can the result of `oras repo ls some.registry/my/repo` be displayed if `my/repo` is a repository?

Then `oras repo ls some.registry/my/repo` does not return 'my/repo', if we choose the latter one. 
Update: we've decided to use the latter display.

cmd/oras/repository/ls.go
wangxiaoxuan273: Used `ParseReference`. Need to discuss which paths are invalid.

cmd/oras/repository/ls.go
wangxiaoxuan273: I need advice on a readable and simpler implementation.

wangxiaoxuan273: Used `strings.HasPrefix(repopath, opts.namespace+"/") `here after discussion.

cmd/oras/repository/ls.go
```diff
@@ -60,15 +68,34 @@ Example - List the repositories under the registry that include values lexically
...
 
+func parseRepoPath(opts *repositoryOptions, arg string) error {
+	path := strings.TrimSuffix(arg, "/")
+	if strings.Contains(path, "/") {
+		reference, err := registry.ParseReference(path)
+		if err != nil || reference.Reference != "" {
+			return err
+		}
+		opts.hostname = reference.Registry
+		opts.namespace = reference.Repository
```
shizhMSFT, line 79:
```suggestion
		opts.namespace = reference.Repository + "/"
```
so that we don't need to calculate it multiple time in a loop at line 96.

wangxiaoxuan273: changed accordingly.

cmd/oras/repository/ls.go
```diff
@@ -60,15 +68,34 @@ Example - List the repositories under the registry that include values lexically
 	return cmd
 }
 
+func parseRepoPath(opts *repositoryOptions, arg string) error {
```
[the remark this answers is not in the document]
wangxiaoxuan273: Added unit tests.

shizhMSFT (approved):
LGTM

Issue #770, opened 2023-01-30
The flag `--to-distribution-spec` does not work with `oras copy`


Wwwsylvia:
### What happened in your environment?

When using the flag `--to-distribution-spec v1.1-referrers-tag` with `oras copy`, I expect a referrers tag to be created in my target registry. However, the flag does not take effect and the referrers tag is not created.

### What did you expect to happen?

Referrers tag schema is enforced, and a referrers tag is created in the target registry.

### How can we reproduce it?

`oras copy -r $srcRegistry $dstRegistry --to-distribution-spec v1.1-referrers-tag`

### What is the version of your ORAS CLI?

Version:        0.16.0+unreleased
Go version:     go1.19.3
Git commit:     6e876c1e785e81576f5f9c28fac66a60986776e3

### What is your OS environment?

Windows

### Are you willing to submit PRs to fix it?

- [X] Yes, I am willing to fix it.

Issue #771, opened 2023-01-30
bug: The flag `--distribution-spec` is not recognized by `oras attach`


Wwwsylvia:
### What happened in your environment?

The flag `--distribution-spec` is not recognized by `oras attach` while there are examples about the flag `--distribution-spec` in the help text.

Error message: 
```shell
Error: unknown flag: --distribution-spec
```
![image](https://user-images.githubusercontent.com/8113721/215431356-c2b88bf2-eade-4643-b9e9-f56962184a7e.png)
![image](https://user-images.githubusercontent.com/8113721/215431064-85e7e4a7-305b-4187-8784-bbb720c50c01.png)

### What did you expect to happen?

The flag should work and should show up in the flag list.

### How can we reproduce it?

```shell
oras attach --artifact-type $artifactType $dstRef $fileToAttach --distribution-spec v1.1-referrers-tag
```

### What is the version of your ORAS CLI?

Version: 0.16.0+unreleased
Go version: go1.19.3
Git commit: https://github.com/oras-project/oras/commit/6e876c1e785e81576f5f9c28fac66a60986776e3

### What is your OS environment?

Windows

### Are you willing to submit PRs to fix it?

- [X] Yes, I am willing to fix it.

shizhMSFT:
Fixed by #772

PR #772, opened 2023-01-30
fix: fix bugs related to the spec `--distribution-spec`

Wwwsylvia:
1. Fix flags `--from-distribution-spec` and `--to-distribution-spec` for `oras copy`
2. Fix the flag `--distribution-spec` for `oras attach`

This PR removes the check when specifying `--distribution-spec` for OCI layout.

Fixes: #770, #771

### DIFF
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -17,7 +17,6 @@ package option
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


shizhMSFT (approved):
LGTM

cmd/oras/internal/option/target.go
```diff
@@ -84,13 +83,11 @@ func (opts *Target) Parse() error {
 	switch {
 	case opts.isOCILayout:
 		opts.Type = TargetTypeOCILayout
-		if opts.Remote.distributionSpec.referrersAPI != nil {
```
qweeah, line 87: Does this check cause any bug?

PR #785, opened 2023-02-03
fix: Have make file fail with wrong go version

TerryHowe:
Latest command `oras repository ls` requires go 1.20

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ CLI_PKG     = $(PROJECT_PKG)/cmd/oras
 GIT_COMMIT  = $(shell git rev-parse HEAD)
 GIT_TAG     = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
 GIT_DIRTY   = $(shell test -n "`git status --porcelain`" && echo "dirty" || echo "clean")
+GO_EXE      = go
 
 TARGET_OBJS ?= checksums.txt darwin_amd64.tar.gz darwin_arm64.tar.gz linux_amd64.tar.gz linux_arm64.tar.gz linux_armv7.tar.gz linux_s390x.tar.gz windows_amd64.zip
 
@@ -31,8 +32,8 @@ LDFLAGS += -X $(PROJECT_PKG)/internal/version.GitCommit=${GIT_COMMIT}
 LDFLAGS += -X $(PROJECT_PKG)/internal/version.GitTreeState=${GIT_DIRTY}
 
 .PHONY: test
-test: vendor check-encoding
-	go test -race -v -coverprofile=coverage.txt -covermode=atomic ./...
+test: tidy vendor check-encoding
+	$(GO_EXE) test -race -v -coverprofile=coverage.txt -covermode=atomic ./...
 
 .PHONY: covhtml
 covhtml:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,6 +17,7 @@ CLI_PKG     = $(PROJECT_PKG)/cmd/oras
 GIT_COMMIT  = $(shell git rev-parse HEAD)
 GIT_TAG     = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
 GIT_DIRTY   = $(shell test -n "`git status --porcelain`" && echo "dirty" || echo "clean")
+GO=go1.20
```
shizhMSFT, line 20: `go1.20` is a bit confusing that most developers do not have commands like `go1.20` or `go1.19` in their `$PATH`. By following the golang official installation [instruction](https://go.dev/doc/install), most developers end up with just the `go` command.

shizhMSFT:
Since we have set the go module to have min go version of `1.20` as follows

https://github.com/oras-project/oras/blob/40d9a15897053a9477c95bb15083d3c2db36207d/go.mod#L3

running `make build` or `go build` with `go 1.19` will result in

```
note: module requires Go 1.20
```

which is expected.

TerryHowe:
The problem is `note: module requires Go 1.20` is printed, but the build doesn't fail except for the problem area. Calling `go mod tidy` has the build fail if the version is too old.

Makefile
```diff
@@ -31,8 +32,8 @@ LDFLAGS += -X $(PROJECT_PKG)/internal/version.GitCommit=${GIT_COMMIT}
 LDFLAGS += -X $(PROJECT_PKG)/internal/version.GitTreeState=${GIT_DIRTY}
 
 .PHONY: test
-test: vendor check-encoding
-	go test -race -v -coverprofile=coverage.txt -covermode=atomic ./...
+test: tidy vendor check-encoding
```
shizhMSFT, line 35: There is a problem. If someone opens a PR and messes the `go.mod` and `go.sum` files, calling `go mod tidy` in `make test` may not detect this kind of issue in the PR check.

Makefile
```diff
@@ -17,6 +17,7 @@ CLI_PKG     = $(PROJECT_PKG)/cmd/oras
 GIT_COMMIT  = $(shell git rev-parse HEAD)
 GIT_TAG     = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
 GIT_DIRTY   = $(shell test -n "`git status --porcelain`" && echo "dirty" || echo "clean")
+GO=go
```
shizhMSFT, line 20: I am assuming that you are managing Go installations according to https://go.dev/doc/manage-install.

I think it would be better to make the variable consistent:
```suggestion
GO_EXE     = go
```

Makefile
TerryHowe: Done

TerryHowe: Done

Makefile
TerryHowe: Sure thing.

I'm running my old go builds for various reasons.

TerryHowe: Done.

Makefile
TerryHowe: I'm not entirely sure what you mean by messes. Changes?

shizhMSFT: Yes, but it seems no harm for now.

shizhMSFT (approved):
LGTM

PR #788, opened 2023-02-06
feat: Add `oras download` command (WIP)

TerryHowe:
This is pretty rough still, but wanted to get it out for conversation.

Sample Usage:
```
oras download     --ca-file "${REGISTRY_CAFILE}" ${REGISTRY}/curated-packages/prometheus/prometheus@sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b  ${REGISTRY}/curated-packages/harbor/harbor-registry@sha256:7a4957ddd9228b312125ceb186af9d7e00fe1b8257191fd69cf4df355d1b4cd4
```

Sample output:
```
out/curated-packages/prometheus
out/curated-packages/prometheus/prometheus
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/ingest
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/oci-layout
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/blobs
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/blobs/sha256
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/blobs/sha256/9bfcd754305e2db4e9a78e80364d1b2834574c4c2563737f69a26b68dd47e64a
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/blobs/sha256/c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b
out/curated-packages/prometheus/prometheus/sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b/blobs/sha256/7a378ca94c3456ca82d63bd9c0db672219e968048bd70cdf1c665a05439323e5
[...]

PR #793, opened 2023-02-08
feat: Add `oras upload` command (WIP)

TerryHowe:
This is just a POC for download/upload commands for discussion. Related to https://github.com/oras-project/oras/pull/788

### DIFF
--- a/cmd/oras/upload.go
+++ b/cmd/oras/upload.go
@@ -1,0 +1,156 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/spf13/cobra"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content/oci"
+	"oras.land/oras-go/v2/registry"
+	"oras.land/oras/cmd/oras/internal/display"
+	"oras.land/oras/cmd/oras/internal/option"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+const (
+	directory = "output"
+)
+
+type uploadOptions struct {
+	option.Common
+	To option.Remote
+}
+
+func uploadCmd() *cobra.Command {
+	var opts uploadOptions
+	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/oras/upload.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func runupload(opts uploadOptions, destination string) error {
+	ctx, _ := opts.SetLoggerLevel()
+
+	var uniqueImages []registry.Reference
+	err := filepath.Walk(directory,
```
qweeah, line 73: This looks VERY interesting! The long description is not updated, but looks like this command is trying to upload all the image layout subfolders to a remote registry, using relative path as the repository names?

cmd/oras/upload.go
```diff
@@ -0,0 +1,156 @@
...
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			ociPath, layout := strings.CutSuffix(path, "/oci-layout")
+			if layout == true {
+				ociPath, _ = strings.CutPrefix(ociPath, directory+"/")
+				lastIdx := strings.LastIndex(ociPath, "/")
+				reference := registry.Reference{
```
qweeah, line 82: use [Reference](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0/registry#ParseReference)(ociPath) instead

cmd/oras/upload.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+	}
+	option.ApplyFlags(&opts, cmd.Flags())
+	return cmd
+}
+
+func runupload(opts uploadOptions, destination string) error {
+	ctx, _ := opts.SetLoggerLevel()
+
+	var uniqueImages []registry.Reference
```
qweeah, line 72: this is an array of repositories?
```suggestion
	var repos []registry. Reference
```

cmd/oras/upload.go
```diff
@@ -0,0 +1,156 @@
...
+		to := option.Target{
+			Path:         path.Join(destination, fileReference.Repository),
+			Reference:    fileReference.Reference,
+			RawReference: fileReference.String(),
+			Remote:       opts.To,
+			Type:         option.TargetTypeRemote,
+		}
+
+		to.Reference = ""
+		dst, err := to.NewTarget(opts.Common)
```
qweeah, line 138: Since the target cannot be a oci layout, you can use opts.To.NewRepository directly

cmd/oras/upload.go
```diff
@@ -0,0 +1,156 @@
...
+
+	var uniqueImages []registry.Reference
+	err := filepath.Walk(directory,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			ociPath, layout := strings.CutSuffix(path, "/oci-layout")
+			if layout == true {
+				ociPath, _ = strings.CutPrefix(ociPath, directory+"/")
```
qweeah, line 80: Looks like you need to 1) create an OCI read-only store 2) get all manifests from the store 3) add manifest reference to `uniqueImages`

cmd/oras/upload.go
TerryHowe: Yes.

PR #794, opened 2023-02-09
feat: enable --header flag to specify headers with requests

### LINKED ISSUE #521: Support specifying http headers via `--header` flag
As an ORAS user, I should be able to specify an array of http headers, which will be added to the outgoing http requests.

wangxiaoxuan273:
Resolves #521 

Designed in accordance to the **behavior**(as exact spec can't be found) of `curl`-H flag. Can be used multiple times.

Terminal output:
Screenshot 2023-02-10 at 00 22 43

Some minor issues/questions:
1. when parsing input like "c: d, e, f" into `map[string][]string` of golang, the spaces before and after the commas are not stripped. i.e. the values are `"d",` `" e"`, `" f"`.
(update: splitting is removed, the argument is kept as what is after the colon.)
3. curl parses input `-H "a:b" -H "a:b"` to two repeated headers "a:b", and `-H "a:b" -H "a:c"` to headers "a:b", "a:c". In this pr these behaviors are maintained.


TerryHowe (commented):
generally looks good to me, couple comments

cmd/oras/internal/option/remote.go
```diff
@@ -243,6 +247,24 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
 	return
 }
 
+func (opts *Remote) parseCustomHeaders() http.Header {
+	if len(opts.headers) != 0 {
+		headers := map[string][]string{}
+		for _, h := range opts.headers {
+			before, after, found := strings.Cut(h, ":")
```
qweeah, line 254: UX: should return an error if a flag is not formatted as `<name>:<comma-separated-values>`, so users can be aware of the potential errors in their input.
cc @FeynmanZhou

qweeah, line 254: how about 
```suggestion
			name, csv, found := strings.Cut(h, ":")
```

qweeah (changes requested):
Made some comments on implementation and UX. Thanks again for contributing!

qweeah:
> when parsing input like "c: d, e, f" into map[string][]string of golang, the spaces before and after the commas are not stripped. i.e. the values are "d", " e", " f".
 
We can just forward the value with trailing space. By design it should be ignored. I didn't find the RFC for this but there are some M[DN docs](https://developer.mozilla.org/en-US/docs/web/http/headers#:~:text=An%20HTTP%20header%20consists%20of%20its%20case%2Dinsensitive%20name%20followed%20by%20a%20colon%20(%3A)%2C%20then%20by%20its%20value.%20Whitespace%20before%20the%20value%20is%20ignored.) mentioning it.

> curl parses input -H "a:b" -H "a:b" to two repeated headers "a:b", and -H "a:b" -H "a:c" to headers "a:b", "a:c". In this pr these behaviors are maintained.
Shouldn't `-H "a:b" -H "a:c"` be parsed to `a:b,c`?

cmd/oras/internal/option/remote.go
```diff
@@ -243,6 +247,24 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
...
+			if found && after != "" {
+				if headers[before] == nil {
+					headers[before] = strings.Split(after, ",")
```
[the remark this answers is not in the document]
wangxiaoxuan273: Did you mean the argument of the `-H` command? I designed it according to the behavior of `curl` -H flag.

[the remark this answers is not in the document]
TerryHowe: No, I was talking about the way you can provide multiple header values with the comma separation.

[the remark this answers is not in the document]
qweeah: @TerryHowe Are you suggesting we shouldn't unbox the input value and just forward what is provided?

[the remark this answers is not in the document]
TerryHowe: I don't have a strong opinion about it, but it seemed a bit odd. I don't think curl uses that comma separated format. I think you just have to specify the same header with multiple `-H` flags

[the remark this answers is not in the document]
qweeah: curl don't use comma to separate header because 
-  value parsing can be done on server side. This also implies that oras don't need to split any header value
-  the value name can be duplicated. This is different here, because in golang, the underlying data struct [http.Header](https://pkg.go.dev/net/http#Header) is a map, and thus header key cannot be duplicated. So we might need to concatenate duplicated header with comma explicitly.

cmd/oras/internal/option/remote.go
```diff
@@ -62,6 +64,7 @@ func (opts *Remote) EnableDistributionSpecFlag() {
 func (opts *Remote) ApplyFlags(fs *pflag.FlagSet) {
 	opts.ApplyFlagsWithPrefix(fs, "", "")
 	fs.BoolVarP(&opts.PasswordFromStdin, "password-stdin", "", false, "read password or identity token from stdin")
+	fs.StringArrayVarP(&opts.headers, "header", "H", []string{}, "use custom headers with requests")
```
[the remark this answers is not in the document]
wangxiaoxuan273: changed accordingly.

cmd/oras/internal/option/remote.go
```diff
@@ -209,7 +212,8 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
 				TLSClientConfig:       config,
 			},
 		},
-		Cache: auth.NewCache(),
+		Cache:  auth.NewCache(),
+		Header: opts.parseCustomHeaders(),
```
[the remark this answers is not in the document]
wangxiaoxuan273: changed accordingly, after off-line discussion.

cmd/oras/internal/option/remote.go
wangxiaoxuan273: Returned error currently, need further discussion and confirmation.

wangxiaoxuan273: changed accordingly.

cmd/oras/internal/option/remote.go
```diff
@@ -243,6 +251,21 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
...
 
+func (opts *Remote) parseCustomHeaders() error {
+	if len(opts.headerFlags) != 0 {
+		headers := map[string][]string{}
+		for _, h := range opts.headerFlags {
+			name, csv, found := strings.Cut(h, ":")
+			if !found || strings.TrimSpace(csv) == "" {
+				return fmt.Errorf("cannot parse headers: %q", h)
+			}
+			headers[name] = append(headers[name], strings.Split(csv, ",")...)
```
qweeah, line 262: No need to unbox the value, just add it to the http headers.
```suggestion
			name, value, found := strings.Cut(h, ":")
			if !found || name == "" || strings.TrimSpace(value) == "" {
				return fmt.Errorf("invalid header: %q", h)
			}
			headers[name] = append(headers[name], value)
```

shizhMSFT, line 262: We should consume the value as it is without splitting.

You can try `-H "foo: a" -H "foo: b,c"` with `curl`. It will generate
```
> foo: a
> foo: b,c
```

wangxiaoxuan273: key can be empty string in curl.

wangxiaoxuan273: changed according to Billy's comment above.

cmd/oras/internal/option/remote.go
```diff
@@ -51,6 +51,9 @@ type Remote struct {
 	resolveDialContext    func(dialer *net.Dialer) func(context.Context, string, string) (net.Conn, error)
 	applyDistributionSpec bool
 	distributionSpec      distributionSpec
+
```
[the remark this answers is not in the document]
wangxiaoxuan273: removed.

cmd/oras/internal/option/remote.go
```diff
@@ -243,6 +250,21 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
 	return
 }
 
+func (opts *Remote) parseCustomHeaders() error {
+	if len(opts.headerFlags) != 0 {
+		headers := map[string][]string{}
+		for _, h := range opts.headerFlags {
+			name, value, found := strings.Cut(h, ":")
+			if !found || strings.TrimSpace(value) == "" {
+				return fmt.Errorf("cannot parse headers: %q", h)
```
shizhMSFT, line 259: nit:
```suggestion
				return fmt.Errorf("invalid header: %q", h)
```

cmd/oras/internal/option/remote.go
```diff
@@ -243,6 +250,21 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
 	return
 }
 
+func (opts *Remote) parseCustomHeaders() error {
+	if len(opts.headerFlags) != 0 {
+		headers := map[string][]string{}
+		for _, h := range opts.headerFlags {
+			name, value, found := strings.Cut(h, ":")
+			if !found || strings.TrimSpace(value) == "" {
```
shizhMSFT, line 258: According to [RFC 2616 Section 4.2](https://www.rfc-editor.org/rfc/rfc2616#section-4.2), the name must not be empty and the value can be empty.

shizhMSFT: Note:
```go
package main

import (
	"net/http"
	"os"
)

func main() {
	headers := make(http.Header)
	headers[""] = []string{"abc"}
	headers["    "] = []string{"efg"}
	headers["empty"] = []string{""}
	headers["space"] = []string{"   "}
	headers["foo"] = []string{"bar"}
	headers["foo2"] = []string{"  bar2"}

	headers.Write(os.Stdout)
}
```
outputs
```
empty: 
foo: bar
foo2: bar2
space: 
```

wangxiaoxuan273: > According to [RFC 2616 Section 4.2](https://www.rfc-editor.org/rfc/rfc2616#section-4.2), the name must not be empty and the value can be empty.

curl's behavior is the other way around. Which should I follow, curl or the RFC spec?

shizhMSFT: Always follow the spec since the spec is the standard.

cmd/oras/internal/option/remote.go
wangxiaoxuan273: changed accordingly.

cmd/oras/internal/option/remote.go
wangxiaoxuan273: Changed the behavior of parsing to be conformant to the spec.

cmd/oras/internal/option/remote.go
wangxiaoxuan273: removed the unboxing.

cmd/oras/internal/option/remote.go
```diff
@@ -243,6 +250,21 @@ func (opts *Remote) authClient(registry string, debug bool) (client *auth.Client
 	return
 }
 
+func (opts *Remote) parseCustomHeaders() error {
+	if len(opts.headerFlags) != 0 {
+		headers := map[string][]string{}
+		for _, h := range opts.headerFlags {
+			name, value, found := strings.Cut(h, ":")
+			if !found || strings.TrimSpace(name) == "" {
```
qweeah, line 258: It's better to add a link to the rfc doc

wangxiaoxuan273: Added.

cmd/oras/internal/option/remote.go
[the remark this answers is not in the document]
wangxiaoxuan273: Update: removed comma unboxing.

cmd/oras/internal/option/remote_test.go
```diff
@@ -285,3 +285,98 @@ func TestRemote_parseResolve_err(t *testing.T) {
...
+			wantErr:     false,
+		},
+		{
+			name:        "valid headers with spaces",
+			headerFlags: []string{"foo:   a", "bar   :b"},
+			want:        map[string][]string{"foo": {"   a"}, "bar   ": {"b"}},
+			wantErr:     false,
+		},
+		{
+			name:        "invalid headers with spaces",
```
qweeah, line 351: suggested to rename. Also should we split those two to make sure both errors are detected independently?
```suggestion
			name:        "pure-space name and empty name",
```

cmd/oras/internal/option/remote_test.go
```diff
@@ -285,3 +285,98 @@ func TestRemote_parseResolve_err(t *testing.T) {
...
+		wantErr     bool
+	}{
+		{
+			name:        "no custom header is provided",
+			headerFlags: []string{},
+			want:        nil,
+			wantErr:     false,
+		},
+		{
+			name:        "valid input, one header, one value",
```
qweeah, line 303:
```suggestion
			name:        "one name-value pair",
```

qweeah (changes requested):

qweeah (approved):
LGTM

cmd/oras/internal/option/remote.go
```diff
@@ -62,6 +64,7 @@ func (opts *Remote) EnableDistributionSpecFlag() {
 func (opts *Remote) ApplyFlags(fs *pflag.FlagSet) {
 	opts.ApplyFlagsWithPrefix(fs, "", "")
 	fs.BoolVarP(&opts.PasswordFromStdin, "password-stdin", "", false, "read password or identity token from stdin")
+	fs.StringArrayVarP(&opts.headerFlags, "header", "H", []string{}, "add custom headers to requests")
```
shizhMSFT, line 67: nit: in golang, nil slice is still a valid slice.
```suggestion
	fs.StringArrayVarP(&opts.headerFlags, "header", "H", nil, "add custom headers to requests")
```

shizhMSFT (approved):
LGTM with suggestions

shizhMSFT (approved):
LGTM

PR #797, opened 2023-02-10
feat!: disable auto type conversion for manifests

### LINKED ISSUE #782: Disable auto conversion of manifest types

0.16.0

Currently ORAS CLI causes artifact manifest with auto fallback to image manifest. As per recent discussions in OCI https://hackmd.io/moB5-fsQTbGmDpnrsY-4yg we feel that auto conversion of manifests leads to different digests that clients may produce. 

If a registry starts supporting the artifact manifest, then ORAS CLI would start pushing the new manifest type which might potentially break downstream consumers who might know now about the this manifest type. This makes the pipeline fragile. 

One option would be to make the default use --v1.1 and fail on push. And the clients can then specify --v1.0 which would only use the image manifest and continue using image manifest throughout its time until the producer chooses to push the new type. 

Currrently this behavior switch depends on the target registry and moving this control to the client is the main ask here. 

qweeah:
This PR changes the default media type of `oras push` and `oras attach` by removing the auto-conversion of the manifest type.

Resolves #782 

### DIFF
--- a/cmd/oras/attach.go
+++ b/cmd/oras/attach.go
@@ -160,26 +160,6 @@ func runAttach(opts attachOptions) error {
 			}
 			return content.Successors(ctx, fetcher, node)
 		}
-
-		if content.Equal(node, root) {
-			
-		}
-		if root.MediaType == ocispec.MediaTypeArtifactManifest {
-			graphCopyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
-				if content.Equal(node, root) {
-					// skip subject
-					return descs, nil
-				}
-				return content.Successors(ctx, fetcher, node)
-			}
-		} else if root.MediaType == ocispec.MediaTypeImageManifest {
-			graphCopyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
-				if content.Equal(node, root) {
-					// skip subject
-					return append(descs, 
-				return content.Successors(ctx, fetcher, node)
-			}
-		}
 		return oras.CopyGraph(ctx, store, dst, root, graphCopyOptions)
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/oras/attach.go
```diff
@@ -152,11 +152,24 @@ func runAttach(opts attachOptions) error {
 				}
 				return content.Successors(ctx, fetcher, node)
 			}
+		} else if root.MediaType == ocispec.MediaTypeImageManifest {
+			graphCopyOptions.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+				if content.Equal(node, root) {
+					// skip subject
+					return append(descs, ocispec.Descriptor{
```
Wwwsylvia, line 159: nit: Should we put the scratch config descriptor in a variable?

qweeah: currently it's not used anywhere else, we might move it to a variable when such need comes.

qweeah: also we can directly reference OCI scratch config once https://github.com/opencontainers/image-spec/pull/1023 once is merged

cmd/oras/attach.go
```diff
@@ -152,11 +152,24 @@ func runAttach(opts attachOptions) error {
...
+						// scratch config blob with content of `{}` (size of 2)
+						MediaType: "application/vnd.oci.example+json",
+						Digest:    `sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a`,
+						Size:      2,
+						Data:      []byte(`{}`)}), nil
+				}
+				return content.Successors(ctx, fetcher, node)
+			}
```
shizhMSFT, line 167: The point of line 147 - 168 is to skip the `subject` node. That is, remove the `subject` node from the successors if the current node is a root node.

shizhMSFT (changes requested):

shizhMSFT (approved):
LGTM

PR #798, opened 2023-02-10
feat!: show subject digest in oras discover

### LINKED ISSUE #789: Make ORAS discover consistent with 0.16

 1.0.0-rc.1

Oras discover used to show the fully qualified reference and it is missing that in the current RC release 

```bash
❯ oras discover -o tree sajay.azurecr.io/oci-ref-delete:v1
v1
└── example/test
    └── sha256:4471e38700eab45826c8ed6231736af30c2ad5561eabb9966fee6b79c0e35781
```

Prior to this it would show the this with fully qualified reference which IMHO was less confusing. 

```bash
❯ ./oras discover -o tree sajay.azurecr.io/oci-ref-delete:v1
sajay.azurecr.io/oci-ref-delete:v1
└── example/test
    └── sha256:4471e38700eab45826c8ed6231736af30c2ad5561eabb9966fee6b79c0e35781
```

Fully qualified references requires less context and is more usable in general. 
However for OCI layout this might be a challenge and it might something we can special case for OCI layout. 

We could even consider displaying this a fully qualified reference with the digest since it makes it clear to what digest the tag resolved to. 

Currently this looks odd when we use a digest 

```bash
❯ oras discover -o tree sajay.azurecr.io/oci-ref-delete@sha256:414e0518bb9228d35e4cd5165567fb91d26c6a214e9c95899e1e056fcd349011
sha256:414e0518bb9228d35e4cd5165567fb91d26c6a214e9c95899e1e056fcd349011
└── example/test
    └── sha256:4471e38700eab45826c8ed6231736af30c2ad5561eabb9966fee6b79c0e35781
```

qweeah:
Changes in this PR prints resolved digest for root node of tree-view output in `oras discover`.

Resolves #789

### DIFF
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -159,6 +159,9 @@ func (opts *Target) NewReadonlyTarget(ctx context.Context, common Common) (ReadO
 		if err != nil {
 			return nil, err
 		}
+		tmp := repo.Reference
+		tmp.Reference = ""
+		opts.Path = tmp.String()
 		opts.Reference = repo.Reference.Reference
 		return repo, nil
 	}
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -41,7 +41,7 @@ type Target struct {
 	RawReference string
 	Type         string
 	Reference    string //contains tag or digest
-	// contains
+	// Path contains
 	//  - path to the OCI image layout target, or
 	//  - registry and repository for the remote target
 	Path string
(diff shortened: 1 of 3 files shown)


cmd/oras/internal/option/target.go
```diff
@@ -41,7 +41,10 @@ type Target struct {
 	RawReference string
 	Type         string
 	Reference    string //contains tag or digest
-	Path         string
+	// contains
```
shizhMSFT, line 44:
```suggestion
	// Path contains
```

shizhMSFT (approved):
LGTM

Issue #799, opened 2023-02-10
Add filter to `oras repo tags` to filter by sha or tag


TerryHowe:
### What is the version of your ORAS CLI

1.0.0

### What would you like to be added?

It would be convenient to have a filter for `oras repo tags` to get all the tags associated with a digest. Something like:
```
 oras repo tags ${REGISTRY}/curated-packages/prometheus/prometheus@sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b
```
Optionally, provide a tag and list other tags associated with the same tag
```
oras repo tags ${REGISTRY}/curated-packages/prometheus/prometheus:latest
```

### Why is this needed for ORAS?

I want to know the version of the latest or I want to know the version associated with a digest

### Are you willing to submit PRs to contribute to this feature?

- [x] Yes, I am willing to implement it.

TerryHowe:
Implemented a simple WIP which works for digests.

shizhMSFT:
The idea is great. However, we still need to consider the performance.

According to https://github.com/oras-project/oras/pull/800#discussion_r1105475134, `docker.io/library/alpine` has 129 tags. Filtering requires roughly 47 seconds. Since the time complexity is linear to the number of tags in the repository, it is estimated to take 6 mins to filter a repo of 1,000 tags.

As @qweeah has suggested, we may want to
1. Introduce `--concurreny` flag to accelerate the process.
2. Users need to be warned for long running command.
3. This feature is required to marked as experimental due to UX concerns.

sajayantony:
There are throttling issues to consider since registries typically throttle the tag listing API. Also we should consider outputting a progress like curl so that users know that there is something happening rather than waiting for a long time. Higher concurrency may result in 429s easily.

TerryHowe:
I order to support concurrency, most likely need to add support for it in oras-go https://github.com/oras-project/oras-go/issues/445

shizhMSFT:
> I order to support concurrency, most likely need to add support for it in oras-go [oras-project/oras-go#445](https://github.com/oras-project/oras-go/issues/445)

Tag listing cannot be done in parallel but the process of resolving a tag to a digest can.

qweeah:
> > I order to support concurrency, most likely need to add support for it in oras-go [oras-project/oras-go#445](https://github.com/oras-project/oras-go/issues/445)
> 
> Tag listing cannot be done in parallel but the process of resolving a tag to a digest can.

[...]

PR #800, opened 2023-02-10
feat: Add ability query tags for `oras repo tags` command

TerryHowe:
Add a ability for the `oras repo tags` command to get the tags associated with a specified digest or tag. Given a tag, it will return the digest. Given a digest, it will return the tag.

Closes: https://github.com/oras-project/oras/issues/799

Signed-off-by: Terry Howe <tlhowe@amazon.com>

### DIFF
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -56,7 +56,7 @@ Example - Show tags of the target OCI layout folder 'layout-dir':
 Example - Show tags of the target OCI layout archive 'layout.tar':
   oras repo tags --oci-layout layout.tar
 
-Example - Show tags also associated with a particular tagged resource:
+Example - Show tags associated with a particular tagged resource:
   oras repo tags localhost:5000/hello:latest
 
 Example - Show tags associated with a digest:
@@ -104,14 +104,15 @@ func showTags(opts showTagsOptions) error {
 				continue
 			}
 			if filter != "" {
+				if tag == opts.Reference {
+					fmt.Println(tag)
+					continue
+				}
 				desc, err := finder.Resolve(ctx, tag)
 				if err != nil {
 					return err
 				}
 				if desc.Digest.String() != filter {
-					continue
-				}
-				if tag == opts.Reference {
 					continue
 				}
 			}
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -55,6 +55,12 @@ Example - Show tags of the target OCI layout folder 'layout-dir':
 
 Example - Show tags of the target OCI layout archive 'layout.tar':
   oras repo tags --oci-layout layout.tar
+
+Example - Show tags also associated with a particular tagged resource:
+  oras repo tags localhost:5000/hello:latest
+
+Example - Show tags associated with a digest:
+  oras repo tags localhost:5000/hello@sha256:c551125a624189
 `,
 		Args:    cobra.ExactArgs(1),
 		Aliases: []string{"show-tags"},
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -17,7 +17,6 @@ package repository
 
 import (
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/opencontainers/go-digest"
@@ -80,7 +79,7 @@ Example - Show tags associated with a digest:
 }
 
 func showTags(opts showTagsOptions) error {
-	ctx, _ := opts.SetLoggerLevel()
+	ctx, logger := opts.SetLoggerLevel()
 	finder, err := opts.NewReadonlyTarget(ctx, opts.Common)
 	if err != nil {
 		return err
@@ -97,7 +96,7 @@ func showTags(opts showTagsOptions) error {
 			}
 			filter = desc.Digest.String()
 		}
-		fmt.Fprintln(os.Stderr, "Tag query by reference may take a while")
+		logger.Infof("[Preview] Querying tags associated to %s, it may take a while.\n", filter)
 	}
 	return finder.Tags(ctx, opts.last, func(tags []string) error {
 		for _, tag := range tags {
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -80,7 +80,8 @@ func showTags(opts showTagsOptions) error {
 	}
 	filter := ""
 	if opts.Reference != "" {
-		if isDigestTag(opts.Reference) {
+		_, err := digest.Parse(opts.Reference)
+		if err == nil {
 			filter = opts.Reference
 		} else {
 			desc, err := finder.Resolve(ctx, opts.Reference)
@@ -88,7 +89,6 @@ func showTags(opts showTagsOptions) error {
 				return err
 			}
 			filter = desc.Digest.String()
-			fmt.Println(filter)
 		}
 	}
 	return finder.Tags(ctx, opts.last, func(tags []string) error {
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -61,7 +61,7 @@ Example - Show tags also associated with a particular tagged resource:
   oras repo tags localhost:5000/hello:latest
 
 Example - Show tags associated with a digest:
-  oras repo tags localhost:5000/hello@sha256:c551125a624189
+  oras repo tags localhost:5000/hello@sha256:c551125a624189cece9135981621f3f3144564ddabe14b523507bf74c2281d9b
 `,
 		Args:    cobra.ExactArgs(1),
 		Aliases: []string{"show-tags"},
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -29,7 +29,6 @@ type showTagsOptions struct {
 	option.Target
 
 	last             string
-	filter           string
 	excludeDigestTag bool
 }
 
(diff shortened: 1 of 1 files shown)


qweeah:
> Given a tag, it will return the digest.

This is an inconsistent user experience since the return value is not tags any more. Functionally it's already covered by `oras manifest get`. I think it's better to return error.

cmd/oras/repository/tags.go
```diff
@@ -78,14 +78,36 @@ func showTags(opts showTagsOptions) error {
 	if err != nil {
 		return err
 	}
+	filter := ""
 	if opts.Reference != "" {
-		return fmt.Errorf("unexpected tag or digest %q found in repository reference %q", opts.Reference, opts.RawReference)
+		if isDigestTag(opts.Reference) {
```
qweeah, line 83: use [digest. Parse](https://pkg.go.dev/github.com/opencontainers/go-digest#Parse)(opts.Reference) to validate whether a reference was digest or not.  `isDigestTag` only checks if the tag is following [fallback schema for referrer API support](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#referrers-tag-schema) or not.

cmd/oras/repository/tags.go
```diff
@@ -78,14 +78,36 @@ func showTags(opts showTagsOptions) error {
...
+			filter = opts.Reference
+		} else {
+			desc, err := finder.Resolve(ctx, opts.Reference)
+			if err != nil {
+				return err
+			}
+			filter = desc.Digest.String()
+			fmt.Println(filter)
```
qweeah, line 91: Should still return `unexpected tag` error for this code path. Return digest for a tag breaks UX consistency.

TerryHowe: I'm willing to add the filter as a flag back, but I think this way is better for the user.

qweeah: There are two kinds of features we are discussing about:
1) filtering: based on a provided tag or digest, give the tag result
2) displaying: display corresponding digests instead of tags

Let's keep this PR to the first feature. So we have below use cases:
1.1)  based on provided tag, give the matching tag list
1.2)  based on the provided digest, give the matching tag list

With out feature 2, 1.1 makes no sense since the output will always be the input tag, not a list. Making it returns digest will expand this issue's scope to #806.

TerryHowe: 1.1 makes sense if you want to know what release is `latest` which is exactly what I want to do.

qweeah: In that case shouldn't `oras manifest get --descriptor` be used?

qweeah: If the output is **tags having same digest to the input tag**, that makes sense.

qweeah: I just realize that this is the current implementation, sorry for the gap. Does below UX looks good to you?
- when input is a digest, print all tags associated with the input digest
- when input is a tag, resolve its digest, print all tags associated with the resolved digest

cmd/oras/repository/tags.go
```diff
@@ -78,14 +78,36 @@ func showTags(opts showTagsOptions) error {
...
+			filter = desc.Digest.String()
+		}
 	}
 	return finder.Tags(ctx, opts.last, func(tags []string) error {
 		for _, tag := range tags {
 			if opts.excludeDigestTag && isDigestTag(tag) {
 				continue
 			}
+			if filter != "" {
+				desc, err := finder.Resolve(ctx, tag)
```
qweeah, line 100: Scan and resolve every tag in the list will degrade the performance. Below are some experiments I made based on the current implementation:
```shell
oras repo tags docker.io/library/alpine | wc
time oras repo tags docker.io/library/alpine:latest
oras repo tags mcr.microsoft.com/cbl-mariner/distroless/minimal | wc
time oras repo tags mcr.microsoft.com/cbl-mariner/distroless/minimal:1.0
``` 

On my computer , it takes DockerHub 47.442 seconds to scan 129 tags, and MCR 19.192 seconds to scan 141 tags.


So we need to:
1) resolve descriptor and filter concurrently, also add `--concurrency` flag so user can use it to tune the behavior. This can be don in another PR.
2) print a warning log, reminding ORAS user how may tag need to be resolved and the command might take a long time to run
3) print experimental hint for this feature

cmd/oras/repository/tags.go
TerryHowe: > I just realize that this is the current implementation, sorry for the gap. Does below UX looks good to you?
> 
> * when input is a digest, print all tags associated with the input digest
> * when input is a tag, resolve its digest, print all tags associated with the resolved digest

Yes!

cmd/oras/repository/tags.go
TerryHowe:
```
tlhowe@3c22fb41240e oras % oras repo tags --ca-file "${REGISTRY_CAFILE}" ${REGISTRY}/curated-packages/prometheus/prometheus:latest
Tag query by reference may take a while
v2
tlhowe@3c22fb41240e oras % oras repo tags --ca-file "${REGISTRY_CAFILE}" ${REGISTRY}/curated-packages/prometheus/prometheus:latest 2>err
v2
```

TerryHowe: https://github.com/oras-project/oras/issues/823 on the first issue. There is another issue ins oras-go to implement the method over there

cmd/oras/repository/tags.go
```diff
@@ -55,6 +56,12 @@ Example - Show tags of the target OCI layout folder 'layout-dir':
 
 Example - Show tags of the target OCI layout archive 'layout.tar':
   oras repo tags --oci-layout layout.tar
+
+Example - Show tags also associated with a particular tagged resource:
+  oras repo tags localhost:5000/hello:latest
+
+Example - Show tags associated with a digest:
+  oras repo tags localhost:5000/hello@sha256:c551125a624189
```
qweeah, line 64: I know the digest is long but it's better to put a full 64-char digest here. Unless we want to support prefix matching for digest.

cmd/oras/repository/tags.go
qweeah: I think we can add warning and preview hint logs merge this PR first.

cmd/oras/repository/tags.go
TerryHowe: Done

cmd/oras/repository/tags.go
```diff
@@ -78,14 +85,37 @@ func showTags(opts showTagsOptions) error {
...
+		if err == nil {
+			filter = opts.Reference
+		} else {
+			desc, err := finder.Resolve(ctx, opts.Reference)
+			if err != nil {
+				return err
+			}
+			filter = desc.Digest.String()
+		}
+		fmt.Fprintln(os.Stderr, "Tag query by reference may take a while")
```
qweeah, line 100: You may get the `logger` from `opts.SetLoggerLevel()` 
```suggestion
		logger.Infof("[Preview] Querying tags associated to %s, it may take a while.", filter)
```

TerryHowe: Does that log to stderr?

TerryHowe: Yeh, that works

qweeah (approved):
LGTM with suggestions. Also, I have added test enhancement in #829. Feel free to comment if you are interested in adding them, or you come up with any new cases.

cmd/oras/repository/tags.go
qweeah: Yes, it logs to stderr.

TerryHowe: Done

qweeah:
@TerryHowe Can you help rebase the PR? @shizhMSFT Please take a look when you have time, thanks.

cmd/oras/repository/tags.go
```diff
@@ -73,19 +79,42 @@ Example - Show tags of the target OCI layout archive 'layout.tar':
...
+				desc, err := finder.Resolve(ctx, tag)
+				if err != nil {
+					return err
+				}
+				if desc.Digest.String() != filter {
+					continue
+				}
+				if tag == opts.Reference {
+					continue
+				}
```
shizhMSFT, line 116: Shouldn't the `oras repo tags` prints the input tag?

cmd/oras/repository/tags.go
```diff
@@ -55,6 +55,12 @@ Example - Show tags of the target OCI layout folder 'layout-dir':
 
 Example - Show tags of the target OCI layout archive 'layout.tar':
   oras repo tags --oci-layout layout.tar
+
+Example - Show tags also associated with a particular tagged resource:
+  oras repo tags localhost:5000/hello:latest
```
shizhMSFT, line 60: Shouldn't the `oras repo tags` prints the input tag?
```suggestion
Example - Show tags associated with a particular tagged resource:
  oras repo tags localhost:5000/hello:latest
```

shizhMSFT:
@TerryHowe The PR overall LGTM but I left some comments on the user experience.

cmd/oras/repository/tags.go
TerryHowe: I was thinking it would be more convenient for the user to not print that to save them the `grep -v latest`. They know that. I could be convinced otherwise though.

shizhMSFT: I think this is about the completion of this functionality. That is, "show all tags of this manifest" versus "show all _other_ tags of this manifest". I think the former is less complex for new users and reduces chances to have bugs in scripting. Otherwise, we need to do something like
```bash
image="docker.io/library/alpine:latest"
tags=$(oras repo tags $image)
tags="$tags ${image#*:}"
```

shizhMSFT: Another benefit to show all tags is that the output will be consistent for both inputs of digest and tag.

shizhMSFT, line 116: This if statement block can be moved to the place in between line 106 and 107 to save a `Resolve()` call.

cmd/oras/repository/tags.go
TerryHowe: Done.

cmd/oras/repository/tags.go
TerryHowe: Done.

qweeah (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #808, opened 2023-02-13
feat: Add ability to use command line from go script

TerryHowe:
What it would look like to expose the command level for use in go "scripts".

Signed-off-by: Terry Howe <tlhowe@amazon.com>

### DIFF
--- a/cmd/oras/repository/tags.go
+++ b/cmd/oras/repository/tags.go
@@ -24,7 +24,7 @@ import (
 	"oras.land/oras/cmd/oras/internal/option"
 )
 
-type showTagsOptions struct {
+type ShowTagsOptions struct {
 	option.Common
 	option.Target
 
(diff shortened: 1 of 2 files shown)


cmd/oras/repository/tags.go
```diff
@@ -97,3 +97,17 @@ func isDigestTag(tag string) bool {
 	_, err := digest.Parse(dgst)
 	return err == nil
 }
+
```
TerryHowe, line 100: This `NewRemoteTarget` would not live here, just here to keep the changes simple.

qweeah:
Can we just use [cobra.Command.Execute](https://pkg.go.dev/github.com/spf13/cobra#Command.Execute) to do that? Like this PR: https://github.com/oras-project/oras/pull/810/files

TerryHowe:
Closing in favor of https://github.com/oras-project/oras/pull/810

Issue #807, opened 2023-02-13
Make public command level methods so users can script in go


TerryHowe:
### What is the version of your ORAS CLI

1.0.0

### What would you like to be added?

Make cmd level methods public so that users can write scripts in go instead of shell. What would be required to do this is just expose the commands, associated options and some aspects of `cmd/oras/internal/option`. Users would create options and call the command. For example for `oras repo tags`, `showTagsOptions` and `showTags` would be made public. The user would also need a way to create a `Target`

### Why is this needed for ORAS?

Shell is pretty terrible to write, test and maintain. It would be a lot easier in go.

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

shizhMSFT:
Linking alternative solution #810

sajayantony:
From previous discussions, anything made public and GA’d should technically be non breaking and also have test coverage. If there is valid ask and the community is able to come to an agreement that this would benefit the project and is ready to cost/implement then it makes sense to prioritize this item.

shizhMSFT:
Based on the community call, Billy's cobra-based proposal is accepted. To implement, code refactoring work is required.

qweeah:
This issue depends on https://github.com/oras-project/oras/issues/824.

PR #810, opened 2023-02-14
DO-NOT-MERGE: example of scripting in go

qweeah:
Related to #807 


TerryHowe (commented):
Ah better yet. Maybe add something to the docs for people like me that haven't dug through Cobra in great detail?

qweeah:
> Ah better yet. Maybe add something to the docs for people like me that haven't dug through Cobra in great detail?

Maybe a runnable godoc example? There are still some refactoring work to do before user can really use this, some commands are in the main package and cannot be exported.

qweeah:
closing as it's already covered in #825

PR #819, opened 2023-02-16
fix: fix CODEOWNERS

shizhMSFT:
Having multiple entries with the same path results in that only the last entry takes effect.
This PR combines multiple entries with the same path into a single entry.


TerryHowe (approved):
LGTM

PR #825, opened 2023-02-20
refactor: add root package for Cobra commands

### LINKED ISSUE #807: Make public command level methods so users can script in go

### LINKED ISSUE #824: refactor CLI main package
Currently both the root command and some sub-commands are implemented in the main package, which is not a good practice for Cobra.

Also to satisfy #807, all the command types should be moved out of the main package with root command type exported publicly.

qweeah:
This PR
- Adds a new root command type to align with Cobra good practices.
~~- Adds a root context so all subcommand execution can use.~~ Removing this, will add changes in another PR to make review easier
~~- Adds a new godoc runnable example for go scripting.~~ Removing this from this PR since it causes coverage drop instantly. To fix the coverage drop, I will propose another change by utilizing [instrumented build](https://github.com/golang/proposal/blob/master/design/51430-revamp-code-coverage.md#running-instrumented-applications).

Related to #824, resolves #807


cmd/oras/cmd/example_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"strings"
+
+	"oras.land/oras/cmd/oras/cmd"
+)
+
+// Example gives playable snippets of using oras CLI in command-line scripting
+// via Golang.
+func Example() {
+	// change below array to play with your own cmd args:
+	args := []string{"repo", "ls", "mcr.microsoft.com"}
```
qweeah, line 28: Even though this is a testable example, but it don't have any `// output` comment so it will be ignored by unit test, i.e. it will be a runnable example in godoc after the next release but it won't be run in CI.

qweeah: Looks like checking in the file will heavily degrade the test coverage ratio by 34%, need to properly set codecov to ignore command type code.

qweeah:
Looks like the ignore settings are not working, need to figure out why, reset to draft.

TerryHowe (approved):
/lgtm

TerryHowe (approved):

qweeah:
Let's hold the merging of this PR before module name changed from `oras.land` to `github.com`

TerryHowe:
> Let's hold the merging of this PR before module name changed from `oras.land` to `github.com`

Not sure why we'd hold this for that.

qweeah:
> > Let's hold the merging of this PR before module name changed from `oras.land` to `github.com`
> 
> Not sure why we'd hold this for that.

Currently oras CLI is delivered as a pre-built binary which is not effected by the module name change. If this PR is merge, there might be potential users who write go scripts based on `oras.land/oras/cmd/oras/root`and make this repo subject to the module name change.

cmd/oras/main.go
```diff
@@ -16,35 +16,18 @@ limitations under the License.
...
-		blob.Cmd(),
-		manifest.Cmd(),
-		tag.TagCmd(),
-		repository.Cmd(),
-	)
-	if err := cmd.Execute(); err != nil {
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer cancel()
+	if err := root.New().ExecuteContext(ctx); err != nil {
+		cancel()
```
shizhMSFT, line 30: nit: we don't need to cancel it since the execution is already done.

shizhMSFT:
> If this PR is merged, there might be potential users who write go scripts based on `oras.land/oras/cmd/oras/root`and make this repo subject to the module name change.

It is OK as long as we change the module name before the next release.

cmd/oras/main.go
qweeah: resolved.

shizhMSFT (approved):
LGTM

PR #830, opened 2023-02-21
test(e2e): take registry logs when e2e fails

qweeah:
This PR takes the last 200 lines of registry logs when the Ginkgo test run fails with any error.

Raising this since we had an unexpected error happened on distribution side during E2E run (see [logs](https://github.com/oras-project/oras/actions/runs/4228638352/jobs/7344264732#step:6:1025)). Since there is no distribution logs taken, it's very hard to trouble-shoot the failed E2E case. Adding this PR so we can have sth to look into when such random flaky error comes next time.

### DIFF
--- a/test/e2e/scripts/e2e.sh
+++ b/test/e2e/scripts/e2e.sh
@@ -58,8 +58,7 @@ run_registry \
   $ORAS_REGISTRY_FALLBACK_PORT
 
 echo " === run tests === "
-ginkgo -r -p --succinct suite || test_fail=true
-if [ "$test_fail" = 'true' ]; then 
+if ! ginkgo -r -p --succinct suite; then 
   echo '-------- oras distribution trace -------------'
   docker logs -t --tail 200 $oras_container_name
   echo '-------- upstream distribution trace -------------'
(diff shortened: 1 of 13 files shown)


wangxiaoxuan273:
LGTM. Thanks for the contribution.

test/e2e/scripts/e2e.sh
```diff
@@ -41,19 +41,28 @@ if [ "$clean_up" = '--clean' ]; then
...
   registry:2.8.1 \
-  oras-e2e-fallback \
+  $upstream_container_name \
   $ORAS_REGISTRY_FALLBACK_PORT
 
 echo " === run tests === "
-ginkgo -r -p --succinct suite
\ No newline at end of file
+ginkgo -r -p --succinct suite || test_fail=true
+if [ "$test_fail" = 'true' ]; then
```
TerryHowe, line 62: This is fine, but a bit superfluous 
```suggestion
if ! ginkgo -r -p --succinct suite; then 
```

TerryHowe (approved):
LGTM

PR #835, opened 2023-02-27
fix: flaky test on multi referrers

qweeah:
This PR fixes a flaky unit test case found in ([CI logs](https://github.com/oras-project/oras/actions/runs/4260457314/jobs/7413642377#step:5:315)). 

The case is comparing a referrer array with provided value, which might fail when order of the elements is different. Reduced the length of array to one so the test result is determined.


shizhMSFT (approved):
LGTM

PR #841, opened 2023-02-28
feat: return error when header flags used on oci layout

### LINKED ISSUE #836: should fail directly if customer headers are specified for OCI image layout target
Customer headers flag `-H/-header` only affects all out-going http requests. When using oras cli to operate OCI image layout targets, those headers will be ignored automatically. Thus, it's better to return error directly.

qweeah:
This PR add check on OCI layout target, returning error when any custom header is provided.

Fixes #836 


cmd/oras/internal/option/target.go
```diff
@@ -86,6 +87,9 @@ func (opts *Target) Parse() error {
 	switch {
 	case opts.isOCILayout:
 		opts.Type = TargetTypeOCILayout
+		if len(opts.headerFlags) != 0 {
+			return errors.New("customer header([--{from|to}]-header/H) cannot be used on OCI image layout target")
```
TerryHowe, line 91:
```suggestion
			return errors.New("customer header flags cannot be used on an OCI image layout target")
```
What you have is more correct, but maybe this would do.

TerryHowe (approved):
LGTM

shizhMSFT (approved):
LGTM

cmd/oras/internal/option/target.go
```diff
@@ -86,6 +87,9 @@ func (opts *Target) Parse() error {
 	switch {
 	case opts.isOCILayout:
 		opts.Type = TargetTypeOCILayout
+		if len(opts.headerFlags) != 0 {
+			return errors.New("customer header flags cannot be used on an OCI image layout target")
```
shizhMSFT, line 91: nit: typo
```suggestion
			return errors.New("custom header flags cannot be used on an OCI image layout target")
```

shizhMSFT (changes requested):

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #852, opened 2023-03-03
fix: Improve error message on oras copy.go with  non-existent tags

### LINKED ISSUE #791: Improve error message on `oras copy` with non-existent tags

Error message is not clear when the tag `3.9` does not exist in the `alpine` folder.

```console
$ oras cp --from-oci-layout alpine:3.9 localhost:5000/alpine
Error: not found
```

The `not found` is correct but it is hard to let the users know whether the folder does not exist or the tag does not exist.

I do notice the error message was improved in [oras-go](https://github.com/oras-project/oras-go/blob/v2.0.0/copy.go#L138) but the following code is not improved:

https://github.com/oras-project/oras/blob/586fa838f00ea406696a29a571a90fab3f5a1cca/cmd/oras/cp.go#L139-L149

Print something like
```
Error: failed to resolve alpine:3.9: not found
```

### How can we reproduce it?

```sh
oras cp docker.io/library/alpine:3.9 --to-oci-layout alpine
oras cp --from-oci-layout alpine:3.9 localhost:5000/alpine
```

Version:        1.0.0-rc.1
Go version:     go1.19.5
Git commit:     a98931bc68a0e7a5de6c51df8e5aa09aadad3057
Git tree state: clean

Ubuntu 20.04

khanjasir90:
Improves Error message on oras copy.go with non-existent tags
Fixes #791 

### DIFF
--- a/cmd/oras/cp.go
+++ b/cmd/oras/cp.go
@@ -139,7 +139,7 @@ func runCopy(opts copyOptions) error {
 	if dstRef := opts.To.Reference; dstRef == "" {
 		desc, err = oras.Resolve(ctx, src, opts.From.Reference, rOpts)
 		if err != nil {
-			return fmt.Errorf("failed to resolve %s: %w", src, err)
+			return fmt.Errorf("failed to resolve %s: %w", opts.From.Reference, err)
 		}
 		if opts.recursive {
 			err = oras.ExtendedCopyGraph(ctx, src, dst, desc, extendedCopyOptions.ExtendedCopyGraphOptions)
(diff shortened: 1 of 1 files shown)


khanjasir90:
@shizhMSFT @qweeah could u pls review it!

cmd/oras/cp.go
```diff
@@ -139,7 +139,7 @@ func runCopy(opts copyOptions) error {
 	if dstRef := opts.To.Reference; dstRef == "" {
 		desc, err = oras.Resolve(ctx, src, opts.From.Reference, rOpts)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to resolve %s: %w", src, err)
```
qweeah, line 142:
```suggestion
			return fmt.Errorf("failed to resolve %s: %w", opts.From.Reference, err)
```
A new ` oras.Resolve` call is added in line 154, better to copy the solution there also.

qweeah (changes requested):

cmd/oras/cp.go
shizhMSFT, line 142: `src` is not a string type

shizhMSFT (changes requested):

khanjasir90:
@qweeah and @shizhMSFT I have made necessary give changes :)

shizhMSFT (approved):
LGTM

PR #880, opened 2023-03-14
test(e2e): layout specs for `oras manifest push`

### LINKED ISSUE #879: add OCI image layout e2e test specs for `oras manifest push`
**Good Paths**
 - should push a manifest from stdin with/without tagging
 - should push a manifest from file with/without media type
 - should be able to output the descriptor of pushed manifest

**Bad Paths**
 - should fail if media type is not in the manifest and `--media-type` flag is not set

qweeah:
This PR adds e2e test specs for `oras manifest push`

Fixes #879 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v3 v3.0.1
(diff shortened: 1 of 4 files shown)


qweeah:
E2E is expected to fail before the https://github.com/oras-project/oras-go/issues/461 is fixed.

shizhMSFT (approved):
LGTM

PR #899, opened 2023-03-23
refactor: unify command contexts

qweeah:
Makes all oras commands executed based on the same context passed from root command. Depends on merging of #825

Related to #824


cmd/oras/internal/option/common.go
```diff
@@ -30,3 +34,8 @@ func (opts *Common) ApplyFlags(fs *pflag.FlagSet) {
 	fs.BoolVarP(&opts.Debug, "debug", "d", false, "debug mode")
 	fs.BoolVarP(&opts.Verbose, "verbose", "v", false, "verbose output")
 }
+
+// WithLogger sets a logger to ctx with proper logging options tuned.
+func (opts *Common) WithLogger(ctx context.Context) (context.Context, logrus.FieldLogger) {
```
TerryHowe, line 39: I agree something needs to be done here, but since you are passing in a context here, I would think this would be called `WithContext`

qweeah: Good idea, resolved.

This PR is pending on the merge of #825, I picked this PR out so the review work is easier.

TerryHowe (approved):
LGTM

cmd/oras/root/attach.go
```diff
@@ -98,8 +97,8 @@ Example - Attach file to the manifest tagged 'v1' in an OCI layout folder 'layou
 	return cmd
 }
 
-func runAttach(opts attachOptions) error {
-	ctx, _ := trace.NewLogger(opts.Debug, opts.Verbose)
+func runAttach(opts attachOptions, cmd *cobra.Command) error {
```
shizhMSFT, line 100: nit: code style: Can we put `*cobra.Command` before `opts`?

cmd/oras/root/attach.go
```diff
@@ -98,8 +97,8 @@ Example - Attach file to the manifest tagged 'v1' in an OCI layout folder 'layou
...
+	ctx, _ := opts.WithContext(cmd.Context())
```
shizhMSFT, line 101: What's the reason to pass `*cobra.Command` instead of `context.Context`?

shizhMSFT (approved):
LGTM

PR #905, opened 2023-03-25
test(cov): make unit tests cover more

qweeah:
This PR introduces no change but only does some code clean and makes code coverage looks better.

### DIFF
--- a/internal/crypto/certificate_test.go
+++ b/internal/crypto/certificate_test.go
@@ -1,0 +1,92 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crypto
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"net/http"
+	"net/http/httptest"
+	"os"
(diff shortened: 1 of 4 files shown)


internal/graph/graph_test.go
```diff
@@ -157,6 +159,7 @@ func TestReferrers(t *testing.T) {
 		{"should return referrers when target is a referrer lister", args{ctx, &refLister{referrers: referrers}, ocispec.Descriptor{}, ""}, referrers, false},
 		{"should return nil for index node", args{ctx, finder, descs[index], ""}, nil, false},
 		{"should return nil for config node", args{ctx, finder, descs[imgConfig], ""}, nil, false},
+		{"should return nil for blob/layer node", args{ctx, finder, descs[blob], ""}, nil, false},
```
TerryHowe, line 162: table tests make more sense in situations like this, but even then you can use a helper function instead

internal/crypto/certificate_test.go
```diff
@@ -0,0 +1,92 @@
...
+	tests := []struct {
+		name    string
+		args    args
+		want    *x509.CertPool
+		wantErr bool
+	}{
+		{"should fail if the path is doesn't exist", args{"/???"}, nil, true},
+		{"should fail if the file is not a pem", args{pemPath}, nil, true},
+	}
+	for _, tt := range tests {
```
TerryHowe, line 80: I dislike table tests because they add unnecessary complexity and can be more troublesome to debug. This would actually be less code probably with just two tests and the stack trace would tell you which test was broken. I see there are several table tests of size one which make very little sense.

TerryHowe (approved):
LGTM

Not a table test fan, but that is just opinion.

internal/crypto/certificate_test.go
qweeah: Thanks for the advice. I am using table test here since it's the default template for unit test auto generation extension. Will change this to two separate unit tests.

TerryHowe (approved):
LGTM

PR #903, opened 2023-03-23
test: generate coverage profile during e2e tests

qweeah:
**What this PR does / why we need it**:
This PR updates E2E test infra to generate test coverage profile when running e2e test. If the build is triggered by tagging and the tag starts with `v` (a release), e2e test will be ran directly on non-instrumented application.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #900.

**Please check the following list**:
- [x]  Does the affected code have corresponding tests, e.g. unit test, E2E test?
- [ ]  Does this change require a documentation update?
- [ ]  Does this introduce breaking changes that would require an announcement or bumping the major version?
- [ ]  Do all new files have an appropriate license header?

PR #906, opened 2023-03-27
test(cov): add e2e tests to make coverage looks better 

qweeah:
The PR added e2e tests to make coverage looks better. Depends on merging of #903. `oras login` is refactored so e2e tests can mock user input.


qweeah:
closing this since the work should be done in #910

PR #910, opened 2023-03-29
refactor: replace terminal package used by `oras login`

### LINKED ISSUE #908: Update `github.com/moby/term` to `golang.org/x/term`

v1.0.0

Update `github.com/moby/term` to `golang.org/x/term`

`oras` uses untagged version of `github.com/moby/term` for reading passwords in `oras login` since the package `term` does not have any tags.

https://github.com/oras-project/oras/blob/67dca8e1d5a9f6c3dae7a5eb61719a94155aab21/go.mod#L7

By updating to `golang.org/x/term`, we can also remove the dependencies on 

https://github.com/oras-project/oras/blob/67dca8e1d5a9f6c3dae7a5eb61719a94155aab21/go.mod#L19

and

https://github.com/oras-project/oras/blob/67dca8e1d5a9f6c3dae7a5eb61719a94155aab21/go.mod#L25

qweeah:
This PR updates the underlying package of `oras login` with minor refactoring.

Fixes #908 


cmd/oras/login.go
```diff
@@ -76,30 +75,37 @@ Example - Log in with username and password in an interactive terminal and no TL
...
 	ctx, _ := trace.NewLogger(opts.Debug, opts.Verbose)
 
-	// prompt for credential
 	if opts.Password == "" {
+		// prompt for credential
+		reader := bufio.NewReader(os.Stdin)
 		if opts.Username == "" {
-			// prompt for username
-			username, err := readLine("Username: ", false)
+			fmt.Print("username: ")
```
shizhMSFT, line 82: We should keep it `Username` instead of `username`. Same applies to other parts.

cmd/oras/login.go
```diff
@@ -131,26 +137,15 @@ func runLogin(opts loginOptions) (err error) {
 	return nil
 }
 
-func readLine(prompt string, silent bool) (string, error) {
-	fmt.Print(prompt)
-	if silent {
-		fd := os.Stdin.Fd()
-		state, err := term.SaveState(fd)
+func readLine(reader *bufio.Reader) (content []byte, err error) {
```
shizhMSFT, line 140: To update the `term` package, we should touch the `readline` function only and should not touch the `runLogin`.

qweeah: >  we should touch the `readline` function only and should not touch the `runLogin`.

I have updated a version according to this suggestion. It's tricky that the newly added E2E specs are expected to fail since the bufio `reader` is generated within the function call, which consumes all mocked stdin during scanning for username and leaves password/token empty.

Can we at least make `reader` as parameter of  `readLine` so it's possible to do E2E tests via mocked stdin?

qweeah: fixed with self-implemented scanning

cmd/oras/login.go
```diff
@@ -133,24 +133,35 @@ func runLogin(opts loginOptions) (err error) {
...
-		term.DisableEcho(fd, state)
-		defer term.RestoreTerminal(fd, state)
-	}
-
-	reader := bufio.NewReader(os.Stdin)
-	line, _, err := reader.ReadLine()
-	if err != nil {
-		return "", err
+		line = string(bytes)
+	} else {
```
TerryHowe, line 144: No easier way?

qweeah: Indeed. This is not relevant to term package. The aim is to implement a scanf-like function to read from **non-terminal** stdin.

Tried [reader.ReadLine](https://pkg.go.dev/bufio#Reader.ReadLine) and [fmt.Fscanf](https://pkg.go.dev/fmt#Fscanln) but both failed. The former will read extra bytes from stdin and the latter[ skips newline](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/fmt/scan.go;l=828) before scanning starts.

TerryHowe (approved):
LGTM

internal/io/io.go
shizhMSFT, line 1: We should keep the original package name since `PrepareManifestContent` and `PrepareBlobContent` are not related to generic I/O.

internal/io/io.go
```diff
@@ -131,3 +131,29 @@ func ParseMediaType(content []byte) (string, error) {
 	}
 	return manifest.MediaType, nil
 }
+
+// ReadLine reads a line from the reader with trailing \r dropped.
+func ReadLine(reader io.Reader) ([]byte, error) {
```
shizhMSFT, line 136: We can put this function in the package `internal/ioutil`.

internal/io/io.go
```diff
@@ -131,3 +131,29 @@ func ParseMediaType(content []byte) (string, error) {
...
+		if s == "\r" {
+			drop = 1
+		} else if s == "\n" {
+			break
+		} else {
+			drop = 0
+		}
+		line = append(line, b[0])
+	}
+	return line[:len(line)-drop], nil
```
shizhMSFT, line 158: The implementation can be simplified as follows.
```suggestion
	var line []byte
	var buffer [1]byte
	for {
		n, err := reader.Read(b[:])
		if err != nil {
			if err == io.EOF {
				return line, nil
			}
			return nil, err
		}
		if n == 0 {
			continue
		}
		switch c := buffer[0]; c {
		case '\r', '\n':
			return line, nil
		default:
			line = append(line, c)
		}
	}
```

cmd/oras/root/login.go
```diff
@@ -133,24 +133,21 @@ func runLogin(opts loginOptions) (err error) {
...
+	fd := int(os.Stdin.Fd())
+	var bytes []byte
+	var err error
+	if silent && term.IsTerminal(fd) {
+		bytes, err = term.ReadPassword(fd)
+		fmt.Println(string(bytes))
+	} else {
+		bytes, err = io.ReadLine(os.Stdin)
+		fmt.Println(string(bytes))
 	}
```
shizhMSFT, line 145: `bytes` should be hidden and not printed.
```suggestion
	fd := int(os.Stdin.Fd())
	var bytes []byte
	var err error
	if silent && term.IsTerminal(fd) {
		bytes, err = term.ReadPassword(fd)
	} else {
		bytes, err = io.ReadLine(os.Stdin)
	}
```

cmd/oras/root/login.go
```diff
@@ -133,24 +133,21 @@ func runLogin(opts loginOptions) (err error) {
...
-
-	reader := bufio.NewReader(os.Stdin)
-	line, _, err := reader.ReadLine()
 	if err != nil {
 		return "", err
 	}
 	if silent {
 		fmt.Println()
 	}
```
shizhMSFT, line 151: `Println` is only required when `term.ReadPassword` is called.

shizhMSFT (changes requested):
Request changes

internal/io/io.go
qweeah: Maybe you are referring to `ParseMediaType`, which does no read at all? The function is not related to file either.

internal/io/io.go
qweeah: 1) according to https://pkg.go.dev/os#File.Read, if we are not doing zero-byte read, the returned read length `n` can only be 0 when EOF, why bother checking?
2) There is an issue in this implementation: ReadLine(strings.NewReader(\`abcde\r\n\`)) will leave unread `\n` in stdin.

Changed without including the above two

qweeah: Discussed offline, will change according to the below UX:
1. string ends CRLF: read until LF, remove CR and return left string
2. string ends with CR and EOF: remove CR and return left string
3. string contains CR before the last character: return string directly

internal/io/io.go
qweeah: Changed accordingly

internal/io/io.go
```diff
@@ -0,0 +1,53 @@
...
+			line = append(line, c)
+		case '\n':
+			// a line ends with \n
+			return line[:len(line)-drop], nil
+		default:
+			drop = 0
+			line = append(line, c)
+		}
+	}
+}
```
shizhMSFT, line 53: nit:
```suggestion
func ReadLine(reader io.Reader) ([]byte, error) {
	var line []byte
	var buffer [1]byte
	for {
		n, err := reader.Read(buffer[:])
		if err != nil {
			if err == io.EOF {
				break
			}
			return nil, err
		}
		if n == 0 {
			continue
		}
		c := buffer[0]
		if c == '\n' {
			break
		}
		line = append(line, c)
	}
	return bytes.TrimSuffix(line, []byte{'\r'}), nil
}
```

qweeah: resolved

shizhMSFT (approved):
LGTM

PR #920, opened 2023-04-06
feat: retry plain http in secure mode

### LINKED ISSUE #914: Avoid requiring passing --plain-http when specifying --insecure

 1.0.0

Currently if the users specifies --insecure the call fails if the server is not responding in http. 

```
oras version Version:    1.0.0+Homebrew Go version: go1.20.2 
oras login -u user--insecure remote-registry:5000 
Error: Get "https://remote-registry:5000/v2/": http: server gave HTTP response to HTTPS client
```
Users are expected to pass a --plain http as well in this case and we can follow precedence from docker for this to make the experience better - https://github.com/docker/cli/blob/de0d30ff24fad55df89adcb3c01d10bb1821c8db/cli/registry/client/client.go#L131

Improve user experience. 

qweeah:
This PR avoid requiring passing `--plain-http` when `--insecure` is specified

Fixes #914 


TerryHowe:
I'm confused by this because normally a `--insecure` option does not necessarily mean http

qweeah:
> I'm confused by this because normally a `--insecure` option does not necessarily mean http

Feeling the same. But it is a [proven user experience](https://docs.docker.com/registry/insecure/) in docker. Users can provide one less flag `--plain-http` with minor performance cost. Considering the flag `--insecure` is targeting non-prod usage, such cost is acceptable.

sajayantony:
+1 on reducing the number of flags in ORAS CLI. My understanding is that this is a fallback behavior and not that --insecure will make the transport default to http. Is that the case?

test/e2e/suite/command/resolve_host.go
```diff
@@ -48,7 +48,7 @@ func ResolveFlags(reg string, host string, flagType resolveType) []string {
 	resolveFlag := "resolve"
 	usernameFlag := "username"
 	passwordFlag := "password"
-	plainHttpFlag := "plain-http"
+	plainHttpFlag := "insecure"
```
TerryHowe, line 51: Should this var be renamed? A test for plain-http?

qweeah: I was meant to test the https->http auto handling via `--insecure`. This PR is pending for the discussion in https://github.com/oras-project/oras/issues/914, feel free to add your comment there.

TerryHowe: Right, but the var name is `plainHttpFlag` and the value is `insecure`

TerryHowe:
Should we kill this PR? Just because docker does something does not make it the correct approach in my book.

qweeah:
Let's close it and keep tracking in #914

PR #921, opened 2023-04-07
fix!: ignore cleanup error when attaching towards OCI 1.0 registry

### LINKED ISSUE #915: Improve the error msg for OCI v1.0 registries or registries that don't support deletion API

When using ORAS CLI to interact with OCI v1.0 compliant registries such as GHCR or registries that [don't support deletion API](https://github.com/docker/hub-feedback/issues/1759) (e.g. Docker Hub), ORAS will return an unexpected error message and may mislead or confuse users. It happens when attaching an artifact to the image for the second time.

Improve the error message to be more descriptive and user-friendly, and tell users about the right result exactly. Giving an example: This artifact has been attached to xxx. ORAS can't delete the referrers index since this is an OCI v1.0 compliant registry.

### How can we reproduce it?

I have attached a file to the sample Alpine image in GHCR. Then I attached another file to the same image and it returns the error as follows:

```
oras attach --artifact-type=sbom/test ghcr.io/feynmanzhou/alpine:latest sbom.json -u feynmanzhou --password-stdin
Uploading 57761fa432ee sbom.json
Uploaded  57761fa432ee sbom.json
Error: failed to delete dangling referrers index sha256:de0b503017452ba6b070981414bfc31f8e6096852a36ff4f3223fdc0ff411566 for referrers tag sha256-1304f174557314a7ed9eddb4eab12fed12cb0cd9809e4c28f29af86979a3c870: DELETE "https://ghcr.io/v2/feynmanzhou/alpine/manifests/sha256:de0b503017452ba6b070981414bfc31f8e6096852a36ff4f3223fdc0ff411566": response status code 405: unsupported: The operation is unsupported.
```

It may confuse users why this `attach` behavior failed due to an old index manifest deletion failure. In fact, this file has been attached to the image successfully but it doesn't tell users the right result unless users verify it using `oras discover` as below.  

```
oras discover ghcr.io/feynmanzhou/alpine:latest -o tree -u feynmanzhou --password-stdin
ghcr.io/feynmanzhou/alpine@sha256:1304f174557314a7ed9eddb4eab12fed12cb0cd9809e4c28f29af86979a3c870
├── application/vnd.cncf.notary.signature
│   └── sha256:85f65b60a6e260d52ba5f8d556d444320b92c2ac140ec8d4944dbd8b03deb18c
└── sbom/test
    └── sha256:ee375fb1111ca3d0c60663115e2cdce3e7061821ab26059852ecc43ceb8ee8cd


```
Version:        1.0.0
Go version:     go1.20.2
Git commit:     b58e7b910ca556973d111e9bd734a71baef03db2
Git tree state: clean
```

Ubuntu 20.08

### PR DESCRIPTION
**Pull request by qweeah**
This resolves #915 

### DIFF
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -40,7 +40,6 @@ func IsReferrersIndexDelete(err error, logger logrus.FieldLogger, path string) b
 	if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
 		return false
 	}
-	logger.Info("Failed to delete the referrers index: %s@%s", path, re.Subject.Digest)
-	logger.Info("Attached successfully but the removal of outdated referrers index from the remote registry failed. Garbage collection may be required.")
+	logger.Info("Failed to remove the outdated referrers index: %s@%s. Garbage collection may be required.", path, re.Subject.Digest)
 	return true
 }
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -16,9 +16,12 @@ limitations under the License.
 package errors
 
 import (
+	"errors"
 	"fmt"
 
+	"github.com/sirupsen/logrus"
 	"oras.land/oras-go/v2/registry"
+	"oras.land/oras-go/v2/registry/remote"
 )
 
 // NewErrInvalidReference creates a new error based on the reference string.
@@ -29,4 +32,15 @@ func NewErrInvalidReference(ref registry.Reference) error {
 // NewErrInvalidReferenceStr creates a new error based on the reference string.
 func NewErrInvalidReferenceStr(ref string) error {
 	return fmt.Errorf("%s: invalid image reference, expecting <name:tag|name@digest>", ref)
+}
+
+// IsReferrersIndexDelete checks if err is a referrers index delete error.
+func IsReferrersIndexDelete(err error, logger logrus.FieldLogger, path string) bool {
+	var re *remote.ReferrersError
+	if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
+		return false
+	}
+	logger.Info("Failed to delete the referrers index: %s@%s", path, re.Subject.Digest)
+	logger.Info("Attached successfully but the removal of outdated referrers index from the remote registry failed. Garbage collection may be required.")
+	return true
 }
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -35,11 +35,11 @@ func NewErrInvalidReferenceStr(ref string) error {
 }
 
 // IsReferrersIndexDelete checks if err is a referrers index delete error.
-func IsReferrersIndexDelete(err error, logger logrus.FieldLogger, path string) bool {
+func IsReferrersIndexDelete(err error, logger logrus.FieldLogger, opDone string, path string) bool {
 	var re *remote.ReferrersError
 	if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
 		return false
 	}
-	logger.Info("Failed to remove the outdated referrers index: %s@%s. Garbage collection may be required.", path, re.Subject.Digest)
+	logger.Info("Successfully %s but failed to remove the outdated referrers index: %s@%s. Garbage collection may be required.", opDone, path, re.Subject.Digest)
 	return true
 }
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -26,6 +26,7 @@ import (
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/content/file"
+	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras/cmd/oras/internal/option"
 	"oras.land/oras/internal/graph"
 )
@@ -38,6 +39,7 @@ type attachOptions struct {
 
 	artifactType string
 	concurrency  int
+	strict       bool
 }
 
 func attachCmd() *cobra.Command {
@@ -91,6 +93,7 @@ Example - Attach file to the manifest tagged 'v1' in an OCI layout folder 'layou
 
 	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
 	cmd.Flags().IntVarP(&opts.concurrency, "concurrency", "", 5, "concurrency level")
+	cmd.Flags().BoolVarP(&opts.strict, "strict", "", false, "strictly handles any ignorable errors as fatal")
 	cmd.MarkFlagRequired("artifact-type")
 	opts.EnableDistributionSpecFlag()
 	option.ApplyFlags(&opts, cmd.Flags())
@@ -164,13 +167,15 @@ func runAttach(ctx context.Context, opts attachOptions) error {
 
 	root, err := pushArtifact(dst, pack, copy)
 	if err != nil {
-		if strings.Contains(err.Error(), "failed to delete dangling referrers index") {
-			// attach is successful but failed to clean up obsolete referrer index
-			// TODO: more deterministic detection based on https://github.com/oras-project/oras-go/issues/479
-			logger.Warn("attach succeeds with minor cleanup issue: %s", err.Error())
-		} else {
+		if opts.strict {
 			return err
 		}
+		var re *remote.ReferrersError
+		if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
+			// not referrer index delete error
+			return err
+		}
+		logger.Info("Attach is successful but removal of outdated referrers index from remote registry failed. Garbage collection may be required.")
 	}
 
 	digest := subject.Digest.String()
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -39,7 +39,6 @@ type attachOptions struct {
 
 	artifactType string
 	concurrency  int
-	strict       bool
 }
 
 func attachCmd() *cobra.Command {
@@ -93,7 +92,6 @@ Example - Attach file to the manifest tagged 'v1' in an OCI layout folder 'layou
 
 	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
 	cmd.Flags().IntVarP(&opts.concurrency, "concurrency", "", 5, "concurrency level")
-	cmd.Flags().BoolVarP(&opts.strict, "strict", "", false, "strictly handles warning as fatal error")
 	cmd.MarkFlagRequired("artifact-type")
 	opts.EnableDistributionSpecFlag()
 	option.ApplyFlags(&opts, cmd.Flags())
@@ -166,13 +164,8 @@ func runAttach(ctx context.Context, opts attachOptions) error {
 	}
 
 	root, err := pushArtifact(dst, pack, copy)
-	if err != nil {
-		if opts.strict {
-			return err
-		}
-		if !oerrors.IsReferrersIndexDelete(err, logger, opts.Path) {
-			return err
-		}
+	if err != nil && !oerrors.IsReferrersIndexDelete(err, logger, opts.Path) {
+		return err
 	}
 
 	digest := subject.Digest.String()
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -167,13 +167,12 @@ func runAttach(ctx context.Context, opts attachOptions) error {
 
 	root, err := pushArtifact(dst, pack, copy)
 	if err != nil {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: cmd/oras/root/attach.go [hunk 1]
```diff
@@ -164,7 +164,13 @@ func runAttach(ctx context.Context, opts attachOptions) error {
 
 	root, err := pushArtifact(dst, pack, copy)
 	if err != nil {
-		return err
+		if strings.Contains(err.Error(), "failed to delete dangling referrers index") {
```
TerryHowe, line 167: this looks fragile. Is there anything we can do to pushArtifact or the underlying code to return something else or ignore it at that level?

qweeah: Indeed it's fragile. Created an issue on SDK side https://github.com/oras-project/oras-go/issues/479

sajayantony: Given that notation is also blocked on this, my vote would be to focus on returning a referrers manifest delete failure and handle that here instead of string checking.

cmd/oras/root/attach.go
```diff
@@ -91,14 +95,15 @@ Example - Attach file to the manifest tagged 'v1' in an OCI layout folder 'layou
 
 	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
 	cmd.Flags().IntVarP(&opts.concurrency, "concurrency", "", 5, "concurrency level")
+	cmd.Flags().BoolVarP(&opts.strict, "strict", "", false, "strictly handles any ignorable errors as fatal")
```
shizhMSFT, line 98: @FeynmanZhou @yizha1 Any comments on the flag name?

cmd/oras/root/attach.go
```diff
@@ -164,7 +169,19 @@ func runAttach(ctx context.Context, opts attachOptions) error {
...
+		var re *remote.ReferrersError
+		if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
+			// not referrer index delete error
+			return err
+		}
+		var errResp *errcode.ErrorResponse
+		if !errors.As(re.Err, &errResp) || errResp.StatusCode != http.StatusMethodNotAllowed {
+			return err
+		}
+		logger.Info("Attach is successful but removal of outdated referrers index from remote registry failed. Garbage collection may be required.")
```
shizhMSFT, line 184: Is this an `Info` or `Warn`?

cmd/oras/root/attach.go
```diff
@@ -164,7 +169,19 @@ func runAttach(ctx context.Context, opts attachOptions) error {
...
+		}
+		var re *remote.ReferrersError
+		if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
+			// not referrer index delete error
+			return err
+		}
+		var errResp *errcode.ErrorResponse
+		if !errors.As(re.Err, &errResp) || errResp.StatusCode != http.StatusMethodNotAllowed {
+			return err
+		}
```
shizhMSFT, line 183: Based on the log, this code block can be removed.

shizhMSFT (commented):
Build check failed. Have you updated the `go.mod`?

cmd/oras/root/attach.go
yizha1: What does ignorable error mean? Is it a warning or error?

cmd/oras/root/attach.go
qweeah: Should user see this log without `--verbose`?

 - yes => it's `Info`
 - no => it's `Warn`

cmd/oras/root/attach.go
qweeah: If this code block is removed, then all non-415 errors occurred during deleting the referrers index will be silently ignored if `--strict` is not set

Is this UX what we want @shizhMSFT?

cmd/oras/root/attach.go
qweeah: > What does ignorable error mean?

Currently it means attach is successful but something went wrong. It should be only a warning if `--strict` is not set.

cmd/oras/root/attach.go
shizhMSFT: Should that the other way around? `--verbose` toggles `Info` level and by default it is `Warn` level.

shizhMSFT: Anyway, this log should be printed even if `--verbose` is not toggled.

cmd/oras/root/attach.go
```diff
@@ -164,7 +169,19 @@ func runAttach(ctx context.Context, opts attachOptions) error {
...
+		logger.Warn("Attach is successful but removal of outdated referrers index from remote registry failed. Garbage collection may be required.")
```
qweeah, line 184: @FeynmanZhou @yizha1 Can you help review the output warning, thanks

TerryHowe:
looks like this needs to be rebased

cmd/oras/root/attach.go
sajayantony: Is this flag passed into the logger or used somewhere? 
Also `--strict` seems more like a type checking kind of usage. 
Make uses `--ignore-errors` which seems more inline with. The issue I see is that `--ignore-errors` needs to be true by default in this case for refererrer delete failure to be ignored, but othere errors might also exists which aren't ignorable. 

That said why do we even need this flag and can we introduce this when there is an ask to block on this error?

qweeah: > Is this flag passed into the logger or used somewhere?

It's not passed to the logger, it will be used to terminate code run if an exceptional error returned from cleaning the old referrer index.

If the index cleanup fails:
- If the return type is deletion not supported, then attaching succeeds
- Otherwise
   - if the `--strict` is set, then attaching fails
   - otherwise attaching succeeds
 
> can we introduce this when there is an ask to block on this error?

@shizhMSFT suggested adding this flag in an offline discussion and @yizha1 @FeynmanZhou agreed on adding it.

cmd/oras/root/attach.go
```diff
@@ -164,7 +167,14 @@ func runAttach(ctx context.Context, opts attachOptions) error {
...
 	if err != nil {
-		return err
+		var re *remote.ReferrersError
+		if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
+			// not referrer index delete error
+			if opts.strict {
+				return err
+			}
+		}
+		logger.Info("Attach is successful but removal of outdated referrers index from remote registry failed. Garbage collection may be required.")
```
FeynmanZhou, line 177: I would suggest revising it to **Attached successfully but the removal of outdated referrers index from the remote registry failed. Garbage collection may be required.**
since [attach is always used as a verb](https://www.dictionary.com/browse/attach).

cmd/oras/root/attach.go
FeynmanZhou: @sajayantony @shizhMSFT @yizha1 @TerryHowe  The scenario of using the `--strict` flag in `oras attach` is when users want to disallow attaching referrers if the Deletion API is disabled by the registry or garbage collection is not supported by the registry. Without this flag, the registry might be filled with lots of outdated referrer index and eventually impact the registry performance. Does it make sense?

The flag name `--strict` might be a little vague. I am thinking if there is a better name to make it much more straightforward.

cmd/oras/root/attach.go
qweeah: Good catch, resolved

cmd/oras/root/attach.go
qweeah: Also `oras cp` need this flag: if referrer API is not supported and the index need to be updated during the copy, consider copying A in the below diagram:

```mermaid
graph TD;
  A--platform b-->B
  A--platform c-->C
  A--platform d-->D
  B--subject-->E
  C--subject-->E
  D--subject-->E
```

If the concurrency is set to 1, the referrer index of E will be updated twice.

sajayantony: I'm still of the opinion that delete failure should be a warning. The concurrency issue seems like something that SDK should fix and not depend on put/delete semantics for consistency at least from the client's perspective. 

I still think we should avoid this flag all together since Manifest delete itself is quite a low level semantic and not something that the user should care about. It's up to the registry operators to GC them since they aren't tagged anymore and if delete is disabled on them. 

If we show a warning  `Delete failed for referrers index: sha256:... . `  that should be good enough for GHCR users or DH users to follow up with clean up if they can or let it dangle and be purged by the registry. 

> Without this flag, the registry might be filled with lots of outdated referrer index and eventually impact the registry performance. Does it make sense?
This is true even with the flag is my understanding and hence with or without the flag the outcome seems the same. The impact is that whenflag is passed we skip the delete manifest API otherwise its a warning. If we truely need a flag we should wait is my take.

qweeah: > If we show a warning something like WARNING: Delete failed for referrers index: sha256:... .  that should be good enough for GHCR users or DH users to follow up with clean up if they can or let it dangle and be purged by the registry.

@sajayantony I agree that we need to show the digest of referrers index. Will change accordingly.

yizha1: > I'm still of the opinion that delete failure should be a warning. The concurrency issue seems like something that SDK should fix and not depend on put/delete semantics for consistency at least from the client's perspective.
> 
> I still think we should avoid this flag all together since Manifest delete itself is quite a low level semantic and not something that the user should care about. It's up to the registry operators to GC them since they aren't tagged anymore and if delete is disabled on them.
> 
> If we show a warning something like `WARNING: Delete failed for referrers index: sha256:... . ` that should be good enough for GHCR users or DH users to follow up with clean up if they can or let it dangle and be purged by the registry.
> 
> > Without this flag, the registry might be filled with lots of outdated referrer index and eventually impact the registry performance. Does it make sense?
> 
> This is true even with the flag is my understanding and hence with or without the flag the outcome seems the same. The impact is that when the flag is passed we skip the delete manifest API otherwise it's a warning. If we truly need a flag we should wait is my take. Introducing this and removing this is not a good progression.

+1

FeynmanZhou: Agree with @sajayantony . We can add this flag until there is a real user who requests this advanced feature in the future. From the ORAS client side, reminding users in a warning log output about GC is enough at this moment.

cmd/oras/internal/errors/errors.go
```diff
@@ -30,3 +33,13 @@ func NewErrInvalidReference(ref registry.Reference) error {
...
 	return fmt.Errorf("%s: invalid image reference, expecting <name:tag|name@digest>", ref)
 }
+
+// IsReferrersIndexDelete checks if err is a referrers index delete error.
+func IsReferrersIndexDelete(err error, logger logrus.FieldLogger, path string) bool {
+	var re *remote.ReferrersError
+	if !errors.As(err, &re) || !re.IsReferrersIndexDelete() {
+		return false
+	}
+	logger.Info("Failed to remove the outdated referrers index: %s@%s. Garbage collection may be required.", path, re.Subject.Digest)
```
FeynmanZhou, line 43: I would suggest printing the status of `oras push/copy/attach` at the beginning if the new referrers are successfully pushed/copied/attached to the registry. How about `Operation succeed but failed to remove the outdated referrers index: %s@%s. Garbage collection may be required.?

qweeah: changed as suggested

FeynmanZhou (approved):
LGTM

qweeah:
Making this PR draft, since `oras attach` depends on `oras.Copy`, which will return empty descriptor if any non-nil error returned. Even though `oras attach` can ignore non-fatal errors like dangling referrers index deletion error, the descriptor is still empty and there is no way to get the digest of attached artifact.

sajayantony:
Can we address the code coverage drop as well. Also feel free to split up the PR but link to the same issue if it helps scope the changes.

qweeah:
> Can we address the code coverage drop as well.

Yes, that what I am trying to do, I am writing unit test and E2E test and found that the currently implementation breaks `oras attach` as well.

> Also feel free to split up the PR but link to the same issue if it helps scope the changes.

Thanks for the suggestion, will do.

qweeah:
Had a discussion offline with @FeynmanZhou @shizhMSFT and @Wwwsylvia. It's impossible for `oras attach` and `oras cp` to skip a referrers index deletion error without breaking current UX. Requiring SDK-side change, see https://github.com/oras-project/oras-go/issues/510. Closing this PR and will add another one once it's unblocked.

FeynmanZhou:
@qweeah Thanks for summarizing it. Let's create another issue in ORAS CLI repo to track the plan to change the default behavior to not delete dangling referrers index deletion and add a new flag `--skip-gc` to all commands related to referrers change. We need to discuss this change in the next ORAS community meeting.

sajayantony:
+1 on another issue. Would like to discuss the flag name separately.

qweeah:
> +1 on another issue. Would like to discuss the flag name separately.

https://github.com/oras-project/oras/issues/954 created one

PR #924, opened 2023-04-11
Add make target for e2e

sajayantony:
Make doesn't have e2e. Potentially update the gitworkflow also use the make target instead of the run bash. 

- [X]  Does the affected code have corresponding tests, e.g. unit test, E2E test?
- [X]  Does this change require a documentation update?
- [X]  Does this introduce breaking changes that would require an announcement or bumping the major version?
- [X]  Do all new files have an appropriate license header?


TerryHowe (approved):
LGTM

qweeah (approved):
LGTM

qweeah:
Just notice that we need to move e2e coverage generation out of the script `e2e.sh`, otherwise the E2E call stack would be Makefile(teste2e) -> E2E tests - > Makefile(e2e-covdata), which is weird. I can take on that.

PR #957, opened 2023-05-29
feat: add flag to skip deleting obsolete referrers index

### LINKED ISSUE #915: Improve the error msg for OCI v1.0 registries or registries that don't support deletion API

### LINKED ISSUE #952: make `oras cp` skip dangling referrers index deletion error

1.0.0

`oras cp` is by-design to stop on the first-met error. if the copy destination is an OCI-1.0-compliant registry with deletion disabled, then `oras cp` operation might be done half-way without copying all the nodes.

Consider copying the below artifacts via index `A` to such a registry(OCI 1.0+delete disabled) with concurrency set to 1.
```mermaid
graph TD;
  A--platform b-->B
  A--platform c-->C
  A--platform d-->D
  B--subject-->E
  C--subject-->E
  D--subject-->E
```
The copy process will cover all referrers of `E`, since the concurrency is set to 1, those three referrers, `B`, `C` and `D` will be copied one by one. Let's assume the sequence is `B` -> `C` -> `D`:
1) When copying `B`, the referrer index is created with one descriptor pointing to `B`
2) When copying `C`, the referrer index created in step 1) is fetched and append another descriptor of `B`. The referrer index created in step 1) is dangling and need to be cleaned up
3) clean up will fail and copy terminated

In this case, `D` is not covered and the copy process stops.

Copying progress should not be terminated when such failure encountered, it is not terminal and user can clean up the dangling referrer index after the cp process.

### LINKED ISSUE #954: allow oras command to skip referrer index clean up

1.0.0

To all oras commands causing referrers update, a new flag `--gc` should be added to skip deleting the dangling referrers index:
- By default, `--gc` is not set, dangling referrers index will not be clean
- If `--gc` is set, clean dangling referrers index when needed

This flag should be added to below commands
- oras attach
- oras cp
- oras manifest push
- oras manifest delete

Referrer index is the alternative that OCI-1.0 compliant registries can utilize to support artifact reference types, see [OCI spec 1.0-rc.2](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc2/spec.md#unavailable-referrers-api).

If a referrers index `i` exists for a certain subject artifact `a`. When adding or removing referrers for `a`, content of `i` will be updated and a new referrers index will be created and retagged following the [tag schema](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc2/spec.md#referrers-tag-schema). The index `i` contains obsolete content without any tag referenced. 

When running commands like `oras attach` and `oras cp`, it's possible that the used token scope doesn't cover delete, or the registry might disable deletion, so it's better to provide oras an option to skip cleaning the dangling referrers index.
 
Also checked OCI spec, cleanup is optional for registry clients.

qweeah:
This PR adds a flag to `oras attach`, `oras cp` and `oras manifest push`. The flag can be used to specify whether to clean up old referrers index when referrers API is not available.

There is not a breaking change and by default, mentioned commands will still do referrers cleanup.

Depends on merging of #967.

Fixes #954, fixes #952, fixes #915


cmd/oras/internal/option/referrers.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// Referrers option struct.
+type Referrers struct {
+	SkipDeleteReferrers bool
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *Referrers) ApplyFlags(fs *pflag.FlagSet) {
+	fs.BoolVarP(&opts.SkipDeleteReferrers, "skip-delete-referrers", "", false, "skip deleting old referrers index, only work on registry when referrers API is not supported")
```
Wwwsylvia, line 32: Question: So ORAS will still delete the dangling referrers index by default as before? Is it a breaking change? My understanding was that we would skip the deletion by default and would have something like `--gc` to enable deletion?

qweeah: >  So ORAS will still delete the dangling referrers index by default as before?

Yes. By default oras **will** delete the dangling referrers index as before.

cmd/oras/internal/option/referrers.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// SetReferrersGC sets the referrers GC option for the passed-in target.
+func (opts *Referrers) SetReferrersGC(target any, logger logrus.FieldLogger) {
+	if repo, ok := target.(*remote.Repository); ok {
+		repo.SkipReferrersGC = opts.SkipDeleteReferrers
+	} else if !opts.SkipDeleteReferrers {
+		logger.Warnln("referrers deletion can only be enforced upon registry")
+	}
```
shizhMSFT, line 40: In the scenario of oci layout folder, it seems that this warning is always printed unexpectedly.

shizhMSFT: Is that `opts.SkipDeleteReferrers` instead of `!opts.SkipDeleteReferrers`?

shizhMSFT (commented):
What about the manifest delete path?

cmd/oras/internal/option/referrers.go
qweeah: Yes you are right, otherwise by default it will always be printed out.

qweeah:
> What about the manifest delete path?

Referrers index deletion happens after manifest deletion right? If the registry already supports deletion and the token used covers manifest deletion, why would referrers index fail?

shizhMSFT:
> Referrers index deletion happens after manifest deletion right? If the registry already supports deletion and the token used covers manifest deletion, why would referrers index fail?

LoL. Makes sense.

cmd/oras/internal/option/referrers.go
```diff
@@ -0,0 +1,43 @@
...
+	fs.BoolVarP(&opts.SkipDeleteReferrers, "skip-delete-referrers", "", false, "skip deleting old referrers index, only work on registry when referrers API is not supported")
+}
+
+// SetReferrersGC sets the referrers GC option for the passed-in target.
+func (opts *Referrers) SetReferrersGC(target any, logger logrus.FieldLogger) {
+	if repo, ok := target.(*remote.Repository); ok {
+		repo.SkipReferrersGC = opts.SkipDeleteReferrers
+	} else if opts.SkipDeleteReferrers {
+		// not a registry, can't skip referrers deletion
+		// TODO: change default value in 2.0
```
shizhMSFT, line 40: Let's remove this comment. Instead, we should create an issue to discuss this. We probably don't want to change the default value but change the flag name.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #926, opened 2023-04-13
test: make e2e coverage data in action

qweeah:
**What this PR does / why we need it**:
As a follow-up of #924, this PR moves the action of coverage data generation out of e2e run script.
Also added some e2e specs to validate the changes.

**Please check the following list**:
- [ ]  Does the affected code have corresponding tests, e.g. unit test, E2E test?
- [ ]  Does this change require a documentation update?
- [ ]  Does this introduce breaking changes that would require an announcement or bumping the major version?
- [ ]  Do all new files have an appropriate license header?

PR #967, opened 2023-06-06
chore: simplify e2e scripts and github action

qweeah:
As follow-up of #926, this PR 
1) uses `make` to run e2e in Github actions.
2) simplifies e2e coverage generation process with unnecessary env variables removed.


TerryHowe (approved):
/lgtm

PR #979, opened 2023-06-20
chore: improve error message of path traversing

### LINKED ISSUE #978: improve path traversal error msg of `oras pull`
If path of artifact blobs are absolute and needs to be pulled outside of current working directory, by default `oras pull` will be terminated with error like `path traversal disallowed`

Need to advise user action for this specific error, like `Use pull with option -T/allow-path-traversal to allow path traversal`

> Can this be added to the error message when the pull fails with "Path travel disallowed". Use pull with option -T or --allow-path-traversal to allow path traversal.

_Originally posted by @suganyas in https://github.com/oras-project/oras/issues/973#issuecomment-1596433851_

qweeah:
This PR provides detailed instruction to user on how to handle path traversing error during pulling.

Fixes #978 

### DIFF
--- a/cmd/oras/root/pull.go
+++ b/cmd/oras/root/pull.go
@@ -26,7 +26,6 @@ import (
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/content/file"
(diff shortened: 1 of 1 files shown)


qweeah:
Making this a draft since #980 might be a better option.

Issue #983, opened 2023-06-21
`--disable-path-validation` not working for `oras push` and `oras attach`


qweeah:
### What happened in your environment?

I can still push files with absolute path even if `--disable-path-validation` is set to false
```console
$ oras push localhost:5000/test:path /my/test/file --disable-path-validation=false
Uploading e3b0c44298fc /my/test/file
Uploaded  e3b0c44298fc /my/test/file
Pushed [registry] localhost:5000/test:path
Digest: sha256:c925e33e51a8ff03e378e2956fdd4aa3a28203ed9aaa778d634c5e647c65bc54
```

### What did you expect to happen?

The operation should fail with warning mentioning that using absolute paths in artifacts is generally insecure and not friendly to cross-platform scenarios.

### How can we reproduce it?

Use absolute path to specify files in `oras push` or `oras attach`

### What is the version of your ORAS CLI?

1.0.0

### What is your OS environment?

20.04

### Are you willing to submit PRs to fix it?

- [ ] Yes, I am willing to fix it.

qweeah:
@suganyas Do you want to work on this?

suganyas:
@qweeah my fix will fix this one too.

qweeah:
completed in https://github.com/oras-project/oras/pull/988

Issue #980, opened 2023-06-20
disallow absolute file path in `oras push` and `oras attach` by default 


qweeah:
### What is the version of your ORAS CLI

oras 1.0.0

### What would you like to be added?

By default, disable pushing files with absolute file path in `oras push` and `oras attach`, add a new flag, like `--allow-absolute-path` to support glass-breaking scenarios.

The artifact packed via `oras attach/push --allow-absolute-path` can only be pulled via `oras pull --T/allow-path-traversal`

### Why is this needed for ORAS?

Absolutely pathed layers are dangerous, since 
1) File might be written into any folder outside of the working directory.
2) The allowed characters in files names are different between Linux and Windows. Artifacts packed in one platform may not be pulled in another platform.

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

qweeah:
@suganyas I am not sure packaging an absolutely pathed artifact is intended in your case. If it's not, maybe oras should guide 有our users not to produce those artifacts in 2.0?

suganyas:
Hi @qweeah thanks for looking into it .  I just tried to push an artifact file from linux environment from a absolute path or from different directory.  The path is implicitly taken by the oras cli or oras sdk when I pushed it. Like the file was in a directory /home/vts/1/a.exe. and I pushed from /home/test. I am ok if the push fails stating me that the file is not in the current directory and for security reasons you have to be in same working directory. But the push passes and pull fails . So disallow is a much better option. I am sure when I used ORAS python SDK it did fail with error stating as here https://github.com/oras-project/oras-py/blob/209c9b98043a00d1b04789cc2967ca7021dc5b2e/oras/provider.py#L651 . The CLI should have same behaviour as SDK. Then push and pull are coherent. It can be a bad experience when push is ok and pull fails and when different people do it can be hard for people to understand why it fails

suganyas:
@qweeah I am happy to contribute or make changes according to our decision

qweeah:
@suganyas My idea is that it's better to prevent user from using absolute path. @FeynmanZhou @shizhMSFT @TerryHowe what do you think?

P.S. This issue introduces breaking change and need to be enabled for ORAS 2.0 release.

FeynmanZhou:
This will change the default behavior of `oras push` and `oras attach` so I agree that it should be planned in ORAS v2.0.0.

> My idea is that it's better to prevent user from using absolute path.
[...]

PR #988, opened 2023-06-26
fix: enforce path validation for push/attach and improve path traversal failure message for pull

suganyas:
 I just tried to push an artifact file from linux environment from a absolute path or from different directory. The path is implicitly taken by the oras cli or oras sdk when I pushed it. Like the file was in a directory /home/vts/1/a.exe. and I pushed from /home/test. I am ok if the push fails stating me that the file is not in the current directory and for security reasons you have to be in same working directory. But the push passes and pull fails . So disallow is a much better option. I am sure when I used ORAS python SDK it did fail with error stating as here https://github.com/oras-project/oras-py/blob/209c9b98043a00d1b04789cc2967ca7021dc5b2e/oras/provider.py#L651 . The CLI should have same behaviour as SDK. Then push and pull are coherent. It can be a bad experience when push is ok and pull fails and when different people do it can be hard for people to understand why it fails and also cross platform can fail if it is not intentional

Fixes #
#980 
#983 
#978 

Screenshot showing the scenarios tested

Screenshot 2023-06-26 at 10 13 47 pm

### DIFF
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -86,7 +86,7 @@ func (opts *Packer) Parse() error {
 			}
 		}
 		if len(failedPaths) > 0 {
-			return fmt.Errorf("%v: %v currentDir :%v", errPathValidation, strings.Join(failedPaths, ", "), currentDir)
+			return fmt.Errorf("%w: %v", errPathValidation, strings.Join(failedPaths, ", "))
 		}
 	}
 	return nil
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -38,6 +39,7 @@ var (
 	errAnnotationConflict    = errors.New("`--annotation` and `--annotation-file` cannot be both specified")
 	errAnnotationFormat      = errors.New("missing key in `--annotation` flag")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
+	errPathValidation        = errors.New("one or more files are not in the current directory.If it's intentional use --disable-path-validation flag to skip this check")
 )
 
 // Packer option struct.
@@ -68,6 +70,35 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 		return err
 	}
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
+}
+func (opts *Packer) Parse() error {
+	currentDir, err := os.Getwd()
+	var failedPaths []string
+	if err != nil {
+		return err
+	}
+	if !opts.PathValidationDisabled && len(opts.FileRefs) != 0 {
+		for _, path := range opts.FileRefs {
+			//Remove the type if specified in the path <file>[:<type>] format
+			lastIndex := strings.LastIndex(path, ":")
+			if lastIndex != -1 {
+				path = path[:lastIndex]
+			}
+			absPath, err := filepath.Abs(path)
+			dirPath := filepath.Dir(absPath)
+			if err != nil {
+				return err
+			}
+			if dirPath != currentDir {
+				failedPaths = append(failedPaths, absPath)
+			}
+		}
+		if len(failedPaths) > 0 {
+			errorMsg := fmt.Sprintf("%v: %v currentDir :%v", errPathValidation, strings.Join(failedPaths, ", "), currentDir)
+			return errors.New(errorMsg)
+		}
+	}
+	return nil
 }
 
 // LoadManifestAnnotations loads the manifest annotation map.
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -40,7 +40,7 @@ var (
 	errAnnotationConflict    = errors.New("`--annotation` and `--annotation-file` cannot be both specified")
 	errAnnotationFormat      = errors.New("missing key in `--annotation` flag")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
-	errPathValidation        = errors.New("one or more files are not in the current directory.If it's intentional use --disable-path-validation flag to skip this check")
+	errPathValidation        = errors.New("absolute file path detected. If it's intentional use --disable-path-validation flag to skip this check")
 )
 
 // Packer option struct.
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -40,7 +40,7 @@ var (
 	errAnnotationConflict    = errors.New("`--annotation` and `--annotation-file` cannot be both specified")
 	errAnnotationFormat      = errors.New("missing key in `--annotation` flag")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
-	errPathValidation        = errors.New("absolute file path detected. If it's intentional use --disable-path-validation flag to skip this check")
+	errPathValidation        = errors.New("absolute file path detected. If it's intentional, use --disable-path-validation flag to skip this check")
 )
 
 // Packer option struct.
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -72,8 +72,8 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
 func (opts *Packer) Parse() error {
-	currentDir, err := os.Getwd()
 	var failedPaths []string
+	currentDir, err := os.Getwd()
 	if err != nil {
 		return err
 	}
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -74,7 +74,7 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 }
 func (opts *Packer) Parse() error {
 	var failedPaths []string
-	if !opts.PathValidationDisabled && len(opts.FileRefs) != 0 {
+	if !opts.PathValidationDisabled {
 		for _, path := range opts.FileRefs {
 			//Remove the type if specified in the path <file>[:<type>] format
 			path, _, err = fileref.Parse(path, "")
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -73,8 +73,8 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
 func (opts *Packer) Parse() error {
-	var failedPaths []string
 	if !opts.PathValidationDisabled {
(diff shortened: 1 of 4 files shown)


qweeah:
Can you help rename the PR following [the conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#commit-message-for-a-fix-using-an-optional-issue-number)? Thanks

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +71,35 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	}
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
+func (opts *Packer) Parse() error {
+	currentDir, err := os.Getwd()
+	var failedPaths []string
```
qweeah, line 76: nit:
```suggestion
	var failedPaths []string
	currentDir, err := os.Getwd()
```

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +71,35 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
...
+		return err
+	}
+	if !opts.PathValidationDisabled && len(opts.FileRefs) != 0 {
+		for _, path := range opts.FileRefs {
+			//Remove the type if specified in the path <file>[:<type>] format
+			lastIndex := strings.LastIndex(path, ":")
+			if lastIndex != -1 {
+				path = path[:lastIndex]
+			}
+			absPath, err := filepath.Abs(path)
```
qweeah, line 87: Better to use the `Parse` function provided in https://github.com/oras-project/oras/tree/main/cmd/oras/internal/fileref. to get `path` since is provides better cross-platform compability.

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +71,35 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
...
+			dirPath := filepath.Dir(absPath)
+			if err != nil {
+				return err
+			}
+			if dirPath != currentDir {
```
qweeah, line 92: Should be `strings.HasPrefix(dirPath, currentDir)` right?

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +71,35 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
...
+				failedPaths = append(failedPaths, absPath)
+			}
+		}
+		if len(failedPaths) > 0 {
+			errorMsg := fmt.Sprintf("%v: %v currentDir :%v", errPathValidation, strings.Join(failedPaths, ", "), currentDir)
```
qweeah, line 97: If we don't user %w to wrap an `errPathValidation`, we don't need to declare that err type. We can merge the content of `errPathValidation` into this Sprintf.

cmd/oras/root/pull.go
```diff
@@ -237,6 +239,10 @@ func runPull(ctx context.Context, opts pullOptions) error {
 	// Copy
 	desc, err := oras.Copy(ctx, src, opts.Reference, dst, opts.Reference, copyOptions)
 	if err != nil {
+		if strings.Contains(err.Error(), "path traversal disallowed") {
```
qweeah, line 242: This should not be done via string comparing, oras-go exported a specific type `ErrPathTraversalDisallowed`, see https://pkg.go.dev/oras.land/oras-go/v2@v2.2.0/content/file#pkg-variables

cmd/oras/root/pull.go
```diff
@@ -237,6 +239,10 @@ func runPull(ctx context.Context, opts pullOptions) error {
...
+			errorMsg := fmt.Sprintf("%v: %v ", err, "To enable path traversal use --allow-path-traversal flag")
```
qweeah, line 243: nit: should use `%w` to wrap the error like
```suggestion
			errorMsg := fmt.Sprintf("%v: %w ", "to enable path traversal use --allow-path-traversal flag", err)
```

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +71,35 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
...
+			return errors.New(errorMsg)
```
qweeah, line 98:
```suggestion
			return fmt.Errorf("%v: %v currentDir :%v", errPathValidation, strings.Join(failedPaths, ", "), currentDir)
```

qweeah (changes requested):

cmd/oras/internal/option/packer.go
suganyas: Addressed and checked for absolute path

cmd/oras/root/pull.go
suganyas: Fixed the issue as errors.is

cmd/oras/root/pull.go
```diff
@@ -237,6 +238,9 @@ func runPull(ctx context.Context, opts pullOptions) error {
 	// Copy
 	desc, err := oras.Copy(ctx, src, opts.Reference, dst, opts.Reference, copyOptions)
 	if err != nil {
+		if errors.Is(err, file.ErrPathTraversalDisallowed) {
+			err = fmt.Errorf("%s: %w", "use option -T/allow-path-traversal to allow pulling outside of working directory", err)
```
qweeah, line 242: minor suggestion: option is an dev term, flag is better.
```suggestion
			err = fmt.Errorf("%s: %w", "use flag -T/allow-path-traversal to allow pulling files outside of working directory", err)
```

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +72,29 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	}
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
+func (opts *Packer) Parse() error {
+	var failedPaths []string
+	currentDir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	if !opts.PathValidationDisabled && len(opts.FileRefs) != 0 {
```
qweeah, line 81: no need to check length of input file references
```suggestion
	if !opts.PathValidationDisabled {
```

cmd/oras/internal/option/packer.go
```diff
@@ -38,6 +40,7 @@ var (
 	errAnnotationConflict    = errors.New("`--annotation` and `--annotation-file` cannot be both specified")
 	errAnnotationFormat      = errors.New("missing key in `--annotation` flag")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
+	errPathValidation        = errors.New("one or more files are not in the current directory.If it's intentional use --disable-path-validation flag to skip this check")
```
qweeah, line 43:
```suggestion
	errPathValidation        = errors.New("absolute file path detected. If it's intentional use --disable-path-validation flag to skip this check")
```

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +72,29 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
...
+			path, _, err = fileref.Parse(path, "")
+			if err != nil {
+				return err
+			}
+			if filepath.IsAbs(path) {
+				failedPaths = append(failedPaths, path)
+			}
+		}
+		if len(failedPaths) > 0 {
+			return fmt.Errorf("%v: %v currentDir :%v", errPathValidation, strings.Join(failedPaths, ", "), currentDir)
```
qweeah, line 93: Since we loosen the check to disallow absolute path, there is no need to output current path
```suggestion
			return fmt.Errorf("%w: %v", errPathValidation, strings.Join(failedPaths, ", "))
```

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +72,29 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	}
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
+func (opts *Packer) Parse() error {
+	var failedPaths []string
+	currentDir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
```
qweeah, line 80:
```suggestion
```

qweeah: There is no need to check current path any more

qweeah (changes requested):

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +72,25 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	}
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
+func (opts *Packer) Parse() error {
+	var failedPaths []string
+	if !opts.PathValidationDisabled {
+		for _, path := range opts.FileRefs {
+			//Remove the type if specified in the path <file>[:<type>] format
```
qweeah, line 79: nit
```suggestion
			// Remove the type if specified in the path <file>[:<type>] format
```

qweeah (changes requested):

qweeah (approved):
LGTM

cmd/oras/internal/option/packer.go
```diff
@@ -38,6 +40,7 @@ var (
 	errAnnotationConflict    = errors.New("`--annotation` and `--annotation-file` cannot be both specified")
 	errAnnotationFormat      = errors.New("missing key in `--annotation` flag")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
+	errPathValidation        = errors.New("absolute file path detected. If it's intentional use --disable-path-validation flag to skip this check")
```
shizhMSFT, line 43: nit: add a comma for better readability.
```suggestion
	errPathValidation        = errors.New("absolute file path detected. If it's intentional, use --disable-path-validation flag to skip this check")
```

cmd/oras/root/pull.go
```diff
@@ -237,6 +238,9 @@ func runPull(ctx context.Context, opts pullOptions) error {
 	// Copy
 	desc, err := oras.Copy(ctx, src, opts.Reference, dst, opts.Reference, copyOptions)
 	if err != nil {
+		if errors.Is(err, file.ErrPathTraversalDisallowed) {
+			err = fmt.Errorf("%s: %w", "use flag -T/allow-path-traversal to allow pulling files outside of working directory", err)
```
shizhMSFT, line 242: We need to point out that `--allow-path-traversal` is a flag for insecure operations. Meanwhile, mentioning the full flag name is good enough.
```suggestion
			err = fmt.Errorf("%s: %w", "use flag --allow-path-traversal to allow insecurely pulling files outside of working directory", err)
```

cmd/oras/internal/option/packer.go
```diff
@@ -69,6 +72,25 @@ func (opts *Packer) ExportManifest(ctx context.Context, fetcher content.Fetcher,
 	}
 	return os.WriteFile(opts.ManifestExportPath, manifestBytes, 0666)
 }
+func (opts *Packer) Parse() error {
+	var failedPaths []string
+	if !opts.PathValidationDisabled {
```
shizhMSFT, line 77: nit: put `failedPaths` in the proper scope.
```suggestion
	if !opts.PathValidationDisabled {
		var failedPaths []string
```

shizhMSFT (approved):
LGTM

qweeah (approved):
LGTM

PR #991, opened 2023-06-27
test(e2e): add test for pushing via absolute path

qweeah:
This PR adds E2E tests that validate the absolute path checking for oras push and oras attach. Depends on merging of #988


shizhMSFT (approved):
LGTM

Issue #1002, opened 2023-07-05
Vote for releasing v1.1.0-rc.1


qweeah:
At least 3 approvals are needed from the 5 owners to cut https://github.com/oras-project/oras/commit/7fc68d4b18ab6d86a2ee4b596dcbdf2ad4948091 as the release branch for oras CLI v1.1.0 release. This branch will be named as `release-1.1` and tagged as `v1.1.0-rc.1` for releasing.

 - [x] Terry Howe (@TerryHowe)
 - [x] Sajay Antony (@sajayantony)
 - [x] Shiwei Zhang (@shizhMSFT)
 - [x] Steve Lasker (@SteveLasker)
 - [x] Billy Zha (@qweeah)

The release will be conducted by following the [release checklist](https://oras.land/docs/cli/developer_guide#release-checklist).

### Bug Fixes
- `oras push` and `oras attach` generates unwanted manifest file (#995)
- Path validation not enforced for `oras attach` and `oras push` (#983, #980, #973)
- Wrong error information when `--distribution-spec` is provided with invalid value (#897)  
- Tree view print bug in `oras discover` (#1003)

### New Features
- Add a new flag `--skip-delete-referrers` to allow oras commands to skip referrer index clean up (#954, #952, #915)
- Support Golang scripting (#807)

### Other Changes
- Improve UX
  - `oras pull` prints better error msg when path traversal is required (#978)
  - Mention OCI support in help doc (#904)
  - Mark filtered tag listing as experimental (#993)
- Improve Documentation
  - Fix link in contribute doc (#950)
- Update dependencies
  - Remove dependencies on docker CLI (#414)
  - Update dependency from `github.com/moby/term` to `golang.org/x/term` (#908)
- Update to golang `1.20.5`
- Improve tests
  - Generate code coverage for E2E tests (#900)
  - Added more tests (#904, #901)

Comparing to [v1.0.0](https://github.com/oras-project/oras/tree/v1.0.0), the release includes below code changes:
## What's Changed
* refactor: Logger initializes logging by @TerryHowe in https://github.com/oras-project/oras/pull/867
* test(cov): make unit tests cover more by @qweeah in https://github.com/oras-project/oras/pull/905
* refactor: add root package for Cobra commands by @qweeah in https://github.com/oras-project/oras/pull/825
* build(deps): bump github.com/docker/cli from 23.0.1+incompatible to 23.0.2+incompatible by @dependabot in https://github.com/oras-project/oras/pull/909
[...]

PR #1010, opened 2023-07-06
Release v1.1.0 rc1

sajayantony:
Draft PR to show version bumps so the release can be PR driven. 

The proposal here is to tag 32bfb875d3e288f0c5133b81af2c742d257cbc9b as v1.1.0-rc1

Basically move voting to the PR rather than an issue - https://github.com/oras-project/oras/issues/1002

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -17,7 +17,7 @@ package version
 
 var (
 	// Version is the current version of the oras.
-	Version = "1.1.0-rc.1"
+	Version = "1.1.0-rc1"
 	// BuildMetadata is the extra build time data
 	BuildMetadata = "unreleased"
 	// GitCommit is the git sha1
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -17,7 +17,7 @@ package version
 
 var (
 	// Version is the current version of the oras.
-	Version = "1.1.0-rc1"
+	Version = "1.1.0-rc1+dev"
 	// BuildMetadata is the extra build time data
 	BuildMetadata = "unreleased"
 	// GitCommit is the git sha1


qweeah:
One question on this: if this PR is merged, then a new commit will be added. We should tag the new commit for releasing. https://github.com/oras-project/oras/commit/32bfb875d3e288f0c5133b81af2c742d257cbc9b will not be the target commit.

qweeah:
ORAS applies the practice of linear commit history, which means a PR will always be merged via squashing or rebasing. This is different from how the version control is done in image because when a PR is merged, the commit digest might be recalculated and changed. So, in ORAS CLI repo we cannot use PR voting facilities to vote for releasing.

internal/version/version.go
```diff
@@ -17,7 +17,7 @@ package version
 
 var (
 	// Version is the current version of the oras.
-	Version = "1.1.0-rc.1"
+	Version = "1.1.0-rc1+dev"
 	// BuildMetadata is the extra build time data
 	BuildMetadata = "unreleased"
```
qweeah, line 22: @sajayantony  We already have extra build time data here, the full version is `<Version>+<BuildMetadata>`
e.g. 
- if you build from this commit, the full version string will be `1.1.0-rc1+dev+unreleased`. 
- Without this code change in this PR, the full version string will be `1.1.0-rc.1+unreleased`, the latter already implies the former with less duplicated info.

shizhMSFT: Definition quotes from [SemVer 2.0.0](https://semver.org/#semantic-versioning-specification-semver):

> 10. Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3----117B344092BD.

sajayantony:
> One question on this: if this PR is merged, then a new commit will be added. We should tag the new commit for releasing. [32bfb87](https://github.com/oras-project/oras/commit/32bfb875d3e288f0c5133b81af2c742d257cbc9b) will not be the target commit.

This is just to show how to get the approval on the PR but keep the code moving forward. 
for e.g. the code should probably now move to RC2+ something. This should be merged and not squashed. Again this is not blocking for the current release but I would like to update the release process to PR based either way. 
Maybe for the next RC and also update the release docs.

internal/version/version.go
sajayantony: @shizhMSFT should we up the version then?

qweeah:
I think there is a gap. You may have observed that the target commit digest in the[ voting issue for 1.1.0-rc.1](https://github.com/oras-project/oras/issues/1002) has been changed after I called for voting. The back-and-forth change happens not because of how we perform release branch cut, but because **we want to bump oras-go include two bug fixes into the release branch after call for vote**. If we use a vote PR for the cut and want to include new changes after raising the vote PR, we still need to rebase and update the target commit digest and call for vote again.

Also, current way of releasing won't block coding. New changes can be merged into the main branch. But if we use a vote PR for release branch cut, other PRs cannot be merged before the vote PR is merged.

internal/version/version.go
qweeah: The version is already updated in https://github.com/oras-project/oras/commit/42650e5778283be7e4926f123f183db12dfbacaa before the branch cut. @sajayantony Are you suggesting we should do the version update for the next version? We don't have a procedure for that and might be worth doing after releasing.

TerryHowe:
I thought the plan was to use the `rc.1`  version. Close?

internal/version/version.go
shizhMSFT: Let's reset the releasing pipeline of `v1.1.0-rc.1`.

shizhMSFT: See details in #1013

qweeah:
Had a discussion offline with Shiwei and we decided to try this release method. I will raise another PR to revert version update and another for to vote for branch cutting.

qweeah:
closing as it's already experimented in #1016

PR #1031, opened 2023-07-21
fix: move stale into .github/workflows

sajayantony:
Fixing the location of the stale workflow. Also we have an option to run this only during "Mon .. Thur" 
Once we have this running we can tweak this more.


TerryHowe (approved):
/lgtm

PR #1036, opened 2023-07-24
chore: exempt stale if milestone is set

qweeah:
This PR adds a workflow change which will avoid marking PRs or issues as stale if any milestone is tagged.


TerryHowe (approved):
/lgtm

sajayantony:
Thanks for the change @qweeah - This also means we should review the milestones regularly so that issues don't get stale in there. Honestly I'm happy that a lot of old issues got bumped and makes the maintainers consider if we actually will ever get to it. 

Future seems like a backlog milestone, should we avoid moving everything to future unless there is plan and/or close them off since there is no immediate plan would be a question for us to agree on.

PR #1040, opened 2023-07-31
adding identity token as an approximate alias of password

### LINKED ISSUE #742: Make `--identity-token` an alias of `--password`

v0.16.0

Make `--identity-token` an alias of `--password`

`oras` supports using identity token as the credential without a username.

The UX of `oras login $REGISTRY --password $TOKEN` does not seem good enough. It will be better if it is
```sh
oras login $REGISTRY --identity-token $TOKEN
```

sbdtu5498:
These changes make the identity token an approximate alias of a password. The underlying implementation still depends on the password implementation, but a few cases to handle exceptions and errors have been added, thus technically speaking, it is not an alias but can be considered an approximate alias to enhance UX in terms of user's understandability.

**Which issue(s) this PR fixes** 
Fixes #742 


cmd/oras/internal/option/remote.go
```diff
@@ -77,14 +78,15 @@ func applyPrefix(prefix, description string) (flagPrefix, notePrefix string) {
 // Commonly used for non-unary remote targets.
 func (opts *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description string) {
 	var (
-		shortUser     string
-		shortPassword string
-		shortHeader   string
-		flagPrefix    string
-		notePrefix    string
```
TerryHowe, line 84: Why are these called `short*` ? Unrelated to this change

cmd/oras/internal/option/remote.go
```diff
@@ -113,10 +116,16 @@ func (opts *Remote) Parse() error {
 	return opts.distributionSpec.Parse()
 }
 
-// readPassword tries to read password with optional cmd prompt.
+// readPassword tries to read password and identity-token with optional cmd prompt.
 func (opts *Remote) readPassword() (err error) {
```
TerryHowe, line 120: I'm fine with this, but there is a Cobra feature for mutually exclusive

sbdtu5498: Oh, thanks for the info. I will check it and get back on this one.

cmd/oras/internal/option/remote.go
qweeah: @sbdtu5498 is right. If `prefix` is not empty, short notation will not be enabled.

cmd/oras/internal/option/remote.go
```diff
@@ -77,14 +78,15 @@ func applyPrefix(prefix, description string) (flagPrefix, notePrefix string) {
...
 func (opts *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description string) {
 	var (
-		shortUser     string
-		shortPassword string
-		shortHeader   string
-		flagPrefix    string
-		notePrefix    string
+		shortUser          string
+		shortPassword      string
+		shortIdentityToken string
```
qweeah, line 83: Should not enable shorthand for `--identity-token`

cmd/oras/internal/option/remote.go
```diff
@@ -77,14 +78,15 @@ func applyPrefix(prefix, description string) (flagPrefix, notePrefix string) {
...
+		shortHeader        string
+		flagPrefix         string
+		notePrefix         string
 	)
 	if prefix == "" {
-		shortUser, shortPassword = "u", "p"
+		shortUser, shortPassword, shortIdentityToken = "u", "p", "i"
```
qweeah, line 89: Should not enable shorthand for `--identity-token`

cmd/oras/internal/option/remote.go
```diff
@@ -94,6 +96,7 @@ func (opts *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
 	}
 	fs.StringVarP(&opts.Username, flagPrefix+"username", shortUser, "", notePrefix+"registry username")
 	fs.StringVarP(&opts.Password, flagPrefix+"password", shortPassword, "", notePrefix+"registry password or identity token")
+	fs.StringVarP(&opts.IdentityToken, flagPrefix+"identity-token", shortIdentityToken, "", notePrefix+"identity token for registry")
```
qweeah, line 99: Should not enable shorthand for `--identity-token`
```suggestion
	fs.StringVarP(&opts.IdentityToken, flagPrefix+"identity-token", "", "", notePrefix+"identity token for registry")
```

cmd/oras/internal/option/remote.go
```diff
@@ -113,10 +116,16 @@ func (opts *Remote) Parse() error {
 	return opts.distributionSpec.Parse()
 }
 
-// readPassword tries to read password with optional cmd prompt.
+// readPassword tries to read password and identity-token with optional cmd prompt.
```
qweeah, line 119: NIT: make line length less than 80
```suggestion
// readPassword tries to read password and identity-token with optional cmd
// prompt.
```

cmd/oras/internal/option/remote.go
qweeah: Please refer to https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive

qweeah: BTW `identity-token` is exclusive to both `password` and `username`

cmd/oras/internal/option/remote.go
```diff
@@ -47,6 +47,7 @@ type Remote struct {
 	Username          string
 	PasswordFromStdin bool
 	Password          string
+	IdentityToken     string
```
qweeah, line 50: Don't need this new value, the `--registry-token` can be applied to `Password` directly. The tricky part is whether we need  `--registry-token-stdin`? Please wait for https://github.com/oras-project/oras/issues/742#issuecomment-1660256192 to be answered?

qweeah (changes requested):

cmd/oras/internal/option/remote.go
shizhMSFT, line 50: Instead of having `IdentityToken`, we can have something like `Secret` and `SecretFromStdin`.

cmd/oras/internal/option/remote.go
```diff
@@ -244,7 +253,19 @@ func (opts *Remote) parseCustomHeaders() error {
...
+			Username: "",
+			Password: opts.IdentityToken,
+		}
+	} else {
+		// If IdentityToken is not provided, use the username and password as credentials
+		return auth.Credential{
+			Username: opts.Username,
+			Password: opts.Password,
+		}
+	}
```
shizhMSFT, line 268: This block of code should be reverted since `credential.Credential()` already handles everything.

cmd/oras/root/login.go
```diff
@@ -77,25 +80,34 @@ Example - Log in with username and password in an interactive terminal and no TL
...
+	// Check if both '--username' and '--identity-token' are provided
+	if opts.Username != "" && opts.IdentityToken != "" {
+		return fmt.Errorf("both --username and --identity-token cannot be used together")
+	}
+
+	// If IdentityToken is provided, use it as the credential without a username
+	if opts.IdentityToken != "" {
+		opts.Username = ""                 // Set the username to empty since it's not required when using identity token
+		opts.Password = opts.IdentityToken // Use the identity token as the password
+	} else if opts.Password == "" {
```
shizhMSFT, line 92: Those part of code should be reverted if we introduce `opts.Secret`.

cmd/oras/internal/option/remote.go
sbdtu5498: Sure will do it

cmd/oras/internal/option/remote.go
sbdtu5498: I see. I will remove the field and would implement identity-token-stdin instead.

sbdtu5498: Oh, So should I add a field secret that will hold the Identity Token, Instead of Password?

cmd/oras/internal/option/remote.go
sbdtu5498: Okay, will fix it.

cmd/oras/root/login.go
sbdtu5498: Oh Okay. I will revert them. Let me know should I go with Secrets and SecretFromStdin or identity-token-stdin and password.

TerryHowe:
Where is this going? Be nice to get this out.

FeynmanZhou:
@sbdtu5498 Will you be able to continue to work on this PR? It has been stale for a long time and might be closed if no follow-ups on it in a few weeks.

shizhMSFT:
Succeeded and closed by #1294

PR #1054, opened 2023-08-08
feat: pack OCI image spec v1.1 manifests

### LINKED ISSUE #1043: update experimental flag `image-spec`
OCI Artifact Manifest no longer exists in https://github.com/opencontainers/image-spec/tree/v1.1.0-rc4, oras CLI should remove the `v1.1-artifact` as an option of `--image-spec`.

For `oras push`:
 - Remove `v1.0-image` and `v1.1-artifact`
 - Add new options
     - v1.0: build manifest compliant to [image spec v1.0](https://github.com/opencontainers/image-spec/blob/v1.0/manifest.md), which doesn't allow `--config` and `--artifact-type` to be used at the same time
     - v1.1: build manifest compliant to [image spec v1.1.0-rc4](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md)

For `oras attach`:
 - Remove `--image-spec`: `oras attach` only generate manifests with both `subject` and `artifactType` presented

### LINKED ISSUE #1062: Enable artifact guidance for ORAS

v1.1.0-rc1

Continuing the conversation here since its about an ORAS CLI experience discussion.  https://github.com/oras-project/oras-www/pull/248#issuecomment-1674110920 @qweeah 

> @sajayantony To align oras's behavior with this subtree, do we need to ban `--config` if no file specified in below scenario?

I think we can warn since this is guidance and not a MUST in the specification. 
The above example has a config file but no blob and that is not a scenario that cannot be executed but not recommended. 

Also we probably need to consider how to enable the user to pass in the config mediaType and also artifact-type. 
```
❯ oras push --config config.json:application/example.config+json  localhost:5000/oras-test:push1 --artifact-type application/example
Error: --artifact-type and --config cannot both be provided
```

Updated support for v1.1 

qweeah:
Fixes #1043, fixes #1062


internal/graph/graph.go
shizhMSFT, line 1: Will `Index` related changes be updated in another PR in terms of `Subject`?

cmd/oras/root/push.go
```diff
@@ -102,15 +98,16 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
...
+			if err := option.Parse(&opts); err != nil {
+				return err
+			}
+			if opts.ImageSpec.PackType == option.PackManifestTypeImageV1_0 && opts.manifestConfigRef != "" && opts.artifactType != "" {
+				return errors.New("--artifact-type and --config cannot both be provided for 1.0 OCI image")
 			}
-			return option.Parse(&opts)
+			if opts.PackType == oras.PackManifestTypeImageV1_1_0_RC4 && opts.manifestConfigRef == "" && opts.artifactType == "" {
+				opts.artifactType = oras.MediaTypeUnknownArtifact
+			}
```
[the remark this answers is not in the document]
qweeah: I think the green block doesn't provide better readability.
```diff
+                       switch opts.PackType {
+			case option.PackManifestTypeImageV1_0:
+				if opts.manifestConfigRef != "" && opts.artifactType != "" {
+					return errors.New("--artifact-type and --config cannot both be provided for 1.0 OCI image")
+				}
+			case oras.PackManifestTypeImageV1_1_0_RC4:
+				if  opts.manifestConfigRef == "" && opts.artifactType == "" {
+					opts.artifactType = oras.MediaTypeUnknownArtifact
+				}
+			}
-			if opts.ImageSpec.PackType == option.PackManifestTypeImageV1_0 && opts.manifestConfigRef != "" && opts.artifactType != "" {
-				return errors.New("--artifact-type and --config cannot both be provided for 1.0 OCI image")
-			}
-			if opts.PackType == oras.PackManifestTypeImageV1_1_0_RC4 && opts.manifestConfigRef == "" && opts.artifactType == "" {
-				opts.artifactType = oras.MediaTypeUnknownArtifact
			}
```

internal/graph/graph.go
qweeah: Yes, changed accordingly.

qweeah: Also found that `FindReferrerPredecessors` is a duplicated implementation of `Referrers`, removed the function

internal/graph/graph.go
```diff
@@ -83,6 +83,19 @@ func Successors(ctx context.Context, fetcher content.Fetcher, node ocispec.Descr
 		}
 		nodes = manifest.Blobs
 		subject = manifest.Subject
+
```
shizhMSFT, line 86: nit: empty line

cmd/oras/root/push.go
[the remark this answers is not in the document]
shizhMSFT: It does. At least, readers won't be confused with the inconsistency of `opts.ImageSpec.PackType` and `opts.PackType`. Also, readers can focus more on the actual boolean check on the `manifestConfigRef` and `artifactType` without repeat thinking about the pack type.

cmd/oras/root/cp.go
```diff
@@ -117,7 +118,9 @@ func runCopy(ctx context.Context, opts copyOptions) error {
 	committed := &sync.Map{}
 	extendedCopyOptions := oras.DefaultExtendedCopyOptions
 	extendedCopyOptions.Concurrency = opts.concurrency
-	extendedCopyOptions.FindPredecessors = graph.FindReferrerPredecessors
+	extendedCopyOptions.FindPredecessors = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+		return graph.Referrers(ctx, src, desc, "")
+	}
```
shizhMSFT, line 123: `graph.Referrers` functions differently with `graph.FindReferrerPredecessors`. They are not interchangeable.

shizhMSFT: Let's have a discussion on this.

internal/graph/graph.go
```diff
@@ -25,6 +25,34 @@ import (
...
+	Blobs []ocispec.Descriptor `json:"blobs,omitempty"`
+
+	// Subject (reference) is an optional link from the artifact to another manifest forming an association between the artifact and the other manifest.
+	Subject *ocispec.Descriptor `json:"subject,omitempty"`
+
+	// Annotations contains arbitrary metadata for the artifact manifest.
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
 // Successors returns the nodes directly pointed by the current node, picking
```
shizhMSFT, line 56: Starting from `image-spec v1.1.0-rc.4`, all manifests have a `subject` field.

@Wwwsylvia Is it the right time to promote this function to `oras-go`?

shizhMSFT: Also, all manifests now have an `artifactType` field.

cmd/oras/root/cp.go
qweeah: > Let's have a discussion on this.

Online or offline?

qweeah: `graph.Referrers` functions as a superset of with `graph.FindReferrerPredecessors`

internal/graph/graph.go
Wwwsylvia: Yeah I think such function will be useful.

cmd/oras/root/cp.go
shizhMSFT: After an offline discussion, the `Referrers()` can be used for `FindPredecessors`.

internal/graph/graph.go
```diff
@@ -109,8 +149,25 @@ func Referrers(ctx context.Context, target content.ReadOnlyGraphStorage, desc oc
...
+				return nil, err
+			}
+			if index.Subject == nil || !content.Equal(*index.Subject, desc) {
+				continue
+			}
+			node.ArtifactType = index.ArtifactType
+			node.Annotations = index.Annotations
 		default:
 			continue
 		}
```
shizhMSFT, line 173: Regarding to line 174:
```go
if node.ArtifactType != "" && (artifactType == "" || artifactType == node.ArtifactType)
```
Should it be
```go
if artifactType == "" || artifactType == node.ArtifactType
```
?

shizhMSFT: It seems an index can have a `subject` without an `artifactType`.

qweeah: According to the spec, `artifactType` is used (not empty) `when the manifest is used for an artifact`. So if `artifactType` is not presented, an index is not an artifact and should not be returned as a referrer.

> if node.ArtifactType != "" && (artifactType == "" || artifactType == node.ArtifactType)

This makes sure that `node.ArtifactType == "" && artifactType == ""` does not pass.

internal/graph/graph.go
```diff
@@ -171,7 +171,9 @@ func Referrers(ctx context.Context, target content.ReadOnlyGraphStorage, desc oc
...
-		if node.ArtifactType != "" && (artifactType == "" || artifactType == node.ArtifactType) {
+		if artifactType == "" || artifactType == node.ArtifactType {
+			// artifactType is allowed to be empty & empty will pass the filtering by default
```
shizhMSFT, line 175: This comment is requested for re-wording as the term `artifactType` is ambiguous since it be the `artifactType` of a manifest or the filter `artifactType`. The term `filtering` is also too generic and is requested to be re-worded so that other folks know it referrers to referrers filtering.

shizhMSFT: Try not to use `&`. Use `and` instead for formal writing.

shizhMSFT (approved):
LGTM with suggestions

PR #1055, opened 2023-08-08
chore: add triage labels

FeynmanZhou:

This PR adds a `triage` label to the ORAS issue templates. New issues will be labeled with `triage` so that we can quickly find the issues that need to be triaged.


shizhMSFT (approved):
LGTM

PR #1057, opened 2023-08-09
feat: support warning in remote targets

### LINKED ISSUE #898: show distribution warning headers as output on the ORAS CLI

1.0.0

Registry implementations MAY include informational warnings in [warning headers](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#warnings), as described in [RFC 7234](https://www.rfc-editor.org/rfc/rfc7234#section-5.5). User should be able to verbosely show the returned warning headers in the output if a target is a remote registry.

With this feature, oras is enabled to report the server-side warnings to the user in an unobtrusive way.

### LINKED ISSUE #1056: support OCI warning header
The PR https://github.com/opencontainers/distribution-spec/pull/393 has been merged to the main branch of the distribution spec.

ORAS commands related to remote registry or repository should print out such header if presented.

qweeah:
Fixes #898


cmd/oras/internal/option/remote.go
```diff
@@ -273,6 +283,16 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
...
 		return nil, err
 	}
+
+	if opts.distributionSpec.referrersAPI == nil || *opts.distributionSpec.referrersAPI {
+		once := sync.Once{}
+		repo.HandleWarning = func(warning remote.Warning) {
+			once.Do(func() {
+				warn(warning.Text)
+			})
+		}
```
shizhMSFT, line 293: I am confused with the `once`. Is that for deduplication? If so, it is probably wrong.

Consider that the server may send out:
```
Warning: foo
Warning: bar
Warning: foo
```
The client should print
```
Warning: foo
Warning: bar
```
The last warning of `foo` is deduplicated.

qweeah: Thanks for reminding, I am not aware of that. Is the warning deduplicated based on the warning full message?

cmd/oras/internal/option/remote.go
```diff
@@ -248,21 +249,30 @@ func (opts *Remote) Credential() auth.Credential {
...
-func (opts *Remote) NewRegistry(hostname string, common Common) (reg *remote.Registry, err error) {
+func (opts *Remote) NewRegistry(hostname string, warn func(...interface{}), common Common) (reg *remote.Registry, err error) {
 	reg, err = remote.NewRegistry(hostname)
 	if err != nil {
 		return nil, err
 	}
 	hostname = reg.Reference.Registry
 	reg.PlainHTTP = opts.isPlainHttp(hostname)
+
+	if opts.distributionSpec.referrersAPI == nil || *opts.distributionSpec.referrersAPI {
```
Wwwsylvia, line 260: Why do we check Referrers API here? Can you help me to understand it?

Wwwsylvia (commented):
We need to depulicate warnings across multiple associated requests.

cmd/oras/internal/option/remote.go
qweeah: I assumed it was introduced after 1.1.0 so should be avoid if referrers API is explicitly specified as not supported. Looks like those two features are irrelevant.

qweeah: Updated

cmd/oras/internal/option/remote.go
qweeah: discussed offline with @Wwwsylvia and resolved this comment

cmd/oras/internal/option/remote.go
```diff
@@ -273,6 +283,16 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
...
 	}
+
+	if opts.distributionSpec.referrersAPI == nil || *opts.distributionSpec.referrersAPI {
+		once := sync.Once{}
+		repo.HandleWarning = func(warning remote.Warning) {
+			once.Do(func() {
+				warn(warning.Text)
+			})
+		}
+	}
```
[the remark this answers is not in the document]
qweeah: removed referrers API checks

qweeah:
> We need to deduplicate warnings across multiple associated requests.

Yes. Requests sent in the same execution are considered as associated.

Wwwsylvia:
Wow many files

cmd/oras/internal/option/remote.go
```diff
@@ -248,21 +250,27 @@ func (opts *Remote) Credential() auth.Credential {
 }
 
 // NewRegistry assembles a oras remote registry.
-func (opts *Remote) NewRegistry(hostname string, common Common) (reg *remote.Registry, err error) {
+func (opts *Remote) NewRegistry(hostname string, warn func(...interface{}), common Common) (reg *remote.Registry, err error) {
```
Wwwsylvia, line 253: nit: Why not using `any` instead of `interface{}`?
```go
warn func(...any)
```
?

cmd/oras/internal/option/remote.go
```diff
@@ -53,6 +54,7 @@ type Remote struct {
 	distributionSpec      distributionSpec
 	headerFlags           []string
 	headers               http.Header
+	hasWarned             sync.Map
```
Wwwsylvia, line 57: nit: How about just call it `warned`?

cmd/oras/internal/option/remote.go
```diff
@@ -273,6 +281,11 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
 	if repo.Client, err = opts.authClient(hostname, common.Debug); err != nil {
 		return nil, err
 	}
+	repo.HandleWarning = func(warning remote.Warning) {
+		if _, loaded := opts.hasWarned.LoadOrStore(warning.WarningValue, true); loaded {
```
Wwwsylvia, line 285: Should we warn only when the warning value is first stored? That it, `!loaded`?

Wwwsylvia (commented):
BTW I happened to find that the temp test file created by [`TestNewStoreError`](https://github.com/oras-project/oras/blob/main/internal/credential/store_test.go#L29) is not properly closed and this is causing the unit tests to fail on Windows.

cmd/oras/internal/option/remote.go
```diff
@@ -248,21 +250,27 @@ func (opts *Remote) Credential() auth.Credential {
...
 		return nil, err
 	}
 	hostname = reg.Reference.Registry
 	reg.PlainHTTP = opts.isPlainHttp(hostname)
+
+	reg.HandleWarning = func(warning remote.Warning) {
+		if _, ok := opts.hasWarned.LoadOrStore(warning.WarningValue, true); ok {
+			warn(warning.Text)
+		}
+	}
```
shizhMSFT: In a copy scenario, should the oras CLI log differently for source or target? I think it can be done by adding a log field to the logger.

cmd/oras/internal/option/remote.go
shizhMSFT: It should be consistent with line 262. Use `ok` or `loaded` for both places.

shizhMSFT: same applies to line 262.

cmd/oras/internal/option/remote.go
qweeah: `NewRegistry` is only used in `oras repo ls` and `oras login` command, so the host name doesn't need to be added here.

cmd/oras/internal/option/remote.go
qweeah: Updated

qweeah:
@Wwwsylvia test fixed, can you help review again.

cmd/oras/root/discover.go
```diff
@@ -189,7 +189,7 @@ func printDiscoveredReferrersTable(refs []ocispec.Descriptor, verbose bool) erro
 		print(ref.ArtifactType, ref.Digest)
 		if verbose {
 			if err := printJSON(ref); err != nil {
-				return fmt.Errorf("Error printing JSON: %w", err)
+				return fmt.Errorf("Eirror printing JSON: %w", err)
```
Wwwsylvia, line 192: nit: typo

Wwwsylvia (approved):
LGTM, but I am not a maintainer

cmd/oras/internal/option/remote.go
```diff
@@ -273,6 +281,12 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
 	if repo.Client, err = opts.authClient(hostname, common.Debug); err != nil {
 		return nil, err
 	}
+	logger = logger.WithField("hostname", hostname)
```
shizhMSFT, line 284: It is better to name the field as `registry` instead of `hostname` since `hostname := repo.Reference.Registry`.

It happens to `docker.io`. The registry name is `docker.io` but the hostname is `registry-1.docker.io`.

cmd/oras/internal/option/remote.go
```diff
@@ -273,6 +281,12 @@ func (opts *Remote) NewRepository(reference string, common Common) (repo *remote
...
+	repo.HandleWarning = func(warning remote.Warning) {
+		if _, loaded := opts.warned.LoadOrStore(warning.WarningValue, true); !loaded {
+			logger.Warn(warning.Text)
+		}
+	}
```
shizhMSFT, line 289: It seems we can refactor this code so that it can also be applied to `NewRegistry()`.
```go
func (opts *Remote) handleWarning(registry string, logger logrus.FieldLogger) func(warning remote.Warning) {
	logger = logger.WithField("registry", registry )
	return func(warning remote.Warning) {
		if _, loaded := opts.warned.LoadOrStore(warning.WarningValue, true); !loaded {
			logger.Warn(warning.Text)
		}
	}
}
```
and
```suggestion
	repo.HandleWarning = opts.handleWarning(hostname, logger)
```

shizhMSFT: With `warned` removed from the `Remote` struct, we don't need to make so many changes for passing the options from values to pointers.

qweeah: Added registry to the key for depulication.

qweeah: @shizhMSFT  If I am copying from `my.registry/repo1` to `my.registry/repo2` and `my.registry` will always give only one warning value, how many warning should be printed out? 
 1 => the map should be in the remote option
 2 => the map can be created within the handler function

shizhMSFT: I see. If they are two repos from the same registry, only 1 should be printed out.

qweeah: will update `sync.Map` to a pointer in remote option

qweeah: updated

cmd/oras/internal/option/remote.go
```diff
@@ -53,6 +55,7 @@ type Remote struct {
 	distributionSpec      distributionSpec
 	headerFlags           []string
 	headers               http.Header
+	warned                *sync.Map
```
shizhMSFT, line 58:
```suggestion
	warned                map[string]*sync.Map
```

cmd/oras/internal/option/remote.go
```diff
@@ -246,33 +250,46 @@ func (opts *Remote) parseCustomHeaders() error {
 func (opts *Remote) Credential() auth.Credential {
 	return credential.Credential(opts.Username, opts.Password)
 }
+func (opts *Remote) handleWarning(registry string, logger logrus.FieldLogger) func(warning remote.Warning) {
```
shizhMSFT, line 253: Maybe we can do
```go
if opts.warned == nil {
	opts.warned = make(map[string]*sync.Map)
}
warned := opts.warned[registry]
if warned == nil {
	warned = &sync.Map{}
	opts.warned[registry] = warned
}
logger = logger.WithField("registry", registry)
return func(warning remote.Warning) {
	if _, loaded := warned.LoadOrStore(warning.WarningValue, struct{}{}); !loaded {
		logger.Warn(warning.Text)
	}
}
```

Wwwsylvia (commented):
Should `ImageSpec.PackType` be changed to `ImageSpec.PackVersion`?

qweeah:
> Should `ImageSpec.PackType` be changed to `ImageSpec.PackVersion`?

Good catch. This is not related to change here, I will do in another PR.

PR #1070, opened 2023-08-22
bump: use new packing function in oras-go

qweeah:
This PR uses `PackManifest` in push and attach.


shizhMSFT (approved):
LGTM

PR #1072, opened 2023-08-23
test(e2e): add zot as a testing backend

### LINKED ISSUE #710: Standardize test data setup for E2E tests
For now, command suites require pre-baked storage image in distribution storage layout. Once https://github.com/oras-project/oras/issues/378 finishes, mount files can be changed to OCI layout so the preparation process can be done via `oras copy`

See original discussion:
        > Looks like there is very docker-specific env setup. Should probably standardize on that also.

The env setup now is based on mount files in distribution storage layout. ZOT storage is designed based on OCI layout so it's not compatible. We will definitely standardize the setup process when ORAS support OCI layout target.

_Originally posted by @qweeah in https://github.com/oras-project/oras/issues/709#issuecomment-1334723703_

### LINKED ISSUE #1071: add zot as a registry service backend for e2e tests

oras 1.1.0-rc.1

Add zot as a registry service backend for e2e tests

oras distribution only implements OCI image-spec v1.1.0 rc2, which doesn't allow image to have `artifactType` field and referrers API filtering is implemented based on `image.config.mediaType`. We should utilize [zot](https://github.com/project-zot/zot/tree/v2.0.0-rc5) to cover the OCI image-spec v1.1.0 rc4 testing to make sure oras is compliant to the latest spec

qweeah:
This is the one of the two PRs that related to #1071. This PR added ZOT as an E2E backend for auth and scenario suite.


shizhMSFT (approved):
LGTM

PR #1081, opened 2023-08-25
test(e2e): update e2e specs for `oras attach` 

qweeah:
This PR make e2e specs of `oras attach` run against ZOT with proper test data.

PR #1083, opened 2023-08-28
test(e2e): use ZOT as testing backend for `oras blob` specs

qweeah:
This PR uses ZOT as the testing backend for `oras blob` E2E specs, depends on #1081


test/e2e/suite/command/blob.go
```diff
@@ -169,28 +171,22 @@ var _ = Describe("1.1 registry users:", func() {
...
+			CopyZOTRepo(BlobRepo, dstRepo)
+			toDeleteRef := RegistryRef(ZOTHost, dstRepo, foobar.FooBlobDigest)
 			ORAS("blob", "delete", toDeleteRef).
 				WithInput(strings.NewReader("y")).
 				MatchKeyWords("Deleted", toDeleteRef).Exec()
-			ORAS("blob", "delete", toDeleteRef).
-				WithDescription("validate").
-				WithInput(strings.NewReader("y")).
-				ExpectFailure().
-				MatchErrKeyWords("Error:", toDeleteRef, "the specified blob does not exist").Exec()
```
wangxiaoxuan273, line 181: Why remove this `validate` part? What happens on Zot if we delete the blob again?

qweeah: ZOT's implementation if not like distribution, it returns 202 for deletion, which mean the deletion will be done asynchronizely. The blob can still be fetched and re-delete after being deleted :)

I raised a query in ZOT's slack channel and waiting for their confirm https://cloud-native.slack.com/archives/C03EGRE4QGH/p1693210694637909

wangxiaoxuan273: So if I delete a blob and **immediately** fetch it, the deletion may not yet be completed and I may still get it? And if I fetch it after a day, the deletion may be done and I'll get 404? Is it right?

qweeah: Sounds strange but yes. I am playing with the GC configuration but haven't find a way to instantly reflect a blob deletion yet.

shizhMSFT (approved):
LGTM

PR #1094, opened 2023-08-31
test(e2e): re-bake test data

qweeah:
This PR re-bakes test data to make it testable in ZOT.

The old tests data contains `mediaType` that cannot be copied to ZOT and it affects `oras cp` tests when copying between v1.1-compliant and v1.0-compliant registries.


shizhMSFT (approved):
LGTM

PR #1095, opened 2023-08-31
test(e2e): migrate oras discover specs to ZOT

qweeah:
Depends on merging of #1094


qweeah:
Closing as `discover` and `cp` should be migrated in one PR otherwise test will fail

PR #1102, opened 2023-09-01
doc: update example for `oras push`

qweeah:
This PR updates example for `oras push` command.

### DIFF
--- a/cmd/oras/root/push.go
+++ b/cmd/oras/root/push.go
@@ -68,11 +68,12 @@ Example - Push file "hi.txt" with the custom media type "application/vnd.me.hi":
 Example - Push multiple files with different media types:
   oras push localhost:5000/hello:v1 hi.txt:application/vnd.me.hi bye.txt:application/vnd.me.bye
 
+Example - Push file "hi.txt" with artifact type "application/vnd.example+type":
+  oras push --artifact-type application/vnd.example+type localhost:5000/hello:v1 hi.txt
+
 Example - Push file "hi.txt" with config type "application/vnd.me.config":
   oras push --image-spec v1.0 --artifact-type application/vnd.me.config localhost:5000/hello:v1 hi.txt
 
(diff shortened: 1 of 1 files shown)


shizhMSFT (commented):
The change LGTM but it seems we miss an example of setting artifact type.

cmd/oras/root/push.go
```diff
@@ -69,7 +69,7 @@ Example - Push multiple files with different media types:
   oras push localhost:5000/hello:v1 hi.txt:application/vnd.me.hi bye.txt:application/vnd.me.bye
 
 Example - Push file "hi.txt" with config type "application/vnd.me.config":
-  oras push --artifact-type application/vnd.me.config localhost:5000/hello:v1 hi.txt
+  oras push --image-spec v1.0 --artifact-type application/vnd.me.config localhost:5000/hello:v1 hi.txt
```
Wwwsylvia, line 72: This seems to be the only example using `--artifact-type`. Do we need another example without `--image-spec v1.0`? Something like

```
Example - Push file "hi.txt" with artifact type "application/vnd.me.artifact":
  oras push --artifact-type application/vnd.me.artifact localhost:5000/hello:v1 hi.txt
```

qweeah: added

cmd/oras/root/push.go
```diff
@@ -69,7 +69,10 @@ Example - Push multiple files with different media types:
...
+
+Example - Push file "hi.txt" with artifact type "application/vnd.example+type":
+  oras push --artifact-type application/vnd.example+type localhost:5000/hello:v1 hi.txt
```
shizhMSFT, line 75: This example of artifact type should be before the example of config type.

Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #893, opened 2023-03-17
Show progress bar when downloading or pushing content from/to a registry


FeynmanZhou:
### What is the version of your ORAS CLI

v1.0.0-RC.2

### What would you like to be added?

Per the discussion at [ORAS slack channel](https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1679022569583329?thread_ts=1679013600.539599&cid=CJ1KHJM5Z), @vsoch suggests design and provide a progress bar when downloading content from the remote registry. 

### Why is this needed for ORAS?

It would be helpful to show the progress bar in ORAS CLI commands such as [oras blob fetch](https://oras.land/cli_reference/2_oras_blob_fetch/), especially when downloading a big layer or blob. Sometimes, the file transfer might be stuck but users are not aware of the progress. If  ORAS output displays no output for a while, it will look broken. A good spinner or progress bar/indicator can make a CLI appear to be easier to use. It’s also good to show the estimated time remaining, or even just have an animated component, to reassure them that you’re still working on it.

Similarly, we could also add a progress bar to the `push` scenario. This would be a UX improvement. 

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

shizhMSFT:
Although we don't have a progress bar implemented for `oras` yet, it is possible with `jq` and `pv`. Here's an example:
```console
$ blob_ref=ghcr.io/oras-project/oras@sha256:74529e03eae02d1fff5c05e9a6ec2089e1f5ae96421169c29a7c165346e042e4
$ size=$(oras blob fetch --descriptor $blob_ref | jq -r .size)
$ oras blob fetch $blob_ref --output - | pv -s $size > layer.bin
1.58MiB 0:00:03 [1.47MiB/s] [================================>                                         ] 45% ETA 0:00:03
```

sbdtu5498:
How about using https://github.com/schollz/progressbar rather than depending on external utility. I have done the basic changes for the issue. Should we also implement the same for blob push as well.

amands98:
I am interested in working on this. // @sajayantony @FeynmanZhou @qweeah

qweeah:
@amands98 Sorry for not updated the status but I have already worked on this, I have two PoC in my own fork, please check
https://github.com/qweeah/oras/releases/tag/v1.2.0-spinner
https://github.com/qweeah/oras/releases/tag/v1.2.0-bar

qweeah:
[...]

PR #1113, opened 2023-09-11
feat: support progressed output in `blob push` and `blob get`

qweeah:
This PR introduces terminal output for certain oras commands. 

Fixes https://github.com/oras-project/oras/issues/893


cmd/oras/internal/display/console/console.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, err
+	}
+	return &Console{c}, nil
+}
+
+// Save saves the current cursor position.
+func (c *Console) Save() {
+	fmt.Fprint(c, aec.Hide)
+	// cannot use aec.Save since DEC has better compatilibity than SCO
+	fmt.Fprint(c, "\0337")
```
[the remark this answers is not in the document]
qweeah: A: should be more direct since `fmt.Fprint` also uses `c.Write`

cmd/oras/internal/display/progress/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	return ch
+}
+
+// Close stops all status and waits for updating and rendering.
+func (m *manager) Close() error {
+	if m.closed() {
+		return errManagerStopped
+	}
+
+	// 1 wait for all model update done
```
shizhMSFT, line 138: nit:
```suggestion
	// 1. wait for all model update done
```

cmd/oras/internal/display/track/reader.go
```diff
@@ -0,0 +1,95 @@
...
+	"os"
+	"sync/atomic"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras/cmd/oras/internal/display/progress"
+)
+
+type reader struct {
+	base         io.Reader
+	offset       atomic.Uint64
```
shizhMSFT, line 29: What's reason to make `offset` an atomic value?
> [!Note]
> All `io.Reader`s are assumed NOT go-routine safe.

cmd/oras/internal/option/common.go
```diff
@@ -17,25 +17,44 @@ package option
...
 	"oras.land/oras/internal/trace"
 )
 
 // Common option struct.
 type Common struct {
 	Debug   bool
 	Verbose bool
+	TTY     *os.File
+
+	avoidTTY bool
```
shizhMSFT, line 35: nit: consistency
```suggestion
	noTTY bool
```

cmd/oras/internal/option/common.go
```diff
@@ -17,25 +17,44 @@ package option
...
 func (opts *Common) WithContext(ctx context.Context) (context.Context, logrus.FieldLogger) {
 	return trace.NewLogger(ctx, opts.Debug, opts.Verbose)
 }
+
+// Parse gets target options from user input.
+func (opts *Common) Parse() error {
+	f := os.Stderr
+	if !opts.avoidTTY && term.IsTerminal(int(f.Fd())) {
+		if opts.Debug {
+			return errors.New("cannot use --debug, add --noTTY to suppress terminal output")
```
shizhMSFT, line 55:
```suggestion
			return errors.New("cannot use --debug, add --no-tty to suppress terminal output")
```

cmd/oras/internal/option/common.go
```diff
@@ -17,25 +17,44 @@ package option
...
 }
 
 // WithContext returns a new FieldLogger and an associated Context derived from ctx.
 func (opts *Common) WithContext(ctx context.Context) (context.Context, logrus.FieldLogger) {
 	return trace.NewLogger(ctx, opts.Debug, opts.Verbose)
 }
+
+// Parse gets target options from user input.
+func (opts *Common) Parse() error {
+	f := os.Stderr
```
shizhMSFT, line 52: Request documentation on setting `os.Stderr` as the output.

cmd/oras/internal/display/progress/manager.go
```diff
@@ -0,0 +1,154 @@
...
+		for {
+			m.render()
+			select {
+			case <-m.renderDone:
+				m.render()
+				close(m.renderClosed)
+				return
+			case <-renderTicker.C:
+			}
+		}
```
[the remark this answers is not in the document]
qweeah: > Note: the above code is different from
Yes, the latter won't render output immediately after being started.

I think the code can be refactored to below so `manager.Close` only need to stop rendering and wait render to be closed.
```go
for {
	select {
	case <-m.renderDone:
		m.updating.Wait()
		m.render()
		close(m.renderClosed)
		return
	case <-renderTicker.C:
		m.render()
	}
}
```

cmd/oras/internal/option/common.go
qweeah: What kind of docs are you suggesting? This is to accommodate the upcoming change for piped UX, so user can see the progress bar and use `-o json` at the same time.

cmd/oras/internal/display/track/reader.go
qweeah: Just trying to be go-routine safe here. Is it too much?

cmd/oras/internal/option/common.go
qweeah: Added

cmd/oras/internal/display/track/reader.go
qweeah: Changed the implementation to non-atomic

cmd/oras/internal/display/progress/manager.go
```diff
@@ -0,0 +1,153 @@
...
+func (m *manager) start() {
+	m.console.Save()
+	renderTicker := time.NewTicker(bufFlushDuration)
+	go func() {
+		defer m.console.Restore()
+		defer renderTicker.Stop()
+		for {
+			select {
+			case <-m.renderDone:
+				m.updating.Wait()
```
shizhMSFT, line 76: `m.updating.Wait()` should be between line 138 and 139 before `close(m.renderDone)` so that the `manager` can keep rendering while waiting for all status to complete.

cmd/oras/internal/display/progress/status.go
```diff
@@ -0,0 +1,207 @@
...
+		lenMargin = 0
+	}
+	return fmt.Sprintf("%s%s%s", left, strings.Repeat(" ", lenMargin), right), fmt.Sprintf("  └─ %s", s.descriptor.Digest.String())
+}
+
+// calculateSpeed calculates the speed of the progress and update last status.
+// caller must hold the lock.
+func (s *status) calculateSpeed() bytes {
+	now := time.Now()
+	secondsTaken := now.Sub(s.lastRenderTime).Seconds()
```
shizhMSFT, line 152: What will happen when `s.lastRenderTime` is zero? Should we use `startTime` instead?

cmd/oras/internal/display/track/reader.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil && err != io.EOF {
+		return n, err
+	}
+
+	r.offset = r.offset + int64(n)
+	if err == io.EOF {
+		if r.offset != r.descriptor.Size {
+			return n, io.ErrUnexpectedEOF
+		}
+		r.status <- progress.NewStatus(r.actionPrompt, r.descriptor, r.offset)
```
shizhMSFT, line 95: There is a perf issue regarding to the channel on line 95 and line 100.

Since `r.status` is **not buffered**, the read process is **blocked** if the manager fail to render in time.
Even if we make `r.status` is buffered, it is still possible that the buffer is full. Therefore, we need to make sure the `Read()` function will never be block by the `r.status <-` operation.

cmd/oras/internal/option/common.go
```diff
@@ -17,25 +17,49 @@ package option
...
+	TTY     *os.File
+
+	noTTY bool
 }
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Common) ApplyFlags(fs *pflag.FlagSet) {
 	fs.BoolVarP(&opts.Debug, "debug", "d", false, "debug mode")
 	fs.BoolVarP(&opts.Verbose, "verbose", "v", false, "verbose output")
+	fs.BoolVarP(&opts.noTTY, "no-tty", "", false, "[Preview] avoid using stdout as a terminal")
```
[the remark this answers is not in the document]
shizhMSFT: /cc @FeynmanZhou seeking inputs on the flag name and description.

cmd/oras/internal/display/progress/humanize.go
```diff
@@ -0,0 +1,51 @@
...
+	unit string
+}
+
+// ToBytes converts size in bytes to human readable format.
+func ToBytes(sizeInBytes int64) bytes {
+	f := float64(sizeInBytes)
+	if f < base {
+		return bytes{f, units[0]}
+	}
+	e := math.Floor(math.Log(f) / math.Log(base))
```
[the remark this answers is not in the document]
qweeah: Let's set it to `len(units)-1`

cmd/oras/internal/display/progress/manager.go
qweeah: Nice catch. Will revert.

cmd/oras/internal/display/progress/status.go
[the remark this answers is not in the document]
qweeah: How about setting it to a value no smaller than `bufFlushDuration`?

cmd/oras/internal/display/track/reader.go
qweeah: Wow, this trick is amazing!!! In this case we only need the buffer size to be 2? Anyway old status doesn't need to be updated to the manager?

cmd/oras/internal/display/progress/manager.go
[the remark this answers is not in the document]
qweeah: Confirmed that using `m.updating.Wait()` during rendering will make UX update stuck in some edge cases.

cmd/oras/internal/option/common.go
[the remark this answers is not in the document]
FeynmanZhou: If the `no-tty` only applies to the progress bar feature in ORAS so far, how about rewording the description to be much simpler and straightforward? 

```
fs.BoolVarP(&opts.noTTY, "no-tty", "", false, "[Preview] do not show progress output")
````

File: cmd/oras/root/blob/push.go [hunk 13]
```diff
@@ -145,3 +140,32 @@ func pushBlob(ctx context.Context, opts pushBlobOptions) (err error) {
...
+		defer trackedReader.StopManager()
+		trackedReader.Start()
+		r = trackedReader
+		if err := t.Push(ctx, desc, r); err != nil {
+			return err
+		}
+		trackedReader.Done()
+	}
+	return nil
+}
```
shizhMSFT, line 171: nit:
```suggestion
func (opts *pushBlobOptions) doPush(ctx context.Context, t oras.Target, desc ocispec.Descriptor, r io.Reader) error {
	if opts.TTY == nil {
		// none tty output
		if err := display.PrintStatus(desc, "Uploading", opts.Verbose); err != nil {
			return err
		}
		if err := t.Push(ctx, desc, r); err != nil {
			return err
		}
		return display.PrintStatus(desc, "Uploaded ", opts.Verbose)
	} 
	
	// tty output
	trackedReader, err := track.NewReader(r, desc, "Uploading", "Uploaded ", opts.TTY)
	if err != nil {
		return err
	}
	defer trackedReader.StopManager()
	trackedReader.Start()
	r = trackedReader
	if err := t.Push(ctx, desc, r); err != nil {
		return err
	}
	trackedReader.Done()
	return nil
}
```

cmd/oras/internal/display/progress/status.go
```diff
@@ -147,15 +148,16 @@ func (s *status) String(width int) (string, string) {
 
 // calculateSpeed calculates the speed of the progress and update last status.
 // caller must hold the lock.
-func (s *status) calculateSpeed() bytes {
+func (s *status) calculateSpeed() humanize.Bytes {
 	now := time.Now()
 	secondsTaken := now.Sub(s.lastRenderTime).Seconds()
+	secondsTaken = math.Max(secondsTaken, float64(bufFlushDuration.Milliseconds())/1000)
```
shizhMSFT, line 154: Instead of using `math.Max`, can you use `if secondsTaken == 0` instead as it is an extremely rare case?

cmd/oras/root/blob/fetch.go
```diff
@@ -147,40 +147,35 @@ func (opts *fetchBlobOptions) doFetch(ctx context.Context, src oras.ReadOnlyTarg
...
-		// save blob content into the local file if the output path is provided
-		file, err := os.Create(opts.outputPath)
-		if err != nil {
-			return ocispec.Descriptor{}, err
+			writer = file
+			defer func() {
+				if err := file.Close(); fetchErr == nil {
+					fetchErr = err
+				}
+			}()
```
shizhMSFT, line 162: nit: coding style: `defer` first after error handling then other code
```suggestion
			defer func() {
				if err := file.Close(); fetchErr == nil {
					fetchErr = err
				}
			}()
			writer = file
```

cmd/oras/internal/display/progress/status.go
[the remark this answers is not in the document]
shizhMSFT: See https://github.com/oras-project/oras/pull/1113#discussion_r1357885268

shizhMSFT: @qweeah This still need to be addressed.

shizhMSFT (approved):
LGTM

PR #1122, opened 2023-09-18
feat: support copying referrers for multi-arch images

### LINKED ISSUE #816: support copying image referrers from index predecessor via `oras copy`
Suppose we have an index directly pointing to several images with different platforms. Each image have an artifact attached as below:
```mermaid
graph TD
Z>v1] -..-> index
A["artifact(linux/s390x)"] -- subject --> B["image"]
C["artifact(linux/amd64)"] -- subject --> D["image"]
E["artifact(linux/arm64)"] -- subject --> F["image"]

index -- linux/s390x --> B
index -- linux/arm64 --> D
index -- linux/amd64 --> F

```

Using below command, oras user can copy all the images and artifacts into a new target (registry or image layout) and tag the index in the destination as `v2`:
```
oras cp $FROM:v1 $TO:v2 --recursive
```

qweeah:
This PR changes `oras cp` to recursively copy referrers of platform-specific images in an index or manifest list.

Resolves #816

### DIFF
--- a/cmd/oras/root/cp.go
+++ b/cmd/oras/root/cp.go
@@ -117,34 +117,23 @@ func runCopy(ctx context.Context, opts copyOptions) error {
 	}
 
 	// Prepare copy options
-	started := &sync.Map{}
 	committed := &sync.Map{}
 	extendedCopyOptions := oras.DefaultExtendedCopyOptions
 	extendedCopyOptions.Concurrency = opts.concurrency
 	extendedCopyOptions.FindPredecessors = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
 		return graph.Referrers(ctx, src, desc, "")
 	}
-	extendedCopyOptions.PreCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
-		if _, loaded := started.LoadOrStore(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle]); !loaded {
-			return display.PrintStatus(desc, "Copying", opts.Verbose)
-		}
-		return nil
-	}
+	extendedCopyOptions.PreCopy = display.StatusPrinter("Copying", opts.Verbose)
 	extendedCopyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
-		started.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
-		if _, loaded := committed.LoadOrStore(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle]); loaded {
-			return nil
-		}
-		if err := display.PrintSuccessorStatus(ctx, desc, "Skipped", dst, started, opts.Verbose); err != nil {
+		committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
+		if err := display.PrintSuccessorStatus(ctx, desc, "Skipped", dst, committed, opts.Verbose); err != nil {
 			return err
 		}
 		return display.PrintStatus(desc, "Copied ", opts.Verbose)
 	}
 	extendedCopyOptions.OnCopySkipped = func(ctx context.Context, desc ocispec.Descriptor) error {
-		if _, loaded := started.LoadOrStore(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle]); !loaded {
-			return display.PrintStatus(desc, "Exists ", opts.Verbose)
-		}
-		return nil
+		committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
+		return display.PrintStatus(desc, "Exists ", opts.Verbose)
 	}
 
 	var desc ocispec.Descriptor
@@ -210,18 +199,25 @@ func recursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.T
 		if err = json.Unmarshal(fetched, &index); err != nil {
 			return nil
 		}
-
-		var wg sync.WaitGroup
+		// point referrers child manifests to root
+		var referrers []ocispec.Descriptor
 		for _, desc := range index.Manifests {
-			wg.Add(1)
-			go func(desc ocispec.Descriptor) {
-				defer wg.Done()
-				err = oras.ExtendedCopyGraph(ctx, src, dst, desc, opts.ExtendedCopyGraphOptions)
-			}(desc)
+			descs, err := graph.Referrers(ctx, src, desc, "")
+			if err != nil {
+				return err
+			}
+			referrers = append(referrers, descs...)
 		}
-		wg.Wait()
-		if err != nil {
-			return err
+		findPredecessor := opts.FindPredecessors
+		opts.FindPredecessors = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+			descs, err := findPredecessor(ctx, src, desc)
+			if err != nil {
+				return nil, err
+			}
+			if content.Equal(desc, root) {
+				descs = append(descs, referrers...)
+			}
+			return descs, nil
 		}
 	}
 
--- a/cmd/oras/root/cp.go
+++ b/cmd/oras/root/cp.go
@@ -211,6 +211,7 @@ func recursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.T
 				return nil, err
 			}
 			if content.Equal(desc, root) {
+				// make sure referrers of child manifests are copied by pointing them to root
 				descs = append(descs, referrers...)
 			}
 			return descs, nil
--- a/cmd/oras/root/cp.go
+++ b/cmd/oras/root/cp.go
@@ -199,12 +199,12 @@ func recursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.T
 		}
 
 		// point referrers of child manifests to root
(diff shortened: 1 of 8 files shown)


cmd/oras/root/cp.go
```diff
@@ -191,3 +196,41 @@ func runCopy(ctx context.Context, opts copyOptions) error {
...
+		}
+	}
+
+	if dstRef == "" || dstRef == root.Digest.String() {
+		err = oras.ExtendedCopyGraph(ctx, src, dst, root, opts.ExtendedCopyGraphOptions)
+	} else {
+		_, err = oras.ExtendedCopy(ctx, src, root.Digest.String(), dst, dstRef, opts)
+	}
+	return err
+
```
shizhMSFT, line 235: nit: remove empty line

cmd/oras/root/cp.go
```diff
@@ -115,60 +117,63 @@ func runCopy(ctx context.Context, opts copyOptions) error {
...
 				CopyGraphOptions: extendedCopyOptions.CopyGraphOptions,
 			}
 			if opts.Platform.Platform != nil {
 				copyOptions.WithTargetPlatform(opts.Platform.Platform)
 			}
-			desc, err = oras.Copy(ctx, src, opts.From.Reference, dst, dstRef, copyOptions)
+			desc, err = oras.Copy(ctx, src, opts.From.Reference, dst, opts.To.Reference, copyOptions)
 		}
 	}
+
```
shizhMSFT, line 176: nit: remove empty line

cmd/oras/root/cp.go
```diff
@@ -191,3 +196,41 @@ func runCopy(ctx context.Context, opts copyOptions) error {
 
 	return nil
 }
+
+// RecursiveCopy copies an artifact and its referrers from one target to another.
+// If the artifact is a manifet list or index, its predecessor's referrers are copied as well.
```
shizhMSFT, line 201:
```suggestion
// If the artifact is a manifest list or index, referrers of its manifests are copied as well.
```

cmd/oras/root/cp.go
```diff
@@ -191,3 +196,41 @@ func runCopy(ctx context.Context, opts copyOptions) error {
...
+func RecursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.Target, dstRef string, root ocispec.Descriptor, opts oras.ExtendedCopyOptions) error {
```
shizhMSFT, line 202: @qweeah Let's discuss offline on the `RecursiveCopy` so that we can do better concurrency control.

cmd/oras/root/cp.go
```diff
@@ -191,3 +195,40 @@ func runCopy(ctx context.Context, opts copyOptions) error {
 
 	return nil
 }
+
+// RecursiveCopy copies an artifact and its referrers from one target to another.
+// If the artifact is a manifest list or index, referrers of its manifests are copied as well.
+func RecursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.Target, dstRef string, root ocispec.Descriptor, opts oras.ExtendedCopyOptions) error {
```
wangxiaoxuan273, line 201: Is it better to make this function private, since it's only used by `runCopy()`

cmd/oras/root/cp.go
```diff
@@ -115,58 +117,60 @@ func runCopy(ctx context.Context, opts copyOptions) error {
 	}
 
 	// Prepare copy options
+	started := &sync.Map{}
 	committed := &sync.Map{}
```
wangxiaoxuan273, line 121: Do we really need the two variables `started` and `committed`? Maybe one is enough?

cmd/oras/root/cp.go
```diff
@@ -191,3 +195,40 @@ func runCopy(ctx context.Context, opts copyOptions) error {
...
+			wg.Add(1)
+			go func(desc ocispec.Descriptor) {
+				defer wg.Done()
+				err = oras.ExtendedCopyGraph(ctx, src, dst, desc, opts.ExtendedCopyGraphOptions)
+			}(desc)
+		}
+		wg.Wait()
+		if err != nil {
+			return err
+		}
```
wangxiaoxuan273, line 225: Should you copy the index/manifest list here and return, instead of going into the next `if` and then `ExtendedCopyGraph`/`ExtendedCopy`?

cmd/oras/root/cp.go
```diff
@@ -191,3 +195,40 @@ func runCopy(ctx context.Context, opts copyOptions) error {
...
+		var wg sync.WaitGroup
+		for _, desc := range index.Manifests {
+			wg.Add(1)
+			go func(desc ocispec.Descriptor) {
+				defer wg.Done()
+				err = oras.ExtendedCopyGraph(ctx, src, dst, desc, opts.ExtendedCopyGraphOptions)
+			}(desc)
+		}
+		wg.Wait()
+		if err != nil {
```
wangxiaoxuan273, line 223: Can this line catch the `err` in line 219?

cmd/oras/root/cp.go
qweeah: Good idea on making it private since it will only be used by `oras cp`. Regarding the naming, better still implies that the copy will be done in recursive way.

cmd/oras/root/cp.go
qweeah: Since `oras.Copy` will be called more than once, we cannot rely on oras-go to de-duplicate the copying. 

Say we need to copy two blobs `A` and `B` with same digest in difference calls of `oras.Copy` with below time sequence:
1) copy A
2) copy B
3) done copying A
4) done copying B

The logs will be like
```
Copying <DIGEST_OF_A>
Copying <DIGEST_OF_B>
Copied  <DIGEST_OF_A>
# no Copied  <DIGEST_OF_B> since A is commited
```
So we need to add `started` to track whether a copying is started or not, to get rid of the `Copying <DIGEST_OF_B>`.

cmd/oras/root/cp.go
qweeah: go routine doesn't support returning error directly.

cmd/oras/root/cp.go
qweeah: It's done in the below if-else block.

cmd/oras/root/cp.go
wangxiaoxuan273: It looks like the error returned may be overwritten, if we write the error handling this way.

cmd/oras/root/cp.go
wangxiaoxuan273: I see that we don't have a public `copyNode` function in `oras-go`, so probably we can only do it this way.

cmd/oras/root/cp.go
wangxiaoxuan273: But then do we still need `committed`? Currently `committed` is like, not actively used anywhere.

cmd/oras/root/cp.go
qweeah: indeed, I have changed implementation, PTAL

cmd/oras/root/cp.go
qweeah: The new recursive copy will be implemented via adding virtual links from referrers of child manifests to index, see dashed links in below graph:
```mermaid
graph TD
Z>v1] -----> index
A["artifact(linux/s390x)"] -- subject --> B["image"]
C["artifact(linux/amd64)"] -- subject --> D["image"]
E["artifact(linux/arm64)"] -- subject --> F["image"]

index -- linux/s390x --> B
index -- linux/arm64 --> D
index -- linux/amd64 --> F

A-.->index
C-.->index
E-.->index
```

cmd/oras/root/cp.go
qweeah: > Should you copy the index/manifest list here and return

Still the index/manifest list should be copied with its referrers.

cmd/oras/root/cp.go
qweeah: `committed` is used for deduplicating post copy logs. Resolving as the implementation is already changed.

cmd/oras/root/cp.go
```diff
@@ -191,3 +184,48 @@ func runCopy(ctx context.Context, opts copyOptions) error {
 
 	return nil
 }
+
+// recursiveCopy copies an artifact and its referrers from one target to another.
+// If the artifact is a manifest list or index, referrers of its manifests are copied as well.
+func recursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.Target, dstRef string, root ocispec.Descriptor, opts oras.ExtendedCopyOptions) error {
+	var err error
+	if root.MediaType == ocispec.MediaTypeImageIndex || root.MediaType == docker.MediaTypeManifestList {
+		var fetched []byte
```
shizhMSFT, line 193: nit:
```suggestion
	if root.MediaType == ocispec.MediaTypeImageIndex || root.MediaType == docker.MediaTypeManifestList {
```

cmd/oras/root/cp.go
```diff
@@ -191,3 +184,48 @@ func runCopy(ctx context.Context, opts copyOptions) error {
...
+		// point referrers of child manifests to root
+		findPredecessor := opts.FindPredecessors
+		var referrers []ocispec.Descriptor
+		for _, desc := range index.Manifests {
+			descs, err := findPredecessor(ctx, src, desc)
+			if err != nil {
+				return err
+			}
+			referrers = append(referrers, descs...)
+		}
```
shizhMSFT, line 212: Can we do it in parallel?

qweeah: refactored with errgroup

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1148, opened 2023-10-23
feat: support progress output in pull

qweeah:
This PR adds progress output for `oras pull`


cmd/oras/root/pull.go
```diff
@@ -105,30 +106,77 @@ Example - Pull artifact files from an OCI layout archive 'layout.tar':
...
+	}
+	if pulledEmpty {
+		fmt.Println("Downloaded empty artifact")
+	}
+	fmt.Println("Pulled", opts.AnnotatedReference())
+	fmt.Println("Digest:", desc.Digest)
+	return nil
+}
+
+func (po *pullOptions) doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget, opts oras.CopyOptions) (ocispec.Descriptor, bool, error) {
```
shizhMSFT, line 148: `doPull` should be a standalone method and should not be a method of `pullOptions`.

cmd/oras/root/pull.go
```diff
@@ -105,30 +106,77 @@ Example - Pull artifact files from an OCI layout archive 'layout.tar':
...
+
+	var tracked track.GraphTarget
+	if po.TTY != nil {
+		tracked, err = track.NewTarget(dst, "Downloading", "Downloaded ", po.TTY)
+		if err != nil {
+			return ocispec.Descriptor{}, false, err
+		}
+		defer tracked.Close()
+		dst = tracked
 	}
```
shizhMSFT, line 166: nit: Like `getTrackedTarget`, it seems can also be refactored although it is invoked only once.

cmd/oras/root/pull.go
```diff
@@ -156,13 +156,9 @@ func (po *pullOptions) doPull(ctx context.Context, src oras.ReadOnlyTarget, dst
...
 
 	var tracked track.GraphTarget
-	if po.TTY != nil {
-		tracked, err = track.NewTarget(dst, "Downloading", "Downloaded ", po.TTY)
-		if err != nil {
-			return ocispec.Descriptor{}, false, err
-		}
-		defer tracked.Close()
```
shizhMSFT: The equivalence of `defer tracked.Close()` is not reflected in the modified code. This is also not observed in `attach` and `push`.

qweeah: `doPush` will handle the closing for `attach` and `push`

shizhMSFT (approved):
LGTM

PR #1157, opened 2023-10-31
feat: remove empty artifact CLI message and show hints for pulling OCI image layout

sajayantony:
Update the message for oras pull of an image to something actionable - 

The current outputs looks like this 
```
❯ oras pull localhost:5000/hello-world:latest
Downloaded empty artifact
Pulled [registry] localhost:5000/hello-world:latest
Digest: sha256:7e9b6e7ba2842c91cf49f3e214d04a7a496f8214356f41d81a6e6dcad11f11e3
```

The PR modifies this to 
```
❯ go run ./cmd/oras  pull localhost:5000/hello-world:latest
Skipped pulling layers without "org.opencontainers.image.title"
Use 'oras copy localhost:5000/hello-world:latest --to-oci-layout <layout-dir>' to pull all layers.

``` 

Also changed to `✓` character for checkmark.

### DIFF
--- a/cmd/oras/root/pull.go
+++ b/cmd/oras/root/pull.go
@@ -130,40 +130,44 @@ func runPull(ctx context.Context, opts pullOptions) error {
 	dst.AllowPathTraversalOnWrite = opts.PathTraversal
 	dst.DisableOverwrite = opts.KeepOldFiles
 
-	desc, pulledEmpty, err := doPull(ctx, src, dst, copyOptions, &opts)
+	desc, skippedLayers, err := doPull(ctx, src, dst, copyOptions, &opts)
 	if err != nil {
 		if errors.Is(err, file.ErrPathTraversalDisallowed) {
 			err = fmt.Errorf("%s: %w", "use flag --allow-path-traversal to allow insecurely pulling files outside of working directory", err)
 		}
 		return err
 	}
-	if pulledEmpty {
-		fmt.Println("Downloaded empty artifact")
+
+	// suggest oras copy for pulling layers without annotation
+	if skippedLayers > 0 {
+		fmt.Printf("Skipped pulling layers without file name in %q\n", ocispec.AnnotationTitle)
+		fmt.Printf("Use 'oras copy %s --to-oci-layout <layout-dir>' to pull all layers.\n", opts.RawReference)
+	} else {
+		fmt.Println("Pulled", opts.AnnotatedReference())
+		fmt.Println("Digest:", desc.Digest)
 	}
-	fmt.Println("Pulled", opts.AnnotatedReference())
-	fmt.Println("Digest:", desc.Digest)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


sajayantony:
Open to rewording this 
> Skipped pulling layers without "org.opencontainers.image.title"

cmd/oras/root/pull.go
```diff
@@ -137,11 +137,15 @@ func runPull(ctx context.Context, opts pullOptions) error {
 		}
 		return err
 	}
+
+	// suggest oras copy for pulling layers without annotation
 	if pulledEmpty {
-		fmt.Println("Downloaded empty artifact")
+		fmt.Printf("Skipped pulling layers without %q\n", ocispec.AnnotationTitle)
+		fmt.Printf("Use 'oras copy %s --to-oci-layout <layout-dir>' to pull all layers.\n", opts.RawReference)
```
qweeah, line 144: Currently `pulledEmpty == true` means an artifact **only** contains unnamed layers. What if an artifact contains both named and unnamed layers? Should we also output this error message?

sajayantony: I would revisit it when those start happening right now the error is confusing. What is an empty artifact from the point of CLI. The variable name should also be something like skippedLayer.

sajayantony: Ok fixed the variable name to count the skipped layers. 

image

The manifest had partial number of layers 

```json
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 1470,
      "digest": "sha256:9c7a54a9a43cca047013b82af109fe963fde787f63f9e016fdc3384500c2823d"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 2457,
         "digest": "sha256:719385e32844401d57ecfd3eacab360bf551a1491c05b85806ed8f1b08d792f6",
         "annotations" : {
                "org.opencontainers.image.title" : "thelayer"
            }
      }
   ]
}
```

cmd/oras/root/pull.go
```diff
@@ -130,40 +130,44 @@ func runPull(ctx context.Context, opts pullOptions) error {
...
 		configPath, configMediaType, err = fileref.Parse(po.ManifestConfigRef, "")
 		if err != nil {
-			return ocispec.Descriptor{}, false, err
+			return ocispec.Descriptor{}, 0, err
 		}
 	}
 
 	var tracked track.GraphTarget
-	dst, tracked, err = getTrackedTarget(dst, po.TTY, "Downloading", "Downloaded ")
+	dst, tracked, err = getTrackedTarget(dst, po.TTY, "Downloading", "Pulled")
```
qweeah, line 163: Better make output aligned to `Skipped    `
```suggestion
	dst, tracked, err = getTrackedTarget(dst, po.TTY, "Downloading", "Pulled     ")
```

qweeah (approved):
LGTM with suggestions:
Action prompt `Pulled` can be formatted into the same length with `Skipped`
image to make output aligned.

Issue #166, opened 2020-06-12
[feature request] Provide python SDK


gaocegege:
Docker provides both Golang and Python SDK here https://docs.docker.com/engine/api/sdk/

If we can provide python sdk too for the custom artifact, I think we can expand the audience.

mirdaki:
This would help out a project I am going to work on soon! Hope a +1 might get it on a roadmap or what not.

swiftdiaries:
+1 on this feature.
I would love to help get this done. I'm not really sure how this would look like. How does Docker manage the client SDKs?
Issue for reference: https://github.com/mlperf/mlbox/issues/89

SteveLasker:
I'd imagine this as another repo `ORAS-python`, so the python sdk could be built and versioned independently, but happy to support the work with a set of capabilities we focus upon.

vsoch:
In case it's helpful, I just finished a first draft of a distribution spec Python client, "Reggie Python" in the [opencontainers](https://vsoch.github.io/oci-python/docs/getting-started#distribution-specification) distribution module to mirror the GoLang counterpart, [Reggie](https://github.com/bloodorangeio/reggie) inspired by @jdolitsky. Has anyone started working on ORAS-python, or need help? I really enjoy this stuffs so I can offer to help, but I'd probably need pointers since I'm familiar with the distribution spec but not oras specifically.

SteveLasker:
If anyone would like to implement ORAS in Python, we'd be happy to adopt it under ORAS. If so, please reach out and we can add it to the ORAS family. We'd love to make ORAS multi-lingual.

vsoch:
@SteveLasker I'd really enjoy doing this as a weekend project, seems up my alley :) Is there any documentation or discussion to be had, or just point at this repository and say "Make it Python!" I could probably do that over the weekends as a fun project.

SteveLasker:
Thanks @vsoch,
We're about to do a split of the libraries and binaries #181, so I'd suggest tracking that item and once complete you should have what you need to do a python version. Or, this will help us understand what we need to doc, so you could write a python version.

Steve

gaocegege:
Hi, is there any progress here? I am still waiting the upstream python SDK to support https://github.com/kleveross/ormb

Or I can help implement such an SDK for the community.

vsoch:
I started but I haven't been able to make the time, so please take the reins if you can! https://github.com/vsoch/oras-python

gaocegege:
Cool!

[...]

PR #1163, opened 2023-11-02
add error message guideline

### LINKED ISSUE #1158: Create a CLI Error Handling Guideline to improve error messages and error handling

1.1.0

The error messages in some ORAS CLI commands do not show the actual value that were provided or show too vague error messages that are hard to troubleshoot and locate the root cause. ORAS maintainers are working on improving some error messages, such as #1152 , #1156 , #1169 .

In order to provide clearer structures for error categorizing and also more actionable error outputs for better user experience, we need to create an ORAS CLI Error Handling Guideline. 

This document aims to provide the guidelines for ORAS contributors to improve existing error messages and handling method as well as the new error output format. It will also provide recommendations for ORAS CLI contributors for how to write friendly and standard error messages and avoid inconsistent error handling & messages.

FeynmanZhou:

add CLI error handling guide

Fixes #1158 


TerryHowe:
Are we going to link this file from https://oras.land/docs/community/developer_guide

FeynmanZhou:
> Are we going to link this file from https://oras.land/docs/community/developer_guide

Yes. We could add this doc link to that guide after it is merged.

sajayantony:
The examples and patterns are really helpful @FeynmanZhou. Thanks for putting this together.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+
+Forth, signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.
+
+Last, error logs can also be useful for post-mortem debugging but make sure they have timestamps, truncate them occasionally so they don’t eat up space on disk, and make sure they don’t contain ansi color codes. Thereby, error logs can be written to a file.
+
+## Error output recommendation
+
+### Dos
+
+- Use the capital letter ahead of an error message
```
qweeah, line 29: Can you be more specific about what is an error message? Can it be multiple line? Maybe `Use the capital letter ahead of each line of any error message`?

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+- Print human readable error message. If the error message is mainly from the server and varies by different servers, tell users that the error response is from server. This implies that users may need to contact server side for troubleshooting.
+- Provide specific and actionable prompt message with argument suggestion or show the example usage for reference. (e.g, Instead of showing showing flag or argument options is missing, please provide available argument options and guide users to "--help" to view more examples)
```
qweeah, line 31: Good point on how to be information. How about adding a new bullet point here on how to be descriptive? E.g. Provide full description if the user input does not match what CLI is expecting. A full description should include 1) what input is received and 2) what input is expected.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Don'Ts
+
+- Do not use a formula-like or a programming expression in the error message. (e.g, `json: cannot unmarshal string into Go value of type map[string]map[string]string.`, or `Parameter 'xyz' must conform to the following pattern: '^[-\\w\\._\\(\\)]+$'`)
+- Do not use ambiguous expressions which mean nothing to users. (e.g, `Something unexpected happens`, or `Error: accepts 2 arg(s), received 0`)
+- Do not print irrelevant error message to make the output noisy. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong.
+
+## How to write friendly error message
+
+### Recommended error meesage structure
```
qweeah, line 44:
```suggestion
### Recommended error message structure
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+```
+$ oras manifest push --oci-layout /tmp/ginkgo2167255592:mediatype-flag
+Error: media type is not recognized. 
+```
+
+Suggested error message:
+
+```
+$ oras manifest push --oci-layout /tmp/ginkgo2167255592:mediatype-flag
+Error: media type is not recognized. Specify an valid media type with "--media-type"
```
qweeah, line 129: `recognized` is too general, the actual action is: ORAS CLI tried to found a valid `mediaType` in the manifest but failed. Any idea on how to make it less general and more descriptive?
```

File: docs/proposals/error-handling-guideline.md [hunk 5]
```diff
@@ -0,0 +1,225 @@
...
+
+- Do not use a formula-like or a programming expression in the error message. (e.g, `json: cannot unmarshal string into Go value of type map[string]map[string]string.`, or `Parameter 'xyz' must conform to the following pattern: '^[-\\w\\._\\(\\)]+$'`)
+- Do not use ambiguous expressions which mean nothing to users. (e.g, `Something unexpected happens`, or `Error: accepts 2 arg(s), received 0`)
+- Do not print irrelevant error message to make the output noisy. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong.
+
+## How to write friendly error message
+
+### Recommended error meesage structure
+
+Here is a sample structure of an error message:
```
qweeah, line 46: Should all error message be output to stderr? Currently it is.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+
+```text
+Error: [Error description] : [Error code] ：[Error title]
+
```
qweeah, line 50: Just wondering, can we not put a blank line here? The prefix has already made each section clear.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+Usage: [Sample command]
```
qweeah, line 51: It's not sample command but a descriptive syntax of the command usage.
```suggestion
Usage: [command usage]
```

docs/proposals/error-handling-guideline.md
qweeah: ` Specify an valid media type with "--media-type"` Should we put this to `Help:`?

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+```
+$ oras attach --artifact-type oras/test localhost:7000/command/images:foobar --distribution-spec ???
+Error: unknown distribution specification flag: "???"
+```
+
+Suggested error message:
+
+```
+$ oras attach --artifact-type oras/test localhost:7000/command/images:foobar --distribution-spec ???
+Error: unknown distribution specification flag: "???". Available options: v1.1-referrers-api, v1.1-referrers-tag
```
qweeah, line 145: `Available options: v1.1-referrers-api, v1.1-referrers-tag` Should we put this to `Help:`?

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+Suggested error message:
+
+```
+$ oras attach --artifact-type oras/test /tmp/ginkgo2977244222:foobar >>
+Error: no blob or manifest annotation are provided
+
+Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<type>] [...]
+Help: Attach files to an existing artifact. Run "oras attach" for more options and examples
```
qweeah, line 164:
```suggestion
Help: Attach files to an existing artifact. Run "oras attach -h" for more options and examples
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+Error: failed to resolve latest: GET "https://registry-1.docker.io/v2/nginx/manifests/latest": response status code 401: unauthorized: authentication required: [map[Action:pull Class: Name:nginx Type:repository]]
+```
+
+Suggested error message:
+
+```
+$ oras pull ghcr.io/nginx:latest
+Error: failed to resolve the resource from server: pull access denied for ghcr.io/nginx:latest : response status code 401: unauthorized: requested access to the resource is denied
+
+Help: repository does not exist or namespace is missing, or may require 'oras login'.
```
qweeah, line 216: To unify the format
```suggestion
Help: Namespace is missing in `/nginx`, or run 'oras login ghcr.io' and provide a token with pull access.
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+
+```
+$ oras pull docker.io/nginx:latest
+Error: failed to resolve latest: GET "https://registry-1.docker.io/v2/nginx/manifests/latest": response status code 401: unauthorized: authentication required: [map[Action:pull Class: Name:nginx Type:repository]]
+```
+
+Suggested error message:
+
+```
+$ oras pull ghcr.io/nginx:latest
```
qweeah, line 213: The registry is changed from `docker.io` to `ghcr.io`?
```suggestion
$ oras pull docker.io/nginx:latest
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Recommended error meesage structure
+
+Here is a sample structure of an error message:
+
+```text
+Error: [Error description] : [Error code] ：[Error title]
+
+Usage: [Sample command]
+Help: [Recommended solution], [TSG]
```
qweeah, line 52: IMHO we can skip the `Help: ` prefix. It's the default behavior of cobra and should be reused by most cobra-based CLIs.

qweeah, line 52:
```suggestion
[Recommended solution], [TSG]
```

docs/proposals/error-handling-guideline.md
TerryHowe: stderr should be used IMO

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+- Error title
+- Error description 
+- How to fix the error
+- URL for more information (Optional, it can be a TSG document link)
+
+Second, when necessary, it is highly suggested for ORAS CLI contributors to provide recommendations for users how to resolve the problems based on the error messages they encountered. Showing descriptive words and straightforward prompt with executable commands as a potential solution is a good practice for error messages.
+
+Third, for unhandled errors you didn’t expect the user to run into. For that, have a way to view full traceback information as well as full debug or verbose logs output, and instructions on how to submit a bug.
+
+Forth, signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.
```
TerryHowe, line 21:
```suggestion
Fourth, signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.
```

docs/proposals/error-handling-guideline.md
TerryHowe: Maybe I'm not understanding title and description, but I would expect something along the lines of `Error: code: title: description`

TerryHowe:
sembr helps reviews on this stuff https://sembr.org/

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+This document aims to provide the guidelines for ORAS contributors to improve existing error messages and error handling method as well as the new error output format. It will also provide recommendations for ORAS CLI contributors for how to write friendly and standard error messages and avoid inconsistent error handling & messages.
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
+
+First and foremost, make the error messages descriptive and informative. Error messages are expected to be helpful to troubleshoot where the user has done something wrong and the program is guiding them in the right direction. A great error message is recommended to contain the following:
+
+- Error code
+- Error title
```
yizha1, line 12: What distinguishes a title from a description? Perhaps we can provide greater clarity. While both the title and description may not always be informative, we can emphasize that they should convey what happened and why it happened.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+- Error description 
+- How to fix the error
+- URL for more information (Optional, it can be a TSG document link)
```
yizha1, line 15: We can put a guideline on using a URL that it will not likely be changed. We met this problem in Notation that the URL changed.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+
+This document aims to provide the guidelines for ORAS contributors to improve existing error messages and error handling method as well as the new error output format. It will also provide recommendations for ORAS CLI contributors for how to write friendly and standard error messages and avoid inconsistent error handling & messages.
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
+
+First and foremost, make the error messages descriptive and informative. Error messages are expected to be helpful to troubleshoot where the user has done something wrong and the program is guiding them in the right direction. A great error message is recommended to contain the following:
+
+- Error code
```
yizha1, line 11: IMO, Error code should be optional

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
+# ORAS CLI Error Handling and Message Guidelines
+
+This document aims to provide the guidelines for ORAS contributors to improve existing error messages and error handling method as well as the new error output format. It will also provide recommendations for ORAS CLI contributors for how to write friendly and standard error messages and avoid inconsistent error handling & messages.
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
```
yizha1, line 7: Maybe we can show a concrete example of good error message, which helps users to understand and is also useful when producing new error messages.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,225 @@
...
+Error: no blob or manifest annotation are provided
+```
+
+Suggested error message:
+
+```
+$ oras attach --artifact-type oras/test /tmp/ginkgo2977244222:foobar >>
+Error: no blob or manifest annotation are provided
+
+Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<type>] [...]
```
yizha1, line 163: The usage is too broad, and not helpful enough, maybe we can show an example for attaching files, and run `oras attach` for more usage and more examples.

docs/proposals/error-handling-guideline.md
yizha1: It’s also worth mentioning that error messages will be printed to a specific I/O as a guideline.

yizha1 (commented):
Thanks @FeynmanZhou I left some comments.

docs/proposals/error-handling-guideline.md
FeynmanZhou: In general, logs from server side contain the error code, such as https://github.com/oras-project/oras/pull/1163/files#diff-87978b235bb041f07bf2e71966c261dd6103d718e95ebb9c5c984bebbaf5d357R207

docs/proposals/error-handling-guideline.md
FeynmanZhou: Yes, That's being said we should use a versioned document link in error message

docs/proposals/error-handling-guideline.md
FeynmanZhou: There are 9 examples at https://github.com/oras-project/oras/pull/1163/files#diff-87978b235bb041f07bf2e71966c261dd6103d718e95ebb9c5c984bebbaf5d357R55

docs/proposals/error-handling-guideline.md
FeynmanZhou: SGTM

docs/proposals/error-handling-guideline.md
FeynmanZhou: Is there a doc that lists all available "media type" as a reference for users? Maybe we could provide a doc link to users.

docs/proposals/error-handling-guideline.md
FeynmanZhou: All error message are expected to output to stderr.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+```
+$ oras attach --artifact-type oras/test /tmp/ginkgo2977244222:foobar >>
+Error: no blob is provided
+```
+
+Suggested error message:
+
+```
+$ oras attach --artifact-type oras/test /tmp/ginkgo2977244222:foobar >>
+Error: no blob is provided
```
toddysm, line 161: This is still not helpful. What does
> no blob is provided 

mean? The error can be more descriptive. Something like:
> Failed to redirect output. Not output file provided in the command.

Also, without listing the full command in the logs it will be hard for the user to troubleshoot.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+Suggested error message:
+
+```
+$ oras pull docker.io/nginx:latest
+Error: failed to resolve the resource from server: pull access denied for ghcr.io/nginx:latest : response status code 401: unauthorized: requested access to the resource is denied
+Namespace is missing, do you mean `oras pull docker.io/library/nginx:latest`? 
+```
+
```
toddysm, line 214: In the above examples, some of the error messages have suggestions while other not. For consistency we should make sure that each error message has a suggestion.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+## Reference
```
toddysm, line 215: One thing that is not covered in this spec is to specify which errors will output error code 0 (automation will continue) and which will output something different from 0 (automation will break). In my experience, ORAS returns error codes that break automation when they shouldn't. The typical example is when manifest index delete fails for OCI 1.0 references.

toddysm (changes requested):

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
+# ORAS CLI Error Handling and Message Guidelines
+
+This document aims to provide the guidelines for ORAS contributors to improve existing error messages and error handling method as well as the new error output format. It will also provide recommendations and examples for ORAS CLI contributors for how to write friendly and standard error messages, avoid generating inconsistent and ambiguous error messages.
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
+
+First and foremost, make the error messages descriptive and informative. Error messages are expected to be helpful to troubleshoot where the user has done something wrong and the program is guiding them in the right direction. A great error message is recommended to contain the following elements:
+
```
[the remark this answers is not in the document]
FeynmanZhou: Agree. Error code is optional and is generally from the server side. So in some cases error code is also necessary to be showed in the error message.
Splitting "What the error is" and "How to solve it" into two lines looks much straightforward for users. If you look at Docker, Docker error messages are also following this style.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Good point. I will investigate current behavior of ORAS, what is the right approach to return error code 0, and when to break automation. Then add a paragraph for this part

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+- Error code (When the logs are generated from the server side)
```
shizhMSFT, line 11: What's a sample of error code? Is it a status code?

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+- Do not print irrelevant error message to make the output noisy. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong.
+
+## How to write friendly error message
+
+### Recommended error message structure
+
+Here is a sample structure of an error message:
+
+```text
+Error: [Error code] : [Error title] : [Error description]
```
shizhMSFT, line 49: nit: convention
```suggestion
Error: [Error code]: [Error title]: [Error description] 
```

docs/proposals/error-handling-guideline.md
FeynmanZhou: Yes, it's status code.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+Fourth, signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.
+
+Last, error logs can also be useful for post-mortem debugging but make sure they have timestamps, truncate them occasionally so they don’t eat up space on disk, and make sure they don’t contain ansi color codes. Thereby, error logs can be written to a file.
+
+## Error output recommendation
+
+### Dos
+
+- Use the capital letter ahead of each line of any error message
+- Print human readable error message. If the error message is mainly from the server and varies by different servers, tell users that the error response is from server. This implies that users may need to contact server side for troubleshooting.
```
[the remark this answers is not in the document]
FeynmanZhou: The purpose is to distinguish whether this error is from the client side or server (registry) side. If it is explicit that the error is from the server side, users would know that they may need to ask for support on the registry side.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,221 @@
...
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: unable to add tag for 'list': invalid reference: missing repository
+```
+
+Suggested error message:
+
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: unable to add tag for 'list': invalid reference: missing repository
```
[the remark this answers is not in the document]
FeynmanZhou: Updated

docs/proposals/error-handling-guideline.md
qweeah: > Is there a doc that lists all available "media type" as a reference for users?

The general media type should follow the description in https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties.

Also, the problem here is `media type should be specified in the manifest or via flag but it's not found`, not `provided media type is invalid`.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+Error: media type is not recognized. 
+```
+
+Suggested error message:
+
+```
+$ oras manifest push --oci-layout /sample/images:foobar:mediatype
+Error: media type is not recognized. 
+Usage: oras manifest push [flags] <name>[:<tag>[,<tag>][...]|@<digest>] <file>
+Specify an valid media type with the "--media-type" flag.
```
qweeah, line 131:
```suggestion
You need to specify a valid media type in the manifest JSON or via the "--media-type" flag.
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+Error: unable to add tag for 'list': invalid reference: missing repository
+```
+
+Suggested error message:
+
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: There is no "list" sub-command for "oras tag" command.
+Usage: oras tag [flags] <name>{:<tag>|@<digest>} <new_tag> [...]
+Tag a manifest in a registry or an OCI image layout. Run "oras tag -h" for more options and examples
```
qweeah, line 95: Should we suggest user to use `oras repo show-tags` in this case? `Tag a manifest in a registry or an OCI image layout.` doesn't sound actionable.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+Error: "/tmp/ginkgo1163328512": no tag or digest when expecting <name:tag|name@digest>
+```
+
+Suggested error message:
+
+```
+$ oras manifest fetch --oci-layout /tmp/ginkgo1163328512
+Error: "/tmp/ginkgo1163328512": no tag or digest specified
+Usage: oras manifest fetch [flags] <name>{:<tag>|@<digest>}
+Fetch manifest of the target artifact. Run "oras manifest fetch -h" for more options and examples
```
qweeah, line 113: `Fetch manifest of the target artifact.` does not sound actionable. How about `You need to specify an artifact reference in form of name:tag or name@digest`

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+Error: accepts 2 arg(s), received 0
+```
+
+Suggested error message:
+
+```
+$ oras cp
+Error: "oras copy" requires exactly 2 arguments.
+Usage: oras copy [flags] <from>{:<tag>|@<digest>} <to>[:<tag>[,<tag>][...]]
+Copy artifacts from one target to another. Run "oras copy -h" for more options and examples
```
qweeah, line 77: `Copy artifacts from one target to another.`  does not sound actionable. How about `You need to specify two arguments as source and destination for copying.`

docs/proposals/error-handling-guideline.md
FeynmanZhou: How about: 
If you want to tag a manifest in a registry or an OCI image layout, use "oras tag". If you want to list available tags in a repository, use "oras repo show tags"

qweeah (approved):
LGTM

docs/proposals/error-handling-guideline.md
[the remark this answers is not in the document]
toddysm: How will the error title be used? Giving a usage example would clarify this.

docs/proposals/error-handling-guideline.md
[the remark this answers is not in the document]
toddysm: How will this information will be shown in the error message? How will the customer know that this is a server error or client error?

docs/proposals/error-handling-guideline.md
toddysm: This got resolved without any comments.

docs/proposals/error-handling-guideline.md
toddysm: I think my ask here is different. What you point me to is what I explained above with 0 and 1 Exist Status. What I am asking is to have a guidance what type of errors will return `EXIT_SUCCESS` and `EXIT_FAILURE`. I am envisioning one specific use case where DELETE of the manifest index fails but the reference is pushed and from customer point of view this is `EXIT_SUCCESS` but we return `EXIT_FAILURE`

toddysm (changes requested):

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
+
+First and foremost, make the error messages descriptive and informative. Error messages are expected to be helpful to troubleshoot where the user has done something wrong and the program is guiding them in the right direction. A great error message is recommended to contain the following elements:
+
+- Error code: optional, when the logs are generated from the server side
+- Error description: describe what the error is
+- Suggestion: How to fix the error. Versioned troubleshooting document link is nice to have.
```
shizhMSFT, line 13: Not all errors can be fixed.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+
+Second, when necessary, it is highly suggested for ORAS CLI contributors to provide recommendations for users how to resolve the problems based on the error messages they encountered. Showing descriptive words and straightforward prompt with executable commands as a potential solution is a good practice for error messages.
+
+Third, for unhandled errors you didn’t expect the user to run into. For that, have a way to view full traceback information as well as full debug or verbose logs output, and instructions on how to submit a bug.
```
shizhMSFT, line 17: nit: use ASCII chars
```suggestion
Third, for unhandled errors you didn't expect the user to run into. For that, have a way to view full traceback information as well as full debug or verbose logs output, and instructions on how to submit a bug.
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+
+Fourth, signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it’s going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.
+
+Fifth, CLI program termination should follow the standard [Exit Status conventions](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) to report execution status information about success or failure. 
+
+Last, error logs can also be useful for post-mortem debugging but make sure they have timestamps, truncate them occasionally so they don’t eat up space on disk, and make sure they don’t contain ansi color codes. Thereby, error logs can be written to a file.
```
shizhMSFT, line 23: Timestamps by apps are not necessary as upper logging system will provide.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+## How to write friendly error message
+
+### Recommended error message structure
+
+Here is a sample structure of an error message:
+
+```text
+Error:[Status code]: [Error description] 
+Usage: [Command usage]
+[Recommended solution]
```
shizhMSFT, line 51:
```suggestion
Error: [{Status code}:] {Error description}
[Usage: {Command usage}]
[{Recommended solution}]
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+- Status code is an optional information. If the error message is generated from the server side, it may include error code or [warn code](https://www.rfc-editor.org/rfc/rfc7234#section-5.5). It could be printed out alongside the error description.
```
shizhMSFT, line 54: Warn code should be prefixed with `Warning:`, not `Error:`. Additionally, warn code is fixed as defined in the [distribution-spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#warnings).

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+- Command usage is also an optional information but it's recommended to be printed out when user input doesn't follow the standard usage or examples. 
+- Recommended solution is required and should follow the general guiding principles described above.
```
shizhMSFT, line 56: Not all errors are fixable / resolvable. Those errors are considered as fatal errors or unknown errors.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Sorry for missing reply to this comment. 

This usage is for general reference if users don't follow the standard form of the command usage. It was defined in the [recommended error message structure](https://github.com/oras-project/oras/pull/1163/files#diff-87978b235bb041f07bf2e71966c261dd6103d718e95ebb9c5c984bebbaf5d357R94).

The detailed prompt message is showed at [line #166](image

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+Error: no blob is provided
+```
+
+Suggested error message:
+
+```
+$ oras attach --artifact-type sbom/example localhost:5000/sample/images:foobar
+Error: failed to attach a file. No file provided in the command.
+Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<type>] [...]
+To attach files to an existing artifact, try "oras attach --artifact-type sbom/example localhost:5000/sample/images:foobar sample.json". Run "oras attach -h" for more options and examples
```
qweeah, line 166: User need to provide at least a file or an annotation for attaching. The proper error message should be:
```suggestion
$ oras attach --artifact-type sbom/example localhost:5000/sample/images:foobar
Error: no file or annotation provided in the command.
Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<type>] [...]
To attach to an existing artifact, please provide files via argument or annotations via flag `-a/--annotation`. Run "oras attach -h" for more options and examples.
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+Error: accepts 2 arg(s), received 0
+```
+
+Suggested error message:
+
+```
+$ oras cp
+Error: "oras copy" requires exactly 2 arguments.
+Usage: oras copy [flags] <from>{:<tag>|@<digest>} <to>[:<tag>[,<tag>][...]]
+You need to specify two arguments as source and destination respectively. Run "oras copy -h" for more options and examples
```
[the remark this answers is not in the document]
FeynmanZhou: I don't find there is a general style guide that answers this question. Even though in Git, we also see it uses "you", "your" in its output messages. My suggestion is that ORAS should try to avoid using [personal pronouns](https://en.wikipedia.org/wiki/Personal_pronoun) but this is still allowed in some cases, such as questioning the CLI user.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Removed "Timestamps"

docs/proposals/error-handling-guideline.md
FeynmanZhou: Updated

docs/proposals/error-handling-guideline.md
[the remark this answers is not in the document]
FeynmanZhou: No. `Help:` is not necessary here because this message is obviously a help message

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: unable to add tag for 'list': invalid reference: missing repository
+```
+
+Suggested error message:
+
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: There is no "list" sub-command for "oras tag" command.
```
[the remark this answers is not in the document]
FeynmanZhou: by exact match. This is a certain case.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
+Error: annotation value doesn't match the required format.
+Try `oras push --annotation "key=value" ghcr.io/library/alpine:3.9`  
+```
+
+#### Example 9: when failed to pull files from a public registry
+
+```
+$ oras pull docker.io/nginx:latest
+Error: failed to resolve latest: GET "https://registry-1.docker.io/v2/nginx/manifests/latest": response status code 401: unauthorized: authentication required: [map[Action:pull Class: Name:nginx Type:repository]]
```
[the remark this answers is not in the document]
FeynmanZhou: Updated.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Updated

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+
+```
+$ oras attach --artifact-type sbom/example localhost:5000/sample/images:foobar
+Error: no blob is provided
+```
+
+Suggested error message:
+
+```
+$ oras attach --artifact-type sbom/example localhost:5000/sample/images:foobar
```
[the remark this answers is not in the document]
FeynmanZhou: Add annotations as one of the error cases

docs/proposals/error-handling-guideline.md
FeynmanZhou: Removed `Warning`

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+```
+$ oras cp
+Error: accepts 2 arg(s), received 0
+```
+
+Suggested error message:
+
+```
+$ oras cp
+Error: "oras copy" requires exactly 2 arguments.
```
qweeah, line 75: What is received is not mentioned, suggest to add number of arguments received
```console
$ oras cp localhost:5000/ test:from localhost:5000/test:to
Error: "oras cp" requires exactly 2 arguments but got 3
Usage: oras cp [flags] <from>{:<tag>|@<digest>} <to>[:<tag>[,<tag>][...]]
Please specify exactly 2 arguments as the source and destination for copying. Run "oras cp -h" for more options and examples

$ oras cp
Error: "oras cp" requires exactly 2 arguments but got 0
Usage: oras cp [flags] <from>{:<tag>|@<digest>} <to>[:<tag>[,<tag>][...]]
Please specify exactly 2 arguments as the source and destination for copying. Run "oras cp -h" for more options and examples
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+```
+$ oras push --annotation-file sbom.json ghcr.io/library/alpine:3.9
+Error: failed to load annotations from sbom.json: json: cannot unmarshal string into Go value of type map[string]map[string]string. Please refer to the document at https://oras.land/docs/how_to_guides/manifest_annotations
+```
+
+Suggested error message:
+
+```
+$ oras push --annotation-file annotation.json ghcr.io/library/alpine:3.9
+Error response from registry: failed to load annotations from sbom.json: json: cannot unmarshal string into Go value of type map[string]map[string]string.
```
qweeah, line 182: This is not a registry error, suggest to change the error description to:
```suggestion
Error: invalid manifest json file: failed to load annotations from sbom.json
```

docs/proposals/error-handling-guideline.md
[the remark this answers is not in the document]
FeynmanZhou: I removed "error title" because this is a bit redundant. A concise and brief error description is enough.
 
The recommended error message structure is listed at [line49](https://github.com/oras-project/oras/pull/1163/files#diff-87978b235bb041f07bf2e71966c261dd6103d718e95ebb9c5c984bebbaf5d357R49).

docs/proposals/error-handling-guideline.md
FeynmanZhou: I remember `json: cannot unmarshal string into Go value of type map[string]map[string]string` is a registry error message. Can we remove it?

docs/proposals/error-handling-guideline.md
FeynmanZhou: @toddysm 
All types of errors will be in stderr and should return `EXIT_FAILURE`, the automation should also be terminated.

For that specific case, I remember it was reproduced in `oras attach` when use ORAS with registries that don't support Deletion API (e.g. Docker Hub). It was a bug and has been fixed. The reference will not be attached successfully and `oras attach` will fail in that case.

@shizhMSFT can correct me if I am wrong.

docs/proposals/error-handling-guideline.md
qweeah: In this case it's not a registry error. Here annotation file is a user input and JSOM unmarshal goes wrong based on the user input。

docs/proposals/error-handling-guideline.md
shizhMSFT: Here's the definition: The `oras` CLI returns `EXIT_FAILURE` if and only if the `oras` CLI reports one or more errors.

In the case of returning an "error" with `EXIT_SUCCESS`, that "error" is considered as a warning and shall be corrected as a warning.

---

> I am envisioning one specific use case where DELETE of the manifest index fails but the reference is pushed and from customer point of view this is `EXIT_SUCCESS` but we return `EXIT_FAILURE`

In the above case the deletion failed shall be corrected (if not corrected already) to a warning instead of an error, and thus the `oras` CLI returns `EXIT_SUCCESS`.

docs/proposals/error-handling-guideline.md
```diff
@@ -148,7 +148,7 @@ Error: unknown distribution specification flag: "v1.0".
 Available options: v1.1-referrers-api, v1.1-referrers-tag
 ```
 
-#### Example 6: when attaching an file, if no file reference is provided
+#### Example 6: when attaching an file, if no file reference or annotation is provided
```
shizhMSFT, line 151:
```suggestion
#### Example 6: when attaching, if neither file reference nor annotation is provided
```

docs/proposals/error-handling-guideline.md
```diff
@@ -161,9 +161,9 @@ Suggested error message:
 
 ```
 $ oras attach --artifact-type sbom/example localhost:5000/sample/images:foobar
-Error: failed to attach a file. No file provided in the command.
+Error: no file or annotation provided in the command
```
shizhMSFT, line 164:
```suggestion
Error: neither file nor annotation provided in the command
```

toddysm (approved):

docs/proposals/error-handling-guideline.md
FeynmanZhou: Thanks @shizhMSFT for the clarification. I added your suggested definition to the General guiding principles section.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,224 @@
...
+- Do not print irrelevant error message to make the output noisy. The more irrelevant output you produce, the longer it's going to take the user to figure out what they did wrong.
+
+## How to write friendly error message
+
+### Recommended error message structure
+
+Here is a sample structure of an error message:
+
+```text
+Error: [{Status code}:] {Error description}
```
qweeah, line 49: This structure need to be updated. Firstly, the error prompt can be `Error:` or `Error response from registry:`. Secondly, status code only applies to errors from registry and should be part of the error description.

```suggestion
{Error|Error response from registry}: {Error description}
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+- Do not print irrelevant error message to make the output noisy. The more irrelevant output you produce, the longer it's going to take the user to figure out what they did wrong.
+
+## How to write friendly error message
+
+### Recommended error message structure
+
+Here is a sample structure of an error message:
+
+```text
+Error: {Error|Error response from registry}: {Error description}
```
qweeah, line 49: `{Error|Error response from registry}` means `Error` or `Error response from registry` will be printed out.

```suggestion
{Error|Error response from registry}: {Error description}
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+Error: Head "https:///v2/oras/manifests/sha256:ffa50b27cd0096150c0338779c5090db41ba50d01179d851d68afa50b393c3a3": http: no Host in request URL
+```
+
+Suggested error message:
+
+```
+$ oras push /oras --format json
+Error: neither registry nor OCI image layout provided in the command 
+Usage: oras push [flags] <name>[:<tag>[,<tag>][...]] <file>[:<type>] [...]
+please specify a registry or an OCI image layout folder to push. Run "oras push -h" for more options and examples
```
qweeah, line 233: Since user haven't use `--oci-layout` here, the provided reference `/oras` is expected to be an artifact in a remote registry. To avoid confusion, suggest not to mention OCI layout.
```suggestion
Error: "/oras" is an invalid reference
Usage: oras push [flags] <name>[:<tag>[,<tag>][...]] <file>[:<type>] [...]
Please specify a valid reference in the form of <registry>/<repo>[:tag|@digest]
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+Suggested error message:
+
+```
+$ oras cp
+Error: "oras copy" requires exactly 2 arguments but received 0.
+Usage: oras copy [flags] <from>{:<tag>|@<digest>} <to>[:<tag>[,<tag>][...]]
+Please specify 2 arguments as source and destination respectively. Run "oras copy -h" for more options and examples
+```
+
+#### Example 2: When reference does not match with the expected format
```
qweeah, line 80: This is not an example of `reference does not match with the expected format`, this is a well-known mistake of using `repo show-tags` as `tag list`
```suggestion
#### Example 2: When user mistakenly use `tag list` command instead of `repo show-tags`
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+Suggested error message:
+
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: There is no "list" sub-command for "oras tag" command.
+Usage: oras tag [flags] <name>{:<tag>|@<digest>} <new_tag> [...]
+If you want to tag a manifest in a registry or an OCI image layout, use "oras tag". If you want to list available tags in a repository, use "oras repo show tags" 
+```
+
+#### Example 3: When fetching a manifest if no manifest tag or digest is provided
```
qweeah, line 98:
```suggestion
#### Example 3: No tag or digest provided when fetching a manifest if 
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+Error: media type is not recognized. 
+```
+
+Suggested error message:
+
+```
+$ oras manifest push --oci-layout /sample/images:foobar:mediatype
+Error: media type is not recognized. 
+Usage: oras manifest push [flags] <name>[:<tag>[,<tag>][...]|@<digest>] <file>
+You need to specify a valid media type "mediaType" in the manifest JSON or via the "--media-type" flag.
```
qweeah, line 131:
```suggestion
You need to specify a valid media type in the manifest JSON or via the "--media-type" flag.
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
+Error: missing key in `--annotation` flag: key:value
+```
+
+Suggested error message:
+
+```
+$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
+Error: annotation value doesn't match the required format.
+Try `oras push --annotation "key=value" ghcr.io/library/alpine:3.9`
```
qweeah, line 200: It's impossible to restore the full input command, the best we can do provide a valid annotation value, like: (
also, simple usage need to be provided)
```
$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
Error: annotation value doesn't match the required format.
Usage: oras push [flags] <name>[:<tag>[,<tag>][...]] <file>[:<type>] [...]
Try `--annotation "key=value"`  
```

qweeah (changes requested):

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+Error: PUT "https://registry-1.docker.io/v2/user/demo/manifests/sha256:d2e409ed5674c9cf4fb5c13615288b130610ed893f97c78668ad233dc403976d": response status code 404: notfound: not found
+```
+
+Suggested error message:
+
+```
+$ oras attach docker.io/user/demo:v1 --artifact-type example/sbom sbom.spdx
+Uploading 6cb759c4296e sbom.spdx
+Uploaded  6cb759c4296e sbom.spdx
+Error response from registry: PUT "https://registry-1.docker.io/v2/user/demo/manifests/sha256:d2e409ed5674c9cf4fb5c13615288b130610ed893f97c78668ad233dc403976d": response status code 404: notfound:
```
qweeah, line 269: Should we hide `PUT`, URI and status code from user? Without using `--debug`, user might not care about those details.

If so, suggest change the error to
```suggestion
Error response from registry: this registry doesn't support storing referrers
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+
+Here is a sample structure of an error message:
+
+```text
+Error: {Error|Error response from registry}: {Error description}
+[Usage: {Command usage}]
+[{Recommended solution}]
+```
+
+- Status code is an optional information. Printed out the status code if the error message is generated from the server side.
```
shizhMSFT, line 54: The status code is inconsistent with the above structure.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+Error: unable to add tag for 'list': invalid reference: missing repository
+```
+
+Suggested error message:
+
+```
+$ oras tag list ghcr.io/oras-project/oras
+Error: There is no "list" sub-command for "oras tag" command.
+Usage: oras tag [flags] <name>{:<tag>|@<digest>} <new_tag> [...]
+If you want to tag a manifest in a registry or an OCI image layout, use "oras tag". If you want to list available tags in a repository, use "oras repo show tags"
```
shizhMSFT, line 95:
```suggestion
If you want to tag a manifest in a registry or an OCI image layout, use "oras tag". If you want to list available tags in a repository, use "oras repo show-tags" 
```
or simply `oras repo tags`.

shizhMSFT, line 95: The users clearly typed `oras tag list`. Why do we need to mention `oras tag`? It seems not related.

docs/proposals/error-handling-guideline.md
shizhMSFT: It is not "impossible" but it is not trivial to do so.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+
+#### Example 12: Attach a file to an incompatible registry
+
+Current behavior and output:
+
+```
+$ oras attach docker.io/user/demo:v1 --artifact-type example/sbom sbom.spdx
+Uploading 6cb759c4296e sbom.spdx
+Uploaded  6cb759c4296e sbom.spdx
+Error: PUT "https://registry-1.docker.io/v2/user/demo/manifests/sha256:d2e409ed5674c9cf4fb5c13615288b130610ed893f97c78668ad233dc403976d": response status code 404: notfound: not found
```
shizhMSFT, line 260: This is interesting that dockerhub returns 404 on PUT for not supporting referrers?

docs/proposals/error-handling-guideline.md
FeynmanZhou: I agree. The default error log is supposed to be simple, readable, and actionable. The detailed server-side log could be showed when specifying `--debug`.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Yes. Line260 is the original logs from Docker Hub, but it looks too simple and confusing.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Agree. Removed the first sentence from the prompt information.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,279 @@
...
+
+This document aims to provide the guidelines for ORAS contributors to improve existing error messages and error handling method as well as the new error output format. It will also provide recommendations and examples for ORAS CLI contributors for how to write friendly and standard error messages, avoid generating inconsistent and ambiguous error messages.
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
+
+First and foremost, make the error messages descriptive and informative. Error messages are expected to be helpful to troubleshoot where the user has done something wrong and the program is guiding them in the right direction. A great error message is recommended to contain the following elements:
+
+- Error code: optional, when the logs are generated from the server side
```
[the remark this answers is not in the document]
FeynmanZhou: It should be corrected to status code. I updated line11 as follows:

- Status code:  optional, when the logs are generated from the server side, it's recommended to print the status code in the error description

docs/proposals/error-handling-guideline.md
FeynmanZhou: Updated. Print the correct usage only in the prompt information is enough. I am fine with this format.

docs/proposals/error-handling-guideline.md
```diff
@@ -215,6 +215,61 @@ Error response from registry: pull access denied for docker.io/nginx:latest : re
 Namespace is missing, do you mean `oras pull docker.io/library/nginx:latest`? 
 ```
 
+#### Example 10: Neither registry nor OCI image layout provided when pushing a folder 
+
+Current behavior and output:
+
+```
```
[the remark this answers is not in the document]
FeynmanZhou: Updated all code blocks with adding `console`

docs/proposals/error-handling-guideline.md
FeynmanZhou: Updated. Status code can be can be printed out in the error description if any.

docs/proposals/error-handling-guideline.md
qweeah: Dockerhub is returning 404 for not supporting manifests with subject field, **not** referrers API is not supported. [404 from listing referrers API call](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers) is a known error. There is no doc or post saying Dockerhub will keep returning 404 until manifests with subject is supported, suggest remove this example.

yizha1 (commented):
Thanks @FeynmanZhou 

I made a summary to understand whether examples have a sufficient coverage.  PTAL

- Wrong usage: missing mandatory parameters 
   - Example 1,3,4,6	
- Wrong usage: wrong format
   - Example 2,5,7,8,10
- Client Error (Users inputs are all OK)
   - Example 11
- Server Error
   - Example 9
   
For error due to wrong usage, we not only list the command usage but also the specific reasons and mitigations if any
For client error with correct usage, we do not print out usage, but the detailed reasons
For server error, we print out error message starting with "Error response from registry" and "error code returned by registry" 

I do not see other cases for now.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+Error: /home/user/hello.txt: no such file or directory
+```
+
+## Reference
+
+Parts of the content are borrowed from these guidelines.
+
+- [Azure CLI Error Handling Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md)
+- [Command Line Interface Guidelines](https://clig.dev/#errors)
+- [12 Factor CLI Apps](https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46)
```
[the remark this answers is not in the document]
FeynmanZhou: I chose to list this blog in the reference because some of the guiding principles in this document are inspired by it.

qweeah:
What if an error is a combination of more than one errors? Consider below scenario:
1) oras tried to access a basic-authed registry
2) the registry returns 401 with challenge (error1)
3) oras follows the challenge to found credentials locally and failed, say not found (error2)

In this scenario, should we add both error1 and error2 into the output? If so, how?

Current output is a combination of both errors like `Error: failed to xxx: $HTTP_REQUEST_METHOD $HTTP_REQUEST_URL: credential required for basic auth` below:
```console
$ oras pull localhost:7000/repo:tag --registry-config auth.config -d 
DEBU[0000] Request #0
> Request URL: "http://localhost:7000/v2/repo/manifests/tag"
> Request method: "GET"
> Request headers:
   "Accept": "application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, application/vnd.oci.artifact.manifest.v1+json"
   "User-Agent": "oras/1.1.0" 
DEBU[0000] Response #0
< Response Status: "401 Unauthorized"
< Response headers:
   "Date": "Thu, 28 Dec 2023 09:28:12 GMT"
   "Content-Length": "283"
   "Content-Type": "application/json"
   "Www-Authenticate": "Basic realm=\"Authorization Required\"" 
Error: failed to resolve tag: GET "http://localhost:7000/v2/repo/manifests/tag": credential required for basic auth
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+
+This document aims to provide the guidelines for ORAS contributors to improve existing error messages and error handling method as well as the new error output format. It will also provide recommendations and examples for ORAS CLI contributors for how to write friendly and standard error messages, avoid generating inconsistent and ambiguous error messages.
+
+## General guiding principles
+
+A clear and actionable error message is very important when raising an error, so make sure your error message describes clearly what the error is and tells users what they need to do if possible.
+
+First and foremost, make the error messages descriptive and informative. Error messages are expected to be helpful to troubleshoot where the user has done something wrong and the program is guiding them in the right direction. A great error message is recommended to contain the following elements:
+
+- Status code: optional, when the logs are generated from the server side, it's recommended to print the status code in the error description
```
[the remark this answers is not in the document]
FeynmanZhou: It's HTTP status code from the server side.

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
+Error: missing key in `--annotation` flag: key:value
+```
+
+Suggested error message:
+
+```console
+$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
+Error: annotation value doesn't match the required format.
+Try --annotation "key=value"
```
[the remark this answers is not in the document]
FeynmanZhou: No. ORAS should prompt the correct format only. I updated this prompt to the following:

```console
$ oras push --annotation "key:value" ghcr.io/library/alpine:3.9
Error: annotation value doesn't match the required format.
Please use the correct format in the flag: --annotation "key=value"  
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,258 @@
...
+
+Third, for unhandled errors you didn't expect the user to run into. For that, have a way to view full traceback information as well as full debug or verbose logs output, and instructions on how to submit a bug.
+
+Fourth, signal-to-noise ratio is crucial. The more irrelevant output you produce, the longer it's going to take the user to figure out what they did wrong. If your program produces multiple errors of the same type, consider grouping them under a single explanatory header instead of printing many similar-looking lines.
+
+Fifth, CLI program termination should follow the standard [Exit Status conventions](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) to report execution status information about success or failure. ORAS returns `EXIT_FAILURE` if and only if ORAS reports one or more errors.
+
+Last, error logs can also be useful for post-mortem debugging, truncate them occasionally so they don't eat up space on disk, and make sure they don't contain ansi color codes. Thereby, error logs can be written to a file.
```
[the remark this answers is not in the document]
FeynmanZhou: Removed the description "eat up space on disk"

qweeah:
Also, what if the last API call to registry uses a HEAD method and registry returns no body? E.g. resolve a manifest's digest
```console
$ oras resolve localhost:7000/command/artifacts:foobar -u t -p 2 -d
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
DEBU[0000] Request #0
> Request URL: "http://localhost:7000/v2/command/artifacts/manifests/foobar"
> Request method: "HEAD"
> Request headers:
   "Accept": "application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, application/vnd.oci.artifact.manifest.v1+json"
   "User-Agent": "oras/1.1.0+unreleased"
DEBU[0000] Response #0
< Response Status: "401 Unauthorized"
< Response headers:
   "Content-Type": "application/json"
   "Www-Authenticate": "Basic realm=\"Authorization Required\""
   "Date": "Tue, 02 Jan 2024 08:23:18 GMT"
   "Content-Length": "283"
DEBU[0000] Request #1
> Request URL: "http://localhost:7000/v2/command/artifacts/manifests/foobar"
> Request method: "HEAD"
> Request headers:
   "User-Agent": "oras/1.1.0+unreleased"
   "Accept": "application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, application/vnd.oci.artifact.manifest.v1+json"
   "Authorization": "*****"
DEBU[0000] Response #1
< Response Status: "401 Unauthorized"
< Response headers:
   "Content-Type": "application/json"
   "Www-Authenticate": "Basic realm=\"Authorization Required\""
   "Date": "Tue, 02 Jan 2024 08:23:18 GMT"
   "Content-Length": "283"
Error response from registry: <nil>
```

In this case, what should we output instead of `nil`

FeynmanZhou:
@qweeah Thanks Billy for proposing more cases. I added two examples (12 and 13) and 1 suggestion in the [Error output recommendation](https://github.com/oras-project/oras/pull/1163/files#diff-87978b235bb041f07bf2e71966c261dd6103d718e95ebb9c5c984bebbaf5d357R25) based on these two cases.

FeynmanZhou:
> Thanks @FeynmanZhou
> 
> I made a summary to understand whether examples have a sufficient coverage. PTAL
> 
>     * Wrong usage: missing mandatory parameters
>       
>       * Example 1,3,4,6
> 
>     * Wrong usage: wrong format
>       
>       * Example 2,5,7,8,10
> 
>     * Client Error ( correct usage)
>       
>       * Example 11
> 
>     * Server Error
>       
>       * Example 9
> 
> 
> For error due to wrong usage, we not only list the command usage but also the specific reasons and mitigations if any For client error with correct usage, we do not print out usage, but the detailed reasons For server error, we print out error message starting with "Error response from registry" and "error code returned by registry"
> 
> I do not see other cases for now.

@yizha1 Exactly. There is a new type of error case "error message from both server and client" that has been covered in example 12 and 13. I just added these two based on @qweeah questions above.

docs/proposals/error-handling-guideline.md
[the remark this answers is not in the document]
shizhMSFT: @FeynmanZhou Can you clarify that in the doc?

docs/proposals/error-handling-guideline.md
```diff
@@ -249,6 +250,42 @@ $ oras push localhost:5000/oras:v1 hello.txt
...
+Error response from registry: <nil>
+```
+
+Suggested error message:
+
+```console
+oras resolve localhost:7000/command/artifacts:foobar -u t -p 2
+WARNING! Using --password via the CLI is insecure. Use --password-stdin.
+Error: resolving the digest of artifact localhost:7000/command/artifacts:foobar failed with status: 401 Unauthorized
+Please use correct username or password
```
shizhMSFT, line 286: We might want to rephase this sentence. Examples from Bing Chat:
1. "Invalid login credentials. Please try again."
2. "Authentication failed. Please verify your login details and try again."
3. "Access denied. Please ensure that your login information is correct and try again."

docs/proposals/error-handling-guideline.md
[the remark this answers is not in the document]
FeynmanZhou: @shizhMSFT Updated *status code* to *HTTP status code* in the doc to avoid ambiguous.

docs/proposals/error-handling-guideline.md
FeynmanZhou: Agree. *Login credentials* is much proper than *username/password* in this case. I updated this sentence to:

"Authentication failed. Please verify your login credentials and try again."

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+Usage: oras manifest fetch [flags] <name>{:<tag>|@<digest>}
+You need to specify an artifact reference in the form of "<name>:<tag>" or "<name>@<digest>". Run "oras manifest fetch -h" for more options and examples 
+```
+
+#### Example 4: Push a manifest if no media type flag provided
+
+Current behavior and output:
+
+```console
+$ oras manifest push --oci-layout /sample/images:foobar:mediatype
```
qweeah, line 122:
```suggestion
$ oras manifest push --oci-layout /sample/images:foobar:mediatype manifest.json
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+Error: media type is not recognized. 
+```
+
+Suggested error message:
+
+```console
+$ oras manifest push --oci-layout /sample/images:foobar:mediatype
+Error: media type is not recognized.
```
qweeah, line 130:
```suggestion
$ oras manifest push --oci-layout /sample/images:foobar:mediatype manifest.json
Error: media type is not found in the manifest.json
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+Suggested error message:
+
+```console
+$ oras manifest fetch --oci-layout /tmp/ginkgo1163328512
+Error: "/tmp/ginkgo1163328512": no tag or digest specified
+Usage: oras manifest fetch [flags] <name>{:<tag>|@<digest>}
+You need to specify an artifact reference in the form of "<name>:<tag>" or "<name>@<digest>". Run "oras manifest fetch -h" for more options and examples 
+```
+
+#### Example 4: Push a manifest if no media type flag provided
```
qweeah, line 117:
```suggestion
#### Example 4: Push a manifest if no media type provided
```

docs/proposals/error-handling-guideline.md
qweeah: Or
```suggestion
$ oras manifest push --oci-layout /sample/images:foobar:mediatype
Error: media type is specified via --media-type and not found in manifest.json
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+```console
+$ oras pull localhost:7000/repo:tag --registry-config auth.config
+Error: failed to resolve tag: GET "http://localhost:7000/v2/repo/manifests/tag": credential required for basic auth
+```
+
+Suggested error message:
+
+```console
+$ oras pull localhost:7000/repo:tag --registry-config auth.config
+Error: pulling content from localhost:7000/repo:tag failed with status: 401 Unauthorized
```
qweeah, line 266: Suggest to remove uri and status code since this is not a registry error
```suggestion
Error: failed to auth for oras pull: no valid credential found in auth.config
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+- Use the capital letter ahead of each line of any error message
+- Print human readable error message. If the error message is mainly from the server and varies by different servers, tell users that the error response is from server. This implies that users may need to contact server side for troubleshooting
+- Provide specific and actionable prompt message with argument suggestion or show the example usage for reference. (e.g, Instead of showing flag or argument options is missing, please provide available argument options and guide users to "--help" to view more examples)
+- If the actionable prompt message is too long to show in the CLI output, consider guide users to ORAS user manual or troubleshooting guide with the versioned permanent link
+- If the error message is not enough for troubleshooting, guide users to use "--verbose" to print much more detailed logs
+- If the error message from the server side is too generic or has no value, especially in a few edge cases like example 12 and 13 below, consider providing customized and trimmed error logs to make it clearer. The original server logs can be displayed in debug mode
```
qweeah, line 35:
```suggestion
- If the error message from the server side is too generic or has no value, especially in a few edge cases like example 13 below, consider providing customized and trimmed error logs to make it clearer. The original server logs can be displayed in debug mode
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+
+Suggested error message:
+
+```console
+$ oras pull localhost:7000/repo:tag --registry-config auth.config
+Error: pulling content from localhost:7000/repo:tag failed with status: 401 Unauthorized
+Please check whether the registry credential stored in the authentication file is correct
+```
+
+#### Example 13: failed to resolve the digest with incorrect username or password
```
qweeah, line 270:
```suggestion
#### Example 13: failed to resolve the digest with empty error response from registry
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+Error response from registry: <nil>
+```
+
+Suggested error message:
+
+```console
+oras resolve localhost:7000/command/artifacts:foobar -u t -p 2
+WARNING! Using --password via the CLI is insecure. Use --password-stdin.
+Error: resolving the digest of artifact localhost:7000/command/artifacts:foobar failed with status: 401 Unauthorized
+Authentication failed. Please verify your login credentials and try again.
```
qweeah, line 286:
```suggestion
Error response from registry: empty response body: failed to resolve digest: HEAD "http://localhost:7000/v2/test/manifests/bar": response status code 401: Unauthorized
```

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+```console
+$ oras pull docker.io/nginx:latest
+Error: failed to resolve latest: GET "https://registry-1.docker.io/v2/nginx/manifests/latest": response status code 401: unauthorized: authentication required: [map[Action:pull Class: Name:nginx Type:repository]]
+```
+
+Suggested error message:
+
+```console
+$ oras pull docker.io/nginx:latest
+Error response from registry: pull access denied for docker.io/nginx:latest : response status code 401: unauthorized: requested access to the resource is denied
```
qweeah, line 215: Should not output status code as registry already returns error code, message and detail.
```suggestion
Error response from registry: pull access denied for docker.io/nginx:latest : unauthorized: requested access to the resource is denied
```

shizhMSFT (approved):
LGTM

qweeah (approved):
LGTM

docs/proposals/error-handling-guideline.md
```diff
@@ -0,0 +1,295 @@
...
+WARNING! Using --password via the CLI is insecure. Use --password-stdin.
+Error response from registry: <nil>
+```
+
+Suggested error message:
+
+```console
+oras resolve localhost:7000/command/artifacts:foobar -u t -p 2
+WARNING! Using --password via the CLI is insecure. Use --password-stdin.
+Error response from registry: empty response body: failed to resolve digest: HEAD "http://localhost:7000/v2/test/manifests/bar": response status code 401: Unauthorized
```
qweeah, line 285: Hi @FeynmanZhou Can we change the error message as below:
```suggestion
Error response from registry: recognizable error message not found: failed to resolve digest: HEAD "http://localhost:7000/v2/test/manifests/bar": response status code 401: Unauthorized
```

The proposed error message is more accurate: the response body may be malformed and cannot be recognized by ORAS, but not empty.

TerryHowe (approved):
LGTM

FeynmanZhou:
@oras-project/oras-cli Are we good to merge this PR now?

docs/proposals/error-handling-guideline.md
TerryHowe: This was resolved

PR #1164, opened 2023-11-02
fix: avoid copy suggestion when only config is skipped

sajayantony:
The recommendation to use ORAS copy will be skipped if the only layer not downloaded is the config. 

## When we try to download an image

image

## When there is an artifact with one layer and only empty config is skipped - 

image

### DIFF
--- a/cmd/oras/root/cp.go
+++ b/cmd/oras/root/cp.go
@@ -155,31 +155,31 @@ func doCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.GraphTar
 	}
 
 	const (
-		Exists  = "Exists "
-		Copying = "Copying"
-		Copied  = "Copied "
-		Skipped = "Skipped"
+		promptExists  = "Exists "
+		promptCopying = "Copying"
+		promptCopied  = "Copied "
+		promptSkipped = "Skipped"
 	)
 
 	if opts.TTY == nil {
 		// none TTY output
 		extendedCopyOptions.OnCopySkipped = func(ctx context.Context, desc ocispec.Descriptor) error {
 			committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
-			return display.PrintStatus(desc, Exists, opts.Verbose)
+			return display.PrintStatus(desc, promptExists, opts.Verbose)
 		}
 		extendedCopyOptions.PreCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
-			return display.PrintStatus(desc, Copying, opts.Verbose)
+			return display.PrintStatus(desc, promptCopying, opts.Verbose)
 		}
 		extendedCopyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
 			committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
-			if err := display.PrintSuccessorStatus(ctx, desc, dst, committed, display.StatusPrinter(Skipped, opts.Verbose)); err != nil {
+			if err := display.PrintSuccessorStatus(ctx, desc, dst, committed, display.StatusPrinter(promptSkipped, opts.Verbose)); err != nil {
 				return err
 			}
-			return display.PrintStatus(desc, Copied, opts.Verbose)
+			return display.PrintStatus(desc, promptCopied, opts.Verbose)
 		}
 	} else {
 		// TTY output
-		tracked, err := track.NewTarget(dst, Copying, Copied, opts.TTY)
+		tracked, err := track.NewTarget(dst, promptCopying, promptCopied, opts.TTY)
 		if err != nil {
 			return ocispec.Descriptor{}, err
 		}
@@ -187,12 +187,12 @@ func doCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.GraphTar
 		dst = tracked
 		extendedCopyOptions.OnCopySkipped = func(ctx context.Context, desc ocispec.Descriptor) error {
 			committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
-			return tracked.Prompt(desc, Exists)
+			return tracked.Prompt(desc, promptExists)
 		}
 		extendedCopyOptions.PostCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
 			committed.Store(desc.Digest.String(), desc.Annotations[ocispec.AnnotationTitle])
 			return display.PrintSuccessorStatus(ctx, desc, tracked, committed, func(desc ocispec.Descriptor) error {
-				return tracked.Prompt(desc, Skipped)
+				return tracked.Prompt(desc, promptSkipped)
 			})
 		}
 	}
--- a/cmd/oras/root/cp.go
+++ b/cmd/oras/root/cp.go
@@ -154,25 +154,32 @@ func doCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.GraphTar
 		return graph.Referrers(ctx, src, desc, "")
 	}
 
+	const (
+		Exists  = "Exists "
+		Copying = "Copying"
+		Copied  = "Copied "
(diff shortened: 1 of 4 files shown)


cmd/oras/root/pull.go
```diff
@@ -219,14 +228,16 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
 		var ret []ocispec.Descriptor
 		for _, s := range nodes {
 			if s.Annotations[ocispec.AnnotationTitle] == "" {
-				skippedLayers++
+				if s.Digest != ocispec.DescriptorEmptyJSON.Digest {
+					skippedLayers++
+				}
```
qweeah, line 233: Suggest revert the change since it may also silently skipped an unnamed layer with content `{}`. We can skip config blob in the above `if config != nil` block.
```suggestion
			skippedLayers++
```

cmd/oras/root/pull.go
```diff
@@ -219,14 +228,16 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
 		var ret []ocispec.Descriptor
```
qweeah, line 228: We can change the above `nodes = append(nodes, *config)` to below
```golang
if config.Size != ocispec.DescriptorEmptyJSON.Size && config.Digest != ocispec.DescriptorEmptyJSON.Digest {
    nodes = append(nodes, *config)
}

```

qweeah: Added in https://github.com/oras-project/oras/pull/1164/commits/3ab0152c38be2bafac2cb985d8731a8a00d3ae54

cmd/oras/root/cp.go
```diff
@@ -154,38 +154,45 @@ func doCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.GraphTar
 		return graph.Referrers(ctx, src, desc, "")
 	}
 
+	const (
+		Exists  = "Exists "
+		Copying = "Copying"
+		Copied  = "Copied "
+		Skipped = "Skipped"
```
Wwwsylvia, line 161: Do these variables need to be public? 
nit: Since these variables are passed to `PrintStatus()` as `status`, should we name them`statusExists`, `statusCopying`, etc.  for better readability?

cmd/oras/root/pull.go
```diff
@@ -159,6 +159,15 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
 		}
 	}
 
+	const (
+		Downloading = "Downloading"
+		Pulled      = "Pulled     "
+		Processing  = "Processing "
+		Skipped     = "Skipped    "
+		Restored    = "Restored   "
+		Downloaded  = "Downloaded "
```
Wwwsylvia, line 168: Do we need to export these?

cmd/oras/root/push.go
```diff
@@ -242,33 +242,41 @@ func doPush(dst oras.Target, pack packFunc, copy copyFunc) (ocispec.Descriptor,
 func updateDisplayOption(opts *oras.CopyGraphOptions, fetcher content.Fetcher, verbose bool, tracked track.GraphTarget) {
 	committed := &sync.Map{}
 
+	const (
+		Skipped   = "Skipped  "
+		Uploaded  = "Uploaded "
+		Exists    = "Exists   "
+		Uploading = "Uploading"
+		Uploaded2 = "Uploaded "
```
Wwwsylvia, line 250: What's the difference between `Uploaded` and `Uploaded2`? 🧐

cmd/oras/root/pull.go
sajayantony: Do you want to send the PR over this @qweeah ?

cmd/oras/root/cp.go
```diff
@@ -154,38 +154,45 @@ func doCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.GraphTar
...
+	)
```
shizhMSFT, line 162: nit: it will be more readable if there is a `label` prefix.
```suggestion
	const (
		labelExists  = "Exists "
		labelCopying = "Copying"
		labelCopied  = "Copied "
		labelSkipped = "Skipped"
	)
```

shizhMSFT: Same suggestions for other files

cmd/oras/root/push.go
shizhMSFT: `Uploaded2` is never used and thus can be removed.

shizhMSFT: I'm curious if this `const` block is somehow auto-generated or not?

cmd/oras/root/pull.go
```diff
@@ -213,20 +222,24 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
 					config.Annotations[ocispec.AnnotationTitle] = configPath
 				}
 			})
-			nodes = append(nodes, *config)
+			if config.Size != ocispec.DescriptorEmptyJSON.Size || config.Digest != ocispec.DescriptorEmptyJSON.Digest || config.Annotations[ocispec.AnnotationTitle] != "" {
```
shizhMSFT, line 225: [`content.Equal()`](https://pkg.go.dev/oras.land/oras-go/v2/content#Equal) is a better method to compare two descriptors.

cmd/oras/root/pull.go
```diff
@@ -213,20 +222,24 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
...
+				nodes = append(nodes, *config)
+			}
 		}
 
 		var ret []ocispec.Descriptor
 		for _, s := range nodes {
 			if s.Annotations[ocispec.AnnotationTitle] == "" {
-				skippedLayers++
+				if s.Digest != ocispec.DescriptorEmptyJSON.Digest || s.MediaType != ocispec.DescriptorEmptyJSON.MediaType || s.Size != ocispec.DescriptorEmptyJSON.Size {
```
shizhMSFT, line 233: [`content.Equal()`](https://pkg.go.dev/oras.land/oras-go/v2/content#Equal) is a better method to compare two descriptors.

shizhMSFT, line 233: Do we need to check here as the config has already been filtered out?

cmd/oras/root/pull.go
```diff
@@ -213,20 +222,24 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
...
+					skippedLayers++
```
shizhMSFT, line 234: I just noticed that `skippedLayers` is not go-routine safe.
@qweeah Let's fix it in another PR.

cmd/oras/root/cp.go
qweeah: According to the naming of progress output, those should be `statusXXX`

cmd/oras/root/pull.go
qweeah: No, since different command contains different prompt and all prompts used by one command should be padding to the maximum length. E.g.  `Skipped` for `oras pull` should be `Skipped    `, but for `oras cp` should be `Skipped`

cmd/oras/root/pull.go
qweeah: No, it cannot be used here since the comparison should skip `mediaType`

cmd/oras/root/cp.go
qweeah: How about `promptXXX`? It's used as status prompt in progress output

cmd/oras/root/pull.go
qweeah: We can change it to [atomic.Bool](https://pkg.go.dev/sync/atomic#Bool) since the returned int value is not actually used.

cmd/oras/root/pull.go
qweeah: No we don't, if the config is already an empty descriptor, it's not in the nodes.

shizhMSFT (approved):
LGTM

PR #1169, opened 2023-11-03
fix: update error message for repo ls

sajayantony:
Improves the error message to show the registry and namespace when catalog listing fails. 

```bash
$ oras repo ls docker.io
Error: could not list repositories for "registry-1.docker.io"
GET "https://registry-1.docker.io/v2/_catalog": GET "https://auth.docker.io/token?scope=registry%3Acatalog%3A%2A&service=registry.docker.io": response status code 400: Bad Request
```

```bash 
$ oras repo ls docker.io/python
Error: could not list repositories for "registry-1.docker.io" with prefix "python/"
GET "https://registry-1.docker.io/v2/_catalog": GET "https://auth.docker.io/token?scope=registry%3Acatalog%3A%2A&service=registry.docker.io": response status code 400: Bad Request
```

### DIFF
--- a/cmd/oras/root/repo/ls.go
+++ b/cmd/oras/root/repo/ls.go
@@ -17,6 +17,7 @@ package repo
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -74,7 +75,7 @@ func listRepository(ctx context.Context, opts repositoryOptions) error {
 	if err != nil {
 		return err
 	}
-	return reg.Repositories(ctx, opts.last, func(repos []string) error {
+	err = reg.Repositories(ctx, opts.last, func(repos []string) error {
 		for _, repo := range repos {
 			if subRepo, found := strings.CutPrefix(repo, opts.namespace); found {
 				fmt.Println(subRepo)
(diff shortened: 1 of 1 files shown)


cmd/oras/root/repo/ls.go
```diff
@@ -74,12 +75,24 @@ func listRepository(ctx context.Context, opts repositoryOptions) error {
...
 				fmt.Println(subRepo)
 			}
 		}
 		return nil
 	})
+
+	if err != nil {
+		var repoErr error
+		if opts.namespace != "" {
+			repoErr = fmt.Errorf("could not list repositories for %q with prefix %q", reg.Reference.Host(), opts.namespace)
```
TerryHowe, line 90: You think the user would understand prefix better than namespace?

qweeah: IMHO `prefix` is more accurate in some edge cases, e.g. you can even use `vscode/devcontainers` in MCR. It's not a valid Docker organization namespace
```console
$ oras repo ls mcr.microsoft.com/vscode/devcontainers/
base
javascript-node
typescript-node
universal
python
cpp
php
rust
ruby
go
java
dotnetcore
dotnet
anaconda
miniconda
repos/microsoft/vscode
jekyll
```

TerryHowe (approved):
/lgtm

PR #1174, opened 2023-11-06
test(e2e): add specs to hint for pulling unnamed layers

### LINKED ISSUE #1173: need E2E specs on hinting for pulling unnamed layers
If a layer's descriptor doesn't have `org.opencontainers.image.title` annotation, then it's unnamed. Recently we have added hint on how to pull unnamed layers like:
```
Skipped pulling layers without file name in "org.opencontainers.image.title"
Use 'oras copy xxx --to-oci-layout <layout-dir>' to pull all layers.
```

Need E2E specs to verify:
1. If an artifact contains at least one unnamed non-empty layer, the hint should be shown.
2. If an artifact contains only one empty layer, the hint should not be shown.
3. If all layers in the artifact are named, the hint should not be shown.
4. For 2 and 3, even if the artifact contains unnamed config blob, the hint should not be shown.

qweeah:
This PR adds specs to hint for pulling unnamed layers

Fixes #1173 


shizhMSFT (approved):
LGTM

PR #1176, opened 2023-11-06
chore: make error of non-existed source OCI layout more readable

### LINKED ISSUE #1156: add reference details to output errors

oras 1.1.0

Some oras commands like `oras copy`, `oras discover` and `oras attach` require user to specify a tag or digest. If not provided, oras would output error information like:

> invalid image reference, expecting <name:tag|name@digest>

oras should output more specific error info like:
> no tag or digest when expecting <name:tag|name@digest>

It would make error message more friendly to newbie users who are not familiar with concepts like image reference.

qweeah:
This PR improves error message for OCI layout.

- If the provided path doesn't exist
  - Before:
  ```console
  ➜  oras cp --from-oci-layout foo:bar localhost:5000/test:copied
  Error: stat foo: no such file or directory
  ```

  - After:

  ```console
  ➜  oras cp --from-oci-layout foo:bar localhost:5000/test:copied
  Error: invalid path in "foo:bar": stat foo: no such file or directory
  ```
- If the provided path is a file but not a tarfile
  - Before:
  ```console
  ➜  oras cp --from-oci-layout foo:bar localhost:5000/test:copied
  Error: unexpected EOF
  ```

  - After:

  ```console
  ➜  oras cp --from-oci-layout foo:bar localhost:5000/test:copied
  Error: file "foo" is not a valid tarball: unexpected EOF
  ```

Fixes #1156


cmd/oras/internal/option/target.go
```diff
@@ -154,7 +154,7 @@ func (opts *Target) NewReadonlyTarget(ctx context.Context, common Common, logger
 		}
 		info, err := os.Stat(opts.Path)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("invalid path in %q: %w", opts.RawReference, err)
```
shizhMSFT, line 157: `os.Stat` failure does not imply `invalid path`. Can you try [`fs.ErrNotExist`](https://pkg.go.dev/io/fs#pkg-variables) for a better error message handling?

cmd/oras/internal/option/target.go
```diff
@@ -148,18 +150,27 @@ func (opts *Target) NewReadonlyTarget(ctx context.Context, common Common, logger
 	switch opts.Type {
 	case TargetTypeOCILayout:
 		var err error
+
```
shizhMSFT, line 153: nit: remove extra line

cmd/oras/internal/option/target.go
```diff
@@ -148,18 +150,27 @@ func (opts *Target) NewReadonlyTarget(ctx context.Context, common Common, logger
...
 		opts.Path, opts.Reference, err = parseOCILayoutReference(opts.RawReference)
 		if err != nil {
 			return nil, err
 		}
 		info, err := os.Stat(opts.Path)
 		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				err = fmt.Errorf("invalid path in %q: %w", opts.RawReference, err)
```
shizhMSFT, line 161: How about something like
```
path "foo" in "foo:bar" cannot be found
```

shizhMSFT, line 161: nit: code style
```suggestion
				return nil, fmt.Errorf("invalid path in %q: %w", opts.RawReference, err)
```

cmd/oras/internal/option/target.go
```diff
@@ -148,18 +150,27 @@ func (opts *Target) NewReadonlyTarget(ctx context.Context, common Common, logger
...
 		}
 		if info.IsDir() {
 			return oci.NewFromFS(ctx, os.DirFS(opts.Path))
 		}
-		return oci.NewFromTar(ctx, opts.Path)
+		store, err := oci.NewFromTar(ctx, opts.Path)
+		if err != nil && errors.Is(err, io.ErrUnexpectedEOF) {
+			err = fmt.Errorf("%q is not a valid tarball archive file: %w", opts.Path, err)
+		}
+		return store, err
```
shizhMSFT, line 172: nit: code style
```suggestion
		store, err := oci.NewFromTar(ctx, opts.Path)
		if err != nil {
			if errors.Is(err, io.ErrUnexpectedEOF) {
				return nil, fmt.Errorf("%q is not a valid tarball archive file: %w", opts.Path, err)
			}
			return nil, err
		}
		return store, nil
```

cmd/oras/internal/option/target.go
```diff
@@ -148,18 +150,27 @@ func (opts *Target) NewReadonlyTarget(ctx context.Context, common Common, logger
...
+			}
 			return nil, err
 		}
 		if info.IsDir() {
 			return oci.NewFromFS(ctx, os.DirFS(opts.Path))
 		}
-		return oci.NewFromTar(ctx, opts.Path)
+		store, err := oci.NewFromTar(ctx, opts.Path)
+		if err != nil && errors.Is(err, io.ErrUnexpectedEOF) {
+			err = fmt.Errorf("%q is not a valid tarball archive file: %w", opts.Path, err)
```
shizhMSFT, line 170: We can borrow the output from `tar`.
```suggestion
			err = fmt.Errorf("%q does not look like a tar archive: %w", opts.Path, err)
```

cmd/oras/internal/option/target.go
TerryHowe: I think `cannot find path "foo" in "foo:bar"` reads a bit better

qweeah: Trying to mention `path "foo" in "foo:bar"` here because such statement is ambiguous, user might think ORAS CLI cannot find valid path (foo) in the provided argument (foo:bar). How about 
> Error: invalid arguent "foo:bar": failed to find path "foo": stat foo: no such file or directory

Or swallow the stat error, just
> Error: invalid arguent "foo:bar": failed to find path "foo": no such file or directory

TerryHowe (approved):
/lgtm

shizhMSFT (approved):
LGTM

sajayantony (approved):

PR #1184, opened 2023-11-10
test(e2e): fix a false positive test for `blob get`

qweeah:
This PR corrects a false positive e2e spec.


shizhMSFT (approved):
LGTM

PR #1188, opened 2023-11-16
test(e2e): add test for `resolve` command

qweeah:
This PR adds e2e specs for `oras reslove`.

### DIFF
--- a/test/e2e/suite/command/resolve.go
+++ b/test/e2e/suite/command/resolve.go
@@ -22,12 +22,21 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
+	"oras.land/oras/test/e2e/internal/testdata/feature"
 	"oras.land/oras/test/e2e/internal/testdata/multi_arch"
 	. "oras.land/oras/test/e2e/internal/utils"
 )
 
 var _ = Describe("ORAS beginners:", func() {
 	When("running resolve command", func() {
+		It("should show help description with feature mark", func() {
+			out := ORAS("resolve", "--help").MatchKeyWords(ExampleDesc).Exec().Out
+			gomega.Expect(out.Contents()).Should(gomega.HavePrefix(feature.Experimental.Mark))
+		})
+		It("should show help description via alias", func() {
+			out := ORAS("digest", "--help").MatchKeyWords(ExampleDesc).Exec().Out
+			gomega.Expect(out.Contents()).Should(gomega.HavePrefix(feature.Experimental.Mark))
+		})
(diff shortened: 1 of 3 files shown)


test/e2e/suite/command/resolve.go
```diff
@@ -40,14 +49,20 @@ var _ = Describe("ORAS beginners:", func() {
...
+	When("running resolve command", func() {
 		It("should resolve with just digest", func() {
 			out := ORAS("resolve", RegistryRef(ZOTHost, ImageRepo, multi_arch.Digest)).Exec().Out
 			outString := string(out.Contents())
 			outString = strings.TrimSpace(outString)
 			gomega.Expect(outString).To(gomega.Equal(multi_arch.Digest))
 		})
 		It("should resolve with a fully qualified reference", func() {
-			out := ORAS("digest", "-l", RegistryRef(ZOTHost, ImageRepo, multi_arch.Tag)).Exec().Out
+			out := ORAS("resolve", "-l", RegistryRef(ZOTHost, ImageRepo, multi_arch.Tag)).Exec().Out
```
sajayantony, line 65: I left one command to actually use the digest alias to make sure the command does execute rather than just the help.

test/e2e/suite/command/resolve.go
```diff
@@ -56,3 +71,25 @@ var _ = Describe("ORAS beginners:", func() {
 		})
 	})
 })
+
+var _ = Describe("OCI image layout users", func() {
+	When("running resolve command", func() {
+		It("should resolve with just digest", func() {
+			tmpRoot := GinkgoT().TempDir()
+			out := ORAS("resolve", Flags.Layout, LayoutRef(tmpRoot, multi_arch.Digest)).Exec().Out
```
sajayantony, line 79: Nice. Thanks for adding the layout tests.

test/e2e/suite/command/resolve.go
qweeah: Covered alias case in `oras digest -h`

TerryHowe (approved):
/lgtm

test/e2e/suite/command/resolve.go
sajayantony: If the help validates the output exactly to match the resolve help text then that’s ok. Otherwise it’s better to actually test the alias itself.

shizhMSFT (approved):
LGTM

test/e2e/suite/command/resolve.go
qweeah: Yes you are right, the experimental spec doesn't test anything specific to `oras resolve`. Reverting this change.

shizhMSFT (approved):
LGTM

PR #1190, opened 2023-11-23
feat: scrub Set-Cookie header in debug logs

### LINKED ISSUE #1189: ORAS debug log should scrub Set-Cookie header in the logs

1.1.0

The [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) header might contains sensitive information and need to be scrubbed from the debug logs.

So user won't need to redact the header when pasting the ORAS logs online.

qweeah:
Fixes #1189.


internal/trace/transport.go
```diff
@@ -68,7 +68,7 @@ func logHeader(header http.Header) string {
 	if len(header) > 0 {
 		headers := []string{}
 		for k, v := range header {
-			if strings.EqualFold(k, "Authorization") {
+			if strings.EqualFold(k, "Authorization") || strings.EqualFold(k, "Set-Cookie") {
```
shizhMSFT, line 71: Can we define an array for this?

qweeah: Refactored and defined a map of it.

TerryHowe (approved):
/lgtm

shizhMSFT (approved):
LGTM

PR #1191, opened 2023-11-23
build(ci): add workflow for releasing main image

### LINKED ISSUE #1155: Adding a new workflow to build a container image tagged main on PR merge
It would be useful to have a Github workflow that automatically builds a container image whenever a PR is merged into the main branch. In this way, users can play with features from the latest version of the code easily via running a GHCR image without having to build it themselves.

kranurag7:
GitHub workflow that will build and push a new image whenever a new commit is merged on main branch.

I added a commit on my fork's main for [testing](https://github.com/kranurag7/oras/pkgs/container/oras). 

Fixes #1155


.github/workflows/release-ghcr-main.yml
```diff
@@ -0,0 +1,47 @@
...
+
+name: release-ghcr-main
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
TerryHowe, line 22: Is this pretty standard as a build image?

kranurag7: This is the environment of the runner in which we execute our job. In this case, it's an `ubuntu-latest` machine. 
If you want the builder instance to be specific, then we can use `runs-on: ubuntu-22.04`. 
I'm in favor of keeping it specific, but in other workflow we use the latest instance, so I preferred the same. 

https://github.com/oras-project/oras/blob/9ecf303eeb2620dfe3ddf15721999d1f7e8fd4bf/.github/workflows/release-ghcr.yml#L23

qweeah (approved):
LGTM. 

I think what's proposed in this PR can be merged into the https://github.com/oras-project/oras/blob/main/.github/workflows/release-ghcr.yml by changing the prepare job to
```
TAG=${GITHUB_REF#refs/*/}
echo "ref=ghcr.io/${{ github.repository }}:${TAG}" >> $GITHUB_OUTPUT
```
This can be done in another PR.

shizhMSFT (approved):
LGTM

PR #1197, opened 2023-11-30
feat: support manifest and blob deletion for OCI layout

### LINKED ISSUE #761: support OCI image layout target in `oras blob delete`
User should be able to use `oras blob delete` to remove a blob in their OCI layout folder with properly garbage collection.

Depending on sdk support: https://github.com/oras-project/oras-go/issues/454

### LINKED ISSUE #762: support OCI image layout target in `oras manifest delete`
User should be able to use `oras manifest delete` to remove a manifest in their OCI layout folder with properly garbage collection.

qweeah:
Fixes #761, fixes #762 


cmd/oras/internal/option/target.go
```diff
@@ -114,26 +116,69 @@ func parseOCILayoutReference(raw string) (path string, ref string, err error) {
...
-		var err error
-		opts.Path, opts.Reference, err = parseOCILayoutReference(opts.RawReference)
+		return opts.newOCIStore()
+	case TargetTypeRemote:
+		return opts.newRepository(common, logger)
+	}
+	return nil, fmt.Errorf("unknown target type: %q", opts.Type)
+}
+
+type resolvableDeleter interface {
```
shizhMSFT, line 151: Should we export this?

shizhMSFT (approved):
LGTM with suggestions

cmd/oras/internal/option/target.go
qweeah: I think exporting `NewBlobDeleter` and `NewManifestDeleter` are already enough.

qweeah: We can add it to oras-go BTW, it's common to resolve before deletion and should be useful to other SDK callers.

TerryHowe (approved):
/lgtm

PR #1201, opened 2023-12-12
feat: improve returned error for invalid arguments

### LINKED ISSUE #458: Improve output if invalid number of command line arguments are provided
For all ORAS commands (login, logout, pull, push, attach, discover, copy), if the number of input argument doesn't match expected, user will get errors like: 
> Error: accepts X arg(s), received Y

It could confuse users who are not familiar with cobra concepts. E.g. they might think number of flags doesn't meet what is expected.

We need to provide more clear details for the error, including:
1) what is expected
2) what is got from input

qweeah:
This PR improves the error output when the number of input argument doesn't match expected. In the new output, user will get
1) What kind of input is expecting, as well as the usage hint
2) What's the wrong input
3) Which command to run to get more detailed examples and help

E.g. Taking `oras discover` as an example
```console
# Before
$ oras discover localhost: 5000/test:demo
Error: accepts 1 arg(s), received 2
# After
$ oras discover localhost: 5000/test:demo
Error: "oras discover" requires exactly 1 argument but got 2
Usage: oras discover [flags] <name>{:<tag>|@<digest>}
Please specify exactly 1 argument as the target artifact to discover referrers from. Run "oras discover -h" for more options and examples
```

Fixes #458

### DIFF
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -17,6 +17,7 @@ package errors
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2/registry"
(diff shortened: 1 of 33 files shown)


cmd/oras/internal/errors/errors.go
```diff
@@ -18,9 +18,43 @@ package errors
 import (
 	"fmt"
 
+	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2/registry"
 )
 
+// Error is the error type for CLI error messaging.
+type Error struct {
```
shizhMSFT, line 26: nit: Should we implement `Unwrap() error`?

shizhMSFT (approved):
LGTM

cmd/oras/internal/errors/errors.go
qweeah: Currently there is no use case but worth doing since an error is already wrapped.

shizhMSFT (approved):
LGTM

Issue #1208, opened 2023-12-22
push with multiple tags failed with TTY


qweeah:
### What happened in your environment?

Got below error when pushing and tagging a manifest with `oras push` from a terminal
```console
$ oras push localhost:7000/test:v1,v2
Uploading empty artifact
✓ Exists    application/vnd.oci.empty.v1+json                                                                         2/2  B 100.00%     0s 
  └─ sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
✓ Uploaded  application/vnd.oci.image.manifest.v1+json                                                            535/535  B 100.00%     0s
  └─ sha256:ccf0a406902c9820f6db75612fd2a9ed5ec928f971256d2e8091c1e7bb35ba0f
Pushed [registry] localhost:7000/test:v1
Error: failed to push content: progress output manager has already been stopped

### What did you expect to happen?

Tagging should be successful.

### How can we reproduce it?

Use `oras push` to push with multiple tags

### What is the version of your ORAS CLI?

https://github.com/oras-project/oras/commit/a377128efdcbd58d832c4164e448f92bc63cf310

### What is your OS environment?

Ubuntu 20.04

### Are you willing to submit PRs to fix it?

- [X] Yes, I am willing to fix it.

FeynmanZhou:
I will verify whether this issue is resolved or not.

PR #1209, opened 2023-12-22
fix: bug in pushing with multiple tags

qweeah:
This PR fixed the bug mentioned in #1208 with minimal refactoring on `getTrackedTarget`

### DIFF
--- a/cmd/oras/internal/display/track/target.go
+++ b/cmd/oras/internal/display/track/target.go
@@ -118,8 +118,3 @@ func (t *graphTarget) Prompt(desc ocispec.Descriptor, prompt string) error {
 func (t *graphTarget) Inner() oras.GraphTarget {
 	return t.GraphTarget
 }
-
-// Inner returns the inner oras.GraphTarget.
-func (t *referenceGraphTarget) Inner() oras.GraphTarget {
-	return t.GraphTarget
-}
--- a/cmd/oras/internal/display/track/target.go
+++ b/cmd/oras/internal/display/track/target.go
@@ -31,6 +31,7 @@ type GraphTarget interface {
 	oras.GraphTarget
 	io.Closer
 	Prompt(desc ocispec.Descriptor, prompt string) error
+	Inner() oras.GraphTarget
 }
 
 type graphTarget struct {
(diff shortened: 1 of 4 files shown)


cmd/oras/internal/display/track/target.go
```diff
@@ -112,3 +113,13 @@ func (t *graphTarget) Prompt(desc ocispec.Descriptor, prompt string) error {
...
+
+// Inner returns the inner oras.GraphTarget.
+func (t *graphTarget) Inner() oras.GraphTarget {
+	return t.GraphTarget
+}
+
+// Inner returns the inner oras.GraphTarget.
+func (t *referenceGraphTarget) Inner() oras.GraphTarget {
+	return t.GraphTarget
+}
```
shizhMSFT, line 125: This looks like redundant as `*referenceGraphTarget` already has `Inner()` from `*graphTarget`.

cmd/oras/root/pull.go
```diff
@@ -303,11 +302,12 @@ func generateContentKey(desc ocispec.Descriptor) string {
...
 }
 
-func printOnce(printed *sync.Map, s ocispec.Descriptor, msg string, verbose bool, tracked track.GraphTarget) error {
+func printOnce(printed *sync.Map, s ocispec.Descriptor, msg string, verbose bool, dst any) error {
 	if _, loaded := printed.LoadOrStore(generateContentKey(s), true); loaded {
 		return nil
 	}
-	if tracked == nil {
+	tracked, ok := dst.(track.GraphTarget)
+	if !ok {
```
shizhMSFT, line 312: It is better to have pattern like
```go
if blablabla, ok := doTypeAssert(); ok {
    doSomething()
    return something
}

doSomething()
return something
```

cmd/oras/root/push.go
```diff
@@ -248,8 +251,8 @@ func updateDisplayOption(opts *oras.CopyGraphOptions, fetcher content.Fetcher, v
 		promptExists    = "Exists   "
 		promptUploading = "Uploading"
 	)
-
-	if tracked == nil {
+	tracked, ok := dst.(track.GraphTarget)
+	if !ok {
```
shizhMSFT, line 257: It is better to have pattern like
```go
if blablabla, ok := doTypeAssert(); ok {
    doSomething()
    return something
}

doSomething()
return something
```

shizhMSFT (approved):
LGTM

PR #1220, opened 2023-12-26
build(ci): enforce coverage check

qweeah:
This enforces the test coverage check by 75% lines per project


.github/codecov.yml
```diff
@@ -15,4 +15,5 @@ coverage:
   status:
     project:
       default:
-        target: 60%
\ No newline at end of file
+        target: 75%
+        if_ci_failed: error
```
TerryHowe, line 19: Maybe not such a big jump in coverage right away?

qweeah: Seems a big change but actually we already achieved ~80% test coverage. Considering we already have 2900+(3600 line * 80%) lines covered, to make coverage lower than 75% means 250 lines of new code were added without testing.

TerryHowe (approved):
/lgtm

shizhMSFT (approved):
LGTM

PR #1222, opened 2023-12-28
chore(ux): show registry error and hint for dockehub

### LINKED ISSUE #1216: improve error message when namespace is not specified in the reference of Dockerhub artifacts
Users might forget to add the namespace (e.g. `library` #1151) when pulling or copying Dockerhub images and get unexpected 401. Need to improve the error message and suggest user to add namespace.

Current behavior
```console
$ oras pull docker.io/nginx:latest
Error: failed to resolve latest: GET "https://registry-1.docker.io/v2/nginx/manifests/latest": response status code 401: unauthorized: authentication required: [map[Action:pull Class: Name:nginx Type:repository]]
```

Suggested error message:

```console
$ oras pull docker.io/nginx:latest
Error response from registry: pull access denied for docker.io/nginx:latest : response status code 401: unauthorized: requested access to the resource is denied
Namespace is missing, do you mean `oras pull docker.io/library/nginx:latest`? 
```

qweeah:
This PR 
1) adds prefix to errors returned by registry. E.g. invalid creds for fetching a manifest
```console
$ oras manifest fetch-config localhost:7000/repo:tag -u hello -p 123 >>
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Error response from registry: unauthorized: authentication required
```

2) gives recommendation if namespace is missing in a Dockerhub reference.
```console
$ oras pull docker.io/apline:latest
Error response from registry: failed to resolve latest: unauthorized: authentication required: [map[Action:pull Class: Name:apline Type:repository]]
Namespace is missing, do you mean `oras pull docker.io/library/apline:latest`?
```

Fixes #1216


cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,51 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+// Command returns an error-handled cobra command.
+func Command(cmd *cobra.Command, handler Handler) *cobra.Command {
+	runE := cmd.RunE
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		err := runE(cmd, args)
+		if err == nil {
+			return nil
+		}
+		processor, err := handler.Handle(err, cmd)
+		if processor == nil {
```
shizhMSFT, line 89:
```suggestion
		if err != nil {
```

cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,51 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+			return err
+		}
+		return processor.Process(err, cmd.CommandPath())
```
shizhMSFT, line 92: The design of `Handler` and `Processor` seems can be simplified.

Here's the analysis:
1. The `err` is returned by `handler.Handle()`. That implies `Handle` has the information of `err`.
2. Similarly, `Handle` has the information of `processor`.

Combine 1 and 2, we can simplify the above code as
```go
processor := handler.Handle(err, cmd)
processor.Process(cmd.CommandPath())
```
or simply
```go
return handler.Handle(err, cmd).Process(cmd.CommandPath())
```
which can be further simplified as
```go
type ErrorHandler interface {
	HandleError(cmd *cobra.Command, err error) error
}
```
and
```go
return handler.HandleError(cmd, err)
```

cmd/oras/internal/option/remote.go
```diff
@@ -322,3 +326,32 @@ func (opts *Remote) isPlainHttp(registry string) bool {
...
+}
+
+// Process processes error into oerrors.Error.
+func (opts *Remote) Process(err error, _ string) *oerrors.Error {
+	ret := oerrors.Error{
+		Err: err,
+	}
+	var errResp *errcode.ErrorResponse
+	if errors.As(err, &errResp) {
+		ret.Err = oerrors.GetInner(err, errResp)
```
shizhMSFT, line 354: This pattern seems can be further refactored.

cmd/oras/internal/errors/errors.go
qweeah: refactored

cmd/oras/internal/errors/errors.go
qweeah: Tried to use `Hande` and `Process` to decouple the logic of setting output prefix, scrubbing error info and give recommendation for know errors. I have merged the two interfaces.

cmd/oras/internal/option/remote.go
qweeah: Could you explain more detail on how to do that?

cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,45 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+	runE := cmd.RunE
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		err := runE(cmd, args)
+		if err == nil {
+			return nil
+		}
+		if handled := handler.Handle(err, cmd); handled != nil {
+			return handled
+		}
+		return err
```
shizhMSFT, line 86:
```suggestion
		err := runE(cmd, args)
		if err != nil {
			return handler.Handle(err, cmd)
		}
		return nil
```

cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,45 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
 	}
 }
 
+// Handler handles error during cmd execution.
+type Handler interface {
+	// Handle handles error during cmd execution.
+	// if nil is returned, then err cannot be handled.
+	Handle(err error, cmd *cobra.Command) error
```
shizhMSFT, line 72: nit: naming convention
```suggestion
	Handle(cmd *cobra.Command, err error) error
```

shizhMSFT: Example: https://pkg.go.dev/io/fs#WalkDirFunc
```go
type WalkDirFunc func(path string, d DirEntry, err error) error
```

cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,45 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+		return err
+	}
+	return cmd
+}
+
+// GetInner gets the inner error from the error response.
+func GetInner(err error, errResp *errcode.ErrorResponse) error {
+	inner := errResp.Errors
+	if len(inner) == 0 {
+		return fmt.Errorf("empty response body: %w", errResp)
```
shizhMSFT, line 95: `len(inner) == 0` does not imply empty response body but unrecognised response body.

See code: https://github.com/mostoja/oras-go/blob/9cac41930202c5f5fd8f5a2041862d593571c564/registry/remote/internal/errutil/errutil.go#L33-L48

cmd/oras/internal/errors/errors.go
shizhMSFT: In fact, this handler is not a handler but a modifier.

Therefore, consider
```go
type Modifier interface {
    Modify(cmd *cobra.Command, err error) error
}

File: cmd/oras/internal/errors/errors.go [hunk 7]
```diff
@@ -60,6 +65,45 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+	} else {
+		errContent := err.Error()
+		errRespContent := errResp.Error()
+		if idx := strings.Index(errContent, errRespContent); idx > 0 {
+			// remove HTTP related info
+			return fmt.Errorf("%s%w", errContent[:idx], inner)
+		}
```
shizhMSFT, line 102: Will it be safer to use `strings.Replace`?

cmd/oras/internal/option/remote.go
```diff
@@ -322,3 +326,18 @@ func (opts *Remote) isPlainHttp(registry string) bool {
...
+		cmd.SetErrPrefix(oerrors.RegistryErrorPrefix)
+		return err
+	} else if errors.As(err, &errResp) {
+		cmd.SetErrPrefix(oerrors.RegistryErrorPrefix)
+		return &oerrors.Error{
+			Err: oerrors.GetInner(err, errResp),
+		}
+	}
+	return nil
+}
```
shizhMSFT, line 343: Followed by above comments, consider
```suggestion
// Modify modifies error during cmd execution.
func (opts *Remote) Modify(cmd *cobra.Command, err error) error {
	if errors.Is(err, errdef.ErrNotFound) {
		cmd.SetErrPrefix(oerrors.RegistryErrorPrefix)
		return err
	}

	var errResp *errcode.ErrorResponse
	if errors.As(err, &errResp) {
		cmd.SetErrPrefix(oerrors.RegistryErrorPrefix)
		return &oerrors.Error{
			Err: oerrors.GetInner(err, errResp),
		}
	}

	return err
}
```

cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,45 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+		if handled := handler.Handle(err, cmd); handled != nil {
+			return handled
+		}
+		return err
+	}
+	return cmd
+}
+
+// GetInner gets the inner error from the error response.
+func GetInner(err error, errResp *errcode.ErrorResponse) error {
```
shizhMSFT, line 92: Based on the code, this method is not to get the inner error from the error response but trim the error message. A proper name should be given to this method.

cmd/oras/internal/option/target.go
```diff
@@ -269,3 +311,11 @@ func (opts *BinaryTarget) Parse() error {
 	opts.To.resolveFlag = append(opts.resolveFlag, opts.To.resolveFlag...)
 	return Parse(opts)
 }
+
+// Handle handles error during cmd execution.
+func (opts *BinaryTarget) Handle(err error, cmd *cobra.Command) error {
+	if err := opts.From.Handle(err, cmd); err != nil {
+		return err
+	}
+	return opts.To.Handle(err, cmd)
```
shizhMSFT, line 320: Can we get line 320 covered?

cmd/oras/internal/errors/errors.go
qweeah: > return handler.Handle(err, cmd)

This is wrong, if the function returns nil, then the original `err` must be returned.

cmd/oras/internal/errors/errors.go
qweeah: What do you suggest as a proper output? 
How about `no error code found: `? 
I suggested to change the prefix to `Empty error response from registry` but rejected by @FeynmanZhou. He thinks it's too complicated for user to know three different kind of prefixes.

cmd/oras/internal/errors/errors.go
qweeah: `strings.Replace` cannot exclude the case when `idx == 0`. Also it requires extra code to do error wrapping.

cmd/oras/internal/option/target.go
qweeah: Marked. Will add test to make it covered once the refactoring is done.

cmd/oras/internal/errors/errors.go
```diff
@@ -60,6 +65,60 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
+	return inner
+}
+
+// reWrap re-wraps errA to errC and trims out errB, returns nil if scrub fails.
+// +---------- errA ----------+
+// |         +---- errB ----+ |      +---- errA ----+
+// |         |    errC      | |  =>  |     errC     |
+// |         +--------------+ |      +--------------+
+// +--------------------------+
+func reWrap(errA error, errB error, errC error) error {
```
shizhMSFT, line 111: nit:
```suggestion
func reWrap(errA, errB, errC error) error {
```

cmd/oras/internal/option/target.go
```diff
@@ -239,8 +244,54 @@ func (opts *Target) EnsureReferenceNotEmpty() error {
...
+		cmd.SetErrPrefix(oerrors.RegistryErrorPrefix)
+		ret := &oerrors.Error{
+			Err: oerrors.Trim(err, errResp),
+		}
+
+		if ref.Registry == "docker.io" && errResp.StatusCode == http.StatusUnauthorized {
+			if ref.Repository != "" && !strings.Contains(ref.Repository, "/") {
+				// docker.io/xxx -> docker.io/library/xxx
+				ref.Repository = "library/" + ref.Repository
+				ret.Recommendation = fmt.Sprintf("Namespace is missing, do you mean `%s %s`?", cmd.CommandPath(), ref)
```
shizhMSFT, line 284: nit:
```suggestion
				ret.Recommendation = fmt.Sprintf("Namespace seems missing. Do you mean `%s %s`?", cmd.CommandPath(), ref)
```

shizhMSFT (approved):
LGTM with suggestions

PR #1227, opened 2024-01-03
WIP: fix: add image spec option back to attach for expanded compatibility …

### LINKED ISSUE #1224: oras attach failing for registries that don't support 1.1-RC image media type

Opening a new issue to track `oras attach` failing to work for 1.0 registries (historically, issue #690 : oras attach on AWS ECR: Invalid JSON syntax). This was at one point working, so new issue to discuss more recent changes.

The changes to address issue #1043 in PR #1054 is where the client dropped the `--image-spec` option. There have been subsequent changes and some refactoring here and in oras-go since then.

With the `--image-spec` option removed, I would have expected the media type to be `image.manifest.v1`. However, without the option, attach uses `image.manifest.v1.1` without any way to select 1.0. 

### How can we reproduce it?

With ECR:
`oras attach --artifact-type some/example <registry-id>.dkr.ecr.<region>.amazonaws.com/<repo>:<tag> example.txt`

Presumably, this is reproducible with any 1.0 registry that does not accept a 1.1 RC media type. 

1.1.0

n/a

jlbutler:
Fixes #1224 

This is a WIP, would like to get opinion from @shizhMSFT @qweeah and others on requiring changes in oras-go or not.

The proposed fix for ECR (and hopefully Gitlab, probably others) is to reinstate the oras client 1.0 behavior which allows the user to specify `--image-spec` w/ `oras attach`, in order to allow users to specify using `v1.1.0-rc2`. This version image manifest was broadly supported by 1.0 registries like ECR. With the default in oras client 1.1 now using `v1.1.0-rc4`, attach is not longer working for ECR (and likely other registries).

I don't think we can cleanly do this with zero changes to oras-go, because the string vals for the oras client `--image-spec` option come from oras-go (`oras.PackManifestVersion`). 

I thought about using `oras.Pack` (as suggested by @shizhMSFT), but I at least need to add `v1.1.0-rc2` as a pack manifest version.

Since the RC version support is unstable and will be removed in the future, I'm leaning toward leaning into an oras-go PR (and version bump for oras client), and putting most of the changes to support deprecated image spec release candidates into oras-go, where it can be limited and eventually removed more easily. 

See accompanying [WIP PR in oras-go](https://github.com/oras-project/oras-go/pull/667) for conversation.

TODO

### DIFF
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -37,6 +37,7 @@ import (
 type attachOptions struct {
 	option.Common
 	option.Packer
+	option.ImageSpec
 	option.Target
 
 	artifactType string
@@ -57,6 +58,9 @@ Example - Attach file 'hi.txt' with aritifact type 'doc/example' to manifest 'he
 
 Example - Push file "hi.txt" with the custom layer media type 'application/vnd.me.hi':
   oras attach --artifact-type doc/example localhost:5000/hello:v1 hi.txt:application/vnd.me.hi
+
+Example - Attach file "hi.txt" with specific image spec release candidate for testing during spec development (default v1.1):
+  oras attach --artifact-type doc/example --image-spec v1.1.0-rc2 localhost:5000/hello:v1 hi.txt 
 
 Example - Attach file "hi.txt" using a specific method for the Referrers API:
   oras attach --artifact-type doc/example --distribution-spec v1.1-referrers-api localhost:5000/hello:v1 hi.txt # via API
@@ -149,8 +153,21 @@ func runAttach(ctx context.Context, opts attachOptions) error {
 		ManifestAnnotations: annotations[option.AnnotationManifest],
 		Layers:              descs,
 	}
+
 	pack := func() (ocispec.Descriptor, error) {
-		return oras.PackManifest(ctx, store, oras.PackManifestVersion1_1_RC4, opts.artifactType, packOpts)
+		switch opts.PackVersion {
+		case oras.PackManifestVersion1_0:
+			return ocispec.Descriptor{}, errors.New("image-spec v1.0 is not supported")
+		case oras.PackManifestVersion1_1_RC2:
+			return oras.Pack(ctx, store, opts.artifactType, packOpts.Layers, oras.PackOptions{
+				Subject:             packOpts.Subject,
+				ManifestAnnotations: packOpts.ManifestAnnotations,
+				PackImageManifest:   true,
+			})
+		default:
+			// default oras.PackManifestVersion1_1_RC4
+			return oras.PackManifest(ctx, store, opts.PackVersion, opts.artifactType, packOpts)
+		}
 	}
 
 	copy := func(root ocispec.Descriptor) error {
--- a/docs/proposals/compatibility-mode.md
+++ b/docs/proposals/compatibility-mode.md
@@ -16,12 +16,17 @@ This document elaborates on the major changes of ORAS CLI v1.1.0 proposed in [is
 
 Using the following flags in `oras push` and `oras attach` respectively with different variables to configure the manifest build and distribution behaviors. 
 
-- Using a flag `--image-spec` with `oras push`
+- Using a flag `--image-spec` with `oras push` and `oras attach` to configure image specification compatibility.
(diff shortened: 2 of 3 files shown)


cmd/oras/root/attach.go
```diff
@@ -149,8 +154,9 @@ func runAttach(ctx context.Context, opts attachOptions) error {
 		ManifestAnnotations: annotations[option.AnnotationManifest],
 		Layers:              descs,
 	}
+
 	pack := func() (ocispec.Descriptor, error) {
-		return oras.PackManifest(ctx, store, oras.PackManifestVersion1_1_RC4, opts.artifactType, packOpts)
+		return oras.PackManifest(ctx, store, opts.PackVersion, opts.artifactType, packOpts)
```
shizhMSFT, line 159: We can do something like
```go
switch opts.PackVersion {
    case ImageSpecV1_0:
        return ocispec.Descriptor{}, errors.New("image-spec v1.0 is not supported")
    case PackManifestVersion1_1_RC2:
        return oras.Pack(ctx, store, opts.artifactType, packOpts.Layers, oras.PackOptions{
            Subject:             packOpts.Subject,
            ManifestAnnotations: packOpts.ManifestAnnotations,
            PackImageManifest:   true,
        })
    default:
        return oras.PackManifest(ctx, store, opts.PackVersion, opts.artifactType, packOpts)
}
```

docs/proposals/compatibility-mode.md
```diff
@@ -16,12 +16,14 @@ This document elaborates on the major changes of ORAS CLI v1.1.0 proposed in [is
...
+- Using a flag `--image-spec` with `oras push` and `oras attach` to configure image specification compatibility.
 - Using a flag `--distribution-spec` with `oras attach`, `oras cp`, and `oras manifest push` to configure compatibility with registry when pushing or copying an OCI image manifest. This flag is also applicable to `oras discover` for viewing and filtering the referrers.
 
 ### Build and push OCI image manifest using a flag `--image-spec`
 
-Use the flag `--image-spec <spec version>` in `oras push` to specify which version of the OCI Image-spec when building and pushing an OCI image manifest. It supports specifying the option `v1.0` or `v1.1` as the spec version. The option `v1.1` is the default behavior in `oras push` since ORAS CLI v1.1.0 so users don't need to manually specify this option.
+Use the flag `--image-spec <spec version>` in `oras push` and `oras attach` to specify which version of the OCI Image-spec when building and pushing an OCI image manifest. It supports specifying the option `v1.0` or `v1.1` as the spec version. The option `v1.1` is the default behavior in `oras push` since ORAS CLI v1.1.0 so users don't need to manually specify this option.
```
shizhMSFT, line 24: `oras attach` does not support `--image-spec v1.0` as it does not specify the `subject` field.

shizhMSFT (changes requested):

docs/proposals/compatibility-mode.md
FeynmanZhou, line 24: Can we list all available options (values) of the flag `--image-spec`?

FeynmanZhou: Also, this flag should be experimental

cmd/oras/root/attach.go
jlbutler: We certainly can, but that still requires a change in oras-go and a dep bump in oras. I figured we could keep the details out of the client with the initial proposal, but if we'd rather have it obvious and in the client, I can make this change.

docs/proposals/compatibility-mode.md
jlbutler: Might users want to use `oras push` for v1.0 manifests into the future?

jlbutler:
I've modified this PR to align with feedback, but given the Gitlab registry issue seems unrelated, this looks to potentially be an ECR-only issue related to supporting requirements for client-managed referrers as specified by Image spec 1.1.0-rc2 and not as specified by rc4.  Specifically, ECR allows `subject` field, but not `artifactType`.

Given fallback support as describe by 1.1. release candidates should be OCI 1.0 compliant, I am going to pause this PR (keep it WIP for now) and investigate extending support in ECR.

TerryHowe:
Still a WIP?

jlbutler:
I did some testing and after some internal discussions, we are going to allow `artifactType` on manifests in ECR. So I'll withdraw this PR.

Thanks so much for the support and discussion along the way.

PR #1242, opened 2024-01-16
feat: support mounting blobs during copy

### LINKED ISSUE #1238: Add mount support for `oras cp`

main

When the source and destination registries are the same, the code needs to simply needs to tell oras-go via the oras.CopyOptions that we want to mount from the source repository if available.

An example of how to do this is [here](https://github.com/oras-project/oras-go/blob/main/example_copy_test.go#L218).  You need to implement the `MountFrom` function and have it return the source repository (if the registry is the same as the destination).  You probably also want to implement `OnMount`.

There is a large performance improvement possible if mounting is supported.  `crane cp` has always supported mounting and now that oras-go supports mounting (via [PR](https://github.com/oras-project/oras-go/pull/665)), we should add it to `oras cp`.

qweeah:
This PR add mount support for `oras cp`. 

Fixes #1238 


cmd/oras/internal/display/track/target.go
```diff
@@ -66,6 +67,14 @@ func NewTarget(t oras.GraphTarget, actionPrompt, donePrompt string, tty *os.File
 	return gt, nil
 }
 
+// Mount makes graphTarget implement oras.Mounter.
+func (t *graphTarget) Mount(ctx context.Context, desc ocispec.Descriptor, fromRepo string, getContent func() (io.ReadCloser, error)) error {
+	if mounter, ok := t.GraphTarget.(registry.Mounter); ok {
+		return mounter.Mount(ctx, desc, fromRepo, getContent)
+	}
+	return errors.New("not implemented")
+}
```
shizhMSFT, line 76: Do we need a new structure to make OCI layout scenarios work?

qweeah: OCI layout will be blocked by `opts.MountFrom`

qweeah: Can you elaborate more about `work`? If there is an OCI layout involved in copying, we only need to block this method from being called. Currently it's via `opts.MountFrom`.

qweeah: Discussed offline with @shizhMSFT, will update to type assertion.

qweeah:
@ktarplee Can you help review if this implementation met the expected design of https://github.com/oras-project/oras-go/pull/665

cmd/oras/internal/display/track/target.go
```diff
@@ -66,6 +66,13 @@ func NewTarget(t oras.GraphTarget, actionPrompt, donePrompt string, tty *os.File
 	return gt, nil
 }
 
+// Mount makes graphTarget implement oras.Mounter.
```
shizhMSFT, line 69:
```suggestion
// Mount mounts a blob from a specified repository. This method is invoked only by the `*remote.Repository` target.
```

ktarplee:
> @ktarplee Can you help review if this implementation met the expected design of [oras-project/oras-go#665](https://github.com/oras-project/oras-go/pull/665)

This looks great!  I did not see any issues.

shizhMSFT (approved):
LGTM

PR #1265, opened 2024-02-01
fix: remove unwanted newline in attach error

qweeah:
This PR fixes unexpected output error in the recommendation.


shizhMSFT (approved):
LGTM

Issue #638, opened 2022-10-10
Oras commands like attach, discover, and pull support  an option to format output as JSON


plooploops:
### What is the version of your ORAS CLI

Version:        0.15.0 Go version:     go1.19 Git commit:     94abaecb072d7c44bf484cb7070db2c94be2a028 Git tree state: clean

### What would you like to be added?

An option to format output for commands like `attach`, `discover`, and `pull` as JSON.

**Updated version** per comments:

#### For oras discover

For example, suppose you make the call:

`oras discover <myacr>.azurecr.io/my/repo:tag -o json`

And then you would get a JSON output like the following:

```json
{
  "referrers": [
    {
      "digest": "sha256:someHash",
      "mediaType": "someType",
      "artifactType": "someArtifactType",
      "reference" : "<myacr>.azurecr.io/my/repo@sha256:someHash",
      "size": sizeInBytes
    },
  ]
}
```

This would allow you to pull in the artifact (e.g. `oras pull <reference value>`.

#### For oras pull

For `oras pull`, it would be helpful to get the downloaded `artifactType`, `reference`, and `files` along with the other fields described in the below sample output.

Suppose you make the following call:

`oras pull <reference value> -o json`

And then you would get a JSON output like the following:

```json
{
  "digest": "sha256:someHash",
  "mediaType": "someType",
  "artifactType": "someArtifactType",
  "reference" : "<myacr>.azurecr.io/my/repo@sha256:someHash",
  "files" : [
    "path/to/my/downloaded/artifact.json",
    "path/to/another/downloaded/artifact.json"
  ],
  "size": sizeInBytes
}
```

#### For oras attach

For `oras attach`, it would also be helpful to get the uploaded `artifactDigest`, `reference`, and `files` along with the other fields described in the below sample output.

Suppose you make the following call:

`oras attach "<myacr>.azurecr.io/my/repo:tag" --artifact-type "someArtifactType" <path/to/some/artifact>:<type> -o json`

And then you would get a JSON output like the following:

```json
{
  "digest": "sha256:someHash",
  "mediaType": "someType",
  "artifactType": "someArtifactType",
  "reference" : "<myacr>.azurecr.io/my/repo@sha256:someHash",
  "files" :  [
    "path/to/my/uploaded/artifact.json",
    "path/to/my/other/uploaded/artifact.json"
  ],
  "size": sizeInBytes
}
```

### Why is this needed for ORAS?

[...]

PR #1288, opened 2024-03-13
feat: Use cobra.Command.OutOrStdout method for output

TerryHowe:

This contains the command refactor I did in another PR.

With this change, the output of a command can be set to something other than stdout. For example capture the output of a command:
```
    var stringWriter strings.Builder
    cmd := root.New()
    cmd.SetArgs([]string{"repo", "ls", "mcr.microsoft.com"})
    cmd.SetOutput(&stringWriter)
    _ := cmd.Execute()
    fmt.Println(stringWriter.String())
```

### DIFF
--- a/cmd/oras/root/blob/delete.go
+++ b/cmd/oras/root/blob/delete.go
@@ -82,13 +82,14 @@ func deleteBlob(cmd *cobra.Command, opts *deleteBlobOptions) (err error) {
 	}
 
 	// add both pull and delete scope hints for dst repository to save potential delete-scope token requests during deleting
+	outWriter := cmd.OutOrStdout()
 	ctx = registryutil.WithScopeHint(ctx, blobs, auth.ActionPull, auth.ActionDelete)
 	desc, err := blobs.Resolve(ctx, opts.Reference)
 	if err != nil {
 		if errors.Is(err, errdef.ErrNotFound) {
 			if opts.Force && !opts.OutputDescriptor {
 				// ignore nonexistent
-				fmt.Println("Missing", opts.RawReference)
+				fmt.Fprintln(outWriter, "Missing", opts.RawReference)
 				return nil
 			}
 			return fmt.Errorf("%s: the specified blob does not exist", opts.RawReference)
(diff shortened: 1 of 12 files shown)


qweeah:
The proposed change in this PR doesn't cover all output of ORAS. E.g. the output of `oras push/attach --format xxx` which is designed to be used for automation. (There are also undergoing efforts for other commands, see #638) We have a draft doc https://hackmd.io/@shizh/HyFZkEcca to categorize all types of output in ORAS.

@TerryHowe Looks like you are trying to do automation and the target output to be changed is metadata output?

FeynmanZhou:
Thanks @TerryHowe . This is a big PR. From users' perspective, I am curious on what's the impact to ORAS CLI output. There is not an issue associated with this PR to describe the problem or enhancement.

TerryHowe:
This change has no effect on the format of the output and from my understanding no effect on CLI output. The change is really just for go scripting.

* This change uses `Fprintln` instead of `Println` so we can pass a writer.
* For the CLI `cmd.OutOrStadout()` will always return `os.Stdout` so no change in behavior

TerryHowe:
The goal is to allow oras go scripting to be able to capture output from commands without messing with stdout

qweeah:
> The goal is to allow oras go scripting to be able to capture output from commands without messing with stdout

The output of `--format` won't go into STDOUT, e.g. `localhost:5000/test@sha256:16ce5d4fb98496ec805e6b2401213598e710dbd936fb58cb7a325d2582924694` cannot captured by the go script.
```console
> oras version
Version:        1.2.0-beta.1
Go version:     go1.21.6
Git commit:     9ffdb3eec60b969d842af1a9e699202e0827fa01
Git tree state: clean
> oras push localhost:5000/test:format --format {{.Ref}}
✓ Uploaded  application/vnd.oci.empty.v1+json                                                      2/2  B 100.00%   37ms
  └─ sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
✓ Uploaded  application/vnd.oci.image.manifest.v1+json                                         535/535  B 100.00%     0s
  └─ sha256:16ce5d4fb98496ec805e6b2401213598e710dbd936fb58cb7a325d2582924694
localhost:5000/test@sha256:16ce5d4fb98496ec805e6b2401213598e710dbd936fb58cb7a325d2582924694
```

TerryHowe:
> The output of `--format` won't go into STDOUT, e.g. `localhost:5000/test@sha256:16ce5d4fb98496ec805e6b2401213598e710dbd936fb58cb7a325d2582924694` cannot captured by the go script.

I'd sooner handle format as a follow up.

qweeah (approved):
LGTM

PR #1290, opened 2024-03-15
fix: Revert github actions coverage until fixed

TerryHowe:

It looks like there are problems with v4 coverage I was not aware off. Reverting for now.

https://github.com/codecov/codecov-action/issues/1292

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,6 @@ jobs:
       - name: Check Version
         run: bin/linux/amd64/oras version
       - name: Upload coverage to codecov.io
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v3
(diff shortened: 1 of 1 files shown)


sajayantony (approved):

qweeah:
Actually, it's not caused by codecov GHA version update, it's a known transient failure on code coverage side. The original PR do passes the build, see [logs](https://github.com/oras-project/oras/actions/runs/8291914448).

qweeah:
Interesting, looks like merging the PR will sign the squashed commit with an unkown GPG key, which makes the codecov actions fail. So even if the PR build passes, the main branch build may still fail.

TerryHowe:
Interesting. I did create an issue https://github.com/oras-project/oras/issues/1291 to update when there is a fix

PR #1292, opened 2024-03-17
Printer refactor

TerryHowe:

Create a printer object to manage output


cmd/oras/root/manifest/push.go
```diff
@@ -184,7 +185,7 @@ func pushManifest(cmd *cobra.Command, opts pushOptions) error {
 		}
 		return opts.Output(os.Stdout, descJSON)
 	}
-	status.Print("Pushed", opts.AnnotatedReference())
+	printer.Print("Pushed", opts.AnnotatedReference())
```
qweeah, line 188: @TerryHowe Is this blocking your implementation of ORAS Golang scripting? I think this output should be classified to metadata output and it's better to 
```golang
_, displayMetadata := display.NewPushHandler("", nil, opts.Verbose) // avoiding formatting and TTY output
displayMetadata.OnCopied(&opts.Target)
```

TerryHowe: What I'm trying to develop here is a layer similar to logrus that doesn't know what is is printing, it just knows how to print. To me at least, ideally the format code would be a layer on top of that.

qweeah: Gotcha. Any good idea to inject the writer in? The writer comes from stdout or cobra command out.

cmd/oras/internal/display/status/print.go
```diff
@@ -27,17 +27,33 @@ import (
 	"oras.land/oras-go/v2/registry"
 )
 
-var printLock sync.Mutex
+type Printer struct {
```
qweeah, line 30: IMHO this object should hold a writer, which accepts the output writer(out or stdout) of command. Printer should always prints to the writer.

qweeah (changes requested):

qweeah:
@TerryHowe I added a new `Printer` interface to display package to unify the output into STDOUT.

https://github.com/oras-project/oras/blob/61e0241fe11342ca27065b8e3cb4bf66ba142d15/cmd/oras/internal/display/metadata/view/printer.go#L33-L40

It provides `PrintXXX` to wrap `fmt.FprintXXX` utilities, also provides `PrintJSON` and `ParseAndWrite` for formatted output

To reflect your changes in #1288, cobra.Command.OutOrStdout can be injected via below interface
https://github.com/oras-project/oras/blob/61e0241fe11342ca27065b8e3cb4bf66ba142d15/cmd/oras/internal/display/metadata/view/printer.go#L27-L31

Please help you help review related changes in #1293 and let me know your comment on this design, thanks.

qweeah:
@TerryHowe I have added another solution in https://github.com/oras-project/oras/pull/1314, please let me know if it meets your expectation.

PR #1295, opened 2024-03-18
feat: support `--format` in `manifest fetch` command

qweeah:
This PR add `--format` flag to `manifest fetch` command to support field formatting for OCI manifests.

Related to #638


cmd/oras/root/manifest/fetch.go
```diff
@@ -72,9 +74,22 @@ Example - Fetch raw manifest from an OCI layout archive file 'layout.tar':
 `,
 		Args: oerrors.CheckArgs(argument.Exactly(1), "the manifest to fetch"),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			if opts.outputPath == "-" && opts.OutputDescriptor {
-				return errors.New("`--output -` cannot be used with `--descriptor` at the same time")
```
TerryHowe, line 76: Any table drive code like table tests I find a lot harder to debug because if there is a problem, you aren't sure which item you were processing when the problem occurred. It also tends to be more lines of code then just performing each action.

cmd/oras/root/manifest/fetch.go
```diff
@@ -130,14 +149,37 @@ func fetchManifest(cmd *cobra.Command, opts *fetchOptions) (fetchErr error) {
...
+					}
+				default:
+					return fmt.Errorf("cannot apply template to %q: unsupported media type %s", opts.RawReference, desc.MediaType)
+				}
+			}
 		}
-
-		// save manifest content into the local file if the output path is provided
-		if err = os.WriteFile(opts.outputPath, content, 0666); err != nil {
-			return err
```
TerryHowe, line 140: This could be generalized into some output writer maybe

cmd/oras/root/manifest/fetch.go
qweeah: How about 
```golang
switch {
case opts.outputPath == "-" && opts.Template != "":
  return fmt.Errorf("`--output -` cannot be used with `--format` at the same time")
case opts.OutputDescriptor && opts.Template != "" :
  return fmt.Errorf("`--descriptor` cannot be used with `--format` at the same time")
case opts.OutputDescriptor && opts.outputPath != "":
  return fmt.Errorf("`--descriptor` cannot be used with `--output` at the same time")
}
```

BTW I will need change to tabulated checking when the number of exclusive flags increase to 4 anyway

cmd/oras/internal/display/metadata/template/manifest_fetch.go
```diff
@@ -0,0 +1,41 @@
...
+// NewManifestFetchHandler creates a new handler for manifest fetch events.
+func NewManifestFetchHandler(template string) metadata.ManifestFetchHandler {
+	return &ManifestFetchHandler{
+		template: template,
+	}
+}
+
+// OnFetched is called after the manifest fetch is completed.
+func (ph *ManifestFetchHandler) OnFetched(manifest ocispec.Manifest) error {
+	return parseAndWrite(model.ToMappable(reflect.ValueOf(manifest)), ph.template)
```
shizhMSFT, line 40: Can we use `json.Marshal` and the n`json.Unmarshal` instead of `ToMappable`? It is safer although slower.

cmd/oras/root/manifest/fetch.go
```diff
@@ -130,14 +149,37 @@ func fetchManifest(cmd *cobra.Command, opts *fetchOptions) (fetchErr error) {
...
+						return err
+					}
+					if err = handler.OnFetched(manifest); err != nil {
+						return err
+					}
+				default:
+					return fmt.Errorf("cannot apply template to %q: unsupported media type %s", opts.RawReference, desc.MediaType)
+				}
+			}
 		}
```
shizhMSFT, line 174: This part does not seem correct. They should be refactored to "metadata" output for different types.

cmd/oras/root/manifest/fetch.go
shizhMSFT: `--descriptor` can be used with `--output` but cannot be used with `--output -`.

shizhMSFT: Is there any reason that `--descriptor` cannot be used with `--format`?

shizhMSFT (changes requested):

cmd/oras/root/manifest/fetch.go
qweeah: > Is there any reason that `--descriptor` cannot be used with `--format`?

Both takes stdout and will corrupt each other.

cmd/oras/internal/display/metadata/template/manifest_fetch.go
qweeah: Maybe I am missing sth here, but how does the marshall+unmarshall operation convert all fields into a map?

cmd/oras/internal/option/pretty.go
```diff
@@ -26,18 +26,18 @@ import (
...
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Pretty) ApplyFlags(fs *pflag.FlagSet) {
-	fs.BoolVarP(&opts.pretty, "pretty", "", false, "prettify JSON objects printed to stdout")
+	fs.BoolVarP(&opts.Pretty, "pretty", "", false, "prettify JSON objects printed to stdout")
 }
 
 // Output outputs the prettified content if `--pretty` flag is used. Otherwise
 // outputs the original content.
 func (opts *Pretty) Output(w io.Writer, content []byte) error {
```
shizhMSFT, line 39: Should it call `utils.PrintJSON`?

cmd/oras/internal/display/handler.go
```diff
@@ -94,3 +96,32 @@ func NewPullHandler(format string, path string, tty *os.File, out io.Writer, ver
 	}
 	return statusHandler, metadataHandler
 }
+
+// NewManifestFetchHandler returns a manifest fetch handler.
+func NewManifestFetchHandler(out io.Writer, format string, outputDescriptor bool, pretty bool, outputPath string) (metadata.ManifestFetchHandler, content.ManifestFetchHandler) {
```
shizhMSFT, line 101: nit:
```suggestion
func NewManifestFetchHandler(out io.Writer, format string, outputDescriptor, pretty bool, outputPath string) (metadata.ManifestFetchHandler, content.ManifestFetchHandler) {
```

cmd/oras/internal/display/handler.go
```diff
@@ -94,3 +96,32 @@ func NewPullHandler(format string, path string, tty *os.File, out io.Writer, ver
...
+	var metadataHandler metadata.ManifestFetchHandler
+	var contentHandler content.ManifestFetchHandler = content.NewManifestFetchHandler(out, pretty, outputPath)
```
shizhMSFT, line 103:
```suggestion
	var contentHandler content.ManifestFetchHandler
	
	// switch ...
	
	if contentHandler == nil {
	 	 contentHandler = content.NewManifestFetchHandler(out, pretty, outputPath)
	}
```

cmd/oras/internal/display/metadata/discard.go
```diff
@@ -0,0 +1,28 @@
...
+
+type discard struct{}
+
+// NewDiscardHandler creates a new handler that discards output for all events.
+func NewDiscardHandler() ManifestFetchHandler {
+	return discard{}
+}
+
+// OnFetched implements ManifestFetchHandler.
+func (discard) OnFetched(ocispec.Descriptor, []byte) error { return nil }
```
shizhMSFT, line 28:
```suggestion
func (discard) OnFetched(ocispec.Descriptor, []byte) error {
    return nil
}
```

cmd/oras/root/manifest/fetch.go
```diff
@@ -86,6 +90,9 @@ Example - Fetch raw manifest from an OCI layout archive file 'layout.tar':
 
 	cmd.Flags().StringSliceVarP(&opts.mediaTypes, "media-type", "", nil, "accepted media types")
 	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "file `path` to write the fetched manifest to, use - for stdout")
+	cmd.Flags().StringVar(&opts.Template, "format", "", `Format output using a custom template:
```
shizhMSFT, line 93: Should it be experimental?

cmd/oras/root/manifest/fetch.go
```diff
@@ -90,7 +90,7 @@ Example - Fetch raw manifest from an OCI layout archive file 'layout.tar':
 
 	cmd.Flags().StringSliceVarP(&opts.mediaTypes, "media-type", "", nil, "accepted media types")
 	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "file `path` to write the fetched manifest to, use - for stdout")
-	cmd.Flags().StringVar(&opts.Template, "format", "", `Format output using a custom template:
+	cmd.Flags().StringVar(&opts.Template, "format", "", `[Experimental] Format fetched manifest using a custom template:
```
shizhMSFT, line 93: It is not "fetched manifest" but the operation metadata.

cmd/oras/internal/display/metadata/template/manifest_fetch.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+// OnFetched is called after the manifest fetch is completed.
+func (h *manifestFetchHandler) OnFetched(path string, desc ocispec.Descriptor, content []byte) error {
+	var manifest map[string]any
+	if err := json.Unmarshal(content, &manifest); err != nil {
+		return err
+	}
+	// return parseAndWrite(h.out, model.ToMappable(reflect.ValueOf(manifest)), h.template)
```
shizhMSFT, line 47: This line should be removed.

cmd/oras/internal/display/metadata/json/manifest_fetch.go
```diff
@@ -0,0 +1,46 @@
...
+	return &manifestFetchHandler{
+		out: out,
+	}
+}
+
+// OnFetched is called after the manifest fetch is completed.
+func (h *manifestFetchHandler) OnFetched(path string, desc ocispec.Descriptor, content []byte) error {
+	var manifest map[string]any
+	if err := json.Unmarshal(content, &manifest); err != nil {
+		return err
```
shizhMSFT, line 43: If the manifest cannot be unmarshaled, it should not block the manifest fetch operation.

cmd/oras/internal/display/metadata/template/manifest_fetch.go
```diff
@@ -0,0 +1,49 @@
...
+		template: template,
+		out:      out,
+	}
+}
+
+// OnFetched is called after the manifest fetch is completed.
+func (h *manifestFetchHandler) OnFetched(path string, desc ocispec.Descriptor, content []byte) error {
+	var manifest map[string]any
+	if err := json.Unmarshal(content, &manifest); err != nil {
+		return err
```
shizhMSFT, line 45: If the manifest cannot be unmarshaled, it should not block the manifest fetch operation.

cmd/oras/root/manifest/fetch.go
```diff
@@ -90,9 +90,9 @@ Example - Fetch raw manifest from an OCI layout archive file 'layout.tar':
 
 	cmd.Flags().StringSliceVarP(&opts.mediaTypes, "media-type", "", nil, "accepted media types")
 	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "file `path` to write the fetched manifest to, use - for stdout")
-	cmd.Flags().StringVar(&opts.Template, "format", "", `[Experimental] Format fetched manifest using a custom template:
-'json':       Print manifest in prettified JSON format
-'$TEMPLATE':  Print output using the given Go template.`)
+	cmd.Flags().StringVar(&opts.Template, "format", "", `[Experimental] Format operation metadata using a custom template:
```
shizhMSFT, line 93: Is there any other term other than `operation metadata`?

cmd/oras/root/manifest/fetch.go
```diff
@@ -86,6 +90,9 @@ Example - Fetch raw manifest from an OCI layout archive file 'layout.tar':
 
 	cmd.Flags().StringSliceVarP(&opts.mediaTypes, "media-type", "", nil, "accepted media types")
 	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "file `path` to write the fetched manifest to, use - for stdout")
+	cmd.Flags().StringVar(&opts.Template, "format", "", `[Experimental] Format metadata using a custom template:
```
qweeah, line 93: @FeynmanZhou Could you help confirm if the description make sense?

shizhMSFT (approved):
LGTM

cmd/oras/root/manifest/fetch.go
qweeah: Feel free to let me know and we can resolve that in another PR

PR #1199, opened 2023-12-04
doc(spec): formatted ORAS CLI output

FeynmanZhou:
**What this PR does / why we need it**:

add spec: Formatted ORAS CLI output

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #638

PR #1302, opened 2024-03-21
fix: discard tagging output when pushing with formatted output

### LINKED ISSUE #1300: pushing with multiple tags will corrupt `--format xxx` output

I want to 
1) push an artifact to localhost:5000/test with tag `v1.1`
2) tag the pushed artifact with `latest` and `v1`

```console
$ oras push localhost:5000/test:v1.1,v1,latest --format {{.Ref}} --no-tty 
Tagged latest
Tagged v1
localhost:5000/test@sha256:0e0e404fe40f2409b1964272434fbe3b9293f262fd62f90ab9df49d54579818e
```

The output contains both tagging log and digest reference of pushed artifact.

The output should only contain digest reference of pushed artifact.
```console
$ oras push localhost:5000/test:v1.1,v1,latest --format {{.Ref}} --no-tty 
localhost:5000/test@sha256:0e0e404fe40f2409b1964272434fbe3b9293f262fd62f90ab9df49d54579818e
```
### How can we reproduce it?

Run below command with oras v1.2.0-beta.1 registry running at port 5000:

```bash
oras push localhost:5000/test:v1.1,v1,latest --format {{.Ref}} --no-tty 
```

v1.2.0-beta.1

ubuntu 20.04

qweeah:
This PR discards the tagging output of `oras push` to avoid corrupting formatted output

Fixes #1300 


cmd/oras/internal/display/status/interface.go
```diff
@@ -30,3 +30,10 @@ type PushHandler interface {
 
 // AttachHandler handles text status output for attach command.
 type AttachHandler PushHandler
+
+type TagHandler interface {
+	// OnTagged is called when each tagging operation is done.
+	OnTagged(tag string) error
+	// PreTagging is called before tagging.
+	PreTagging(reference string)
+}
```
TerryHowe, line 39: General comment about these display handlers, they are very specific to the command and IMO should be part of the command rather than the status/printing. They are like tasks for the command or steps.

cmd/oras/internal/display/status/discard.go
```diff
@@ -45,3 +45,11 @@ func (DiscardHandler) TrackTarget(gt oras.GraphTarget) (oras.GraphTarget, error)
...
 // UpdateCopyOptions updates the copy options for the artifact push.
 func (DiscardHandler) UpdateCopyOptions(opts *oras.CopyGraphOptions, fetcher content.Fetcher) {}
+
+// OnTagged implements TagHandler interface.
+func (DiscardHandler) OnTagged(tag string) error {
+	return nil
+}
+
+// PreTagging implements TagHandler interface.
+func (DiscardHandler) PreTagging(reference string) {}
```
TerryHowe, line 55: I know these is all implemented this way already, but green field, I would think of text as a format and we wouldn't be discarding anything, it would just be another format.

qweeah: First of all, it's necessary to do so since user explicitly wants to make the output structured via `--format <json | go-template>`, that's why we need to add such implementation to discard text output.

Secondly, you are right about making the tagging info into another format here. We should add the tag metadata into the structured output of `oras push`. The tag metadata currently not included in [spec PR](https://github.com/oras-project/oras/pull/1199) because it's user input and there is no clear use case to use it to automate stuffs.

cmd/oras/internal/display/status/interface.go
qweeah: Any good idea on the refactoring? This is only interface and will be used by handler factory (see handlers in `display` package) and command codes.

shizhMSFT (commented):
`Tagged xxx` in `oras push` should be metadata output instead of status output.

cmd/oras/internal/display/status/interface.go
TerryHowe: I don't think they need to be refactored, it just seesm like they shouldn't be under `display`. I'll look at it more eventually, but I like to at least throw out these ideas for consideration.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #1275, opened 2024-02-22
bump: update golang to `1.22`


shizhMSFT:
[Go 1.22](https://go.dev/blog/go1.22) is released with [release notes](https://go.dev/doc/go1.22)!

It is time to bump up golang version to `1.22` by updating

- `go.mod` for `oras` and its E2E tests.
- GitHub Actions
- Dockerfile
- `snapcraft.yaml`

Reference:
- #1074

shizhMSFT:
This issue depends on https://github.com/oras-project/oras-go/issues/713

PR #1058, opened 2023-08-09
chore: make error returned by `oras tag` more readable

qweeah:
This PR improves the readability of error info when an invalid reference is provided:
```console
$ oras tag list test
Error: unable to add tag for 'list': invalid reference: missing repository
```

PR #1303, opened 2024-03-22
bump: bump `oras-go` to 2.5.0

### LINKED ISSUE #1304: `oras tag` cannot retag an artifact in OCI image layout

```console
$ oras tag root:test --oci-layout test1
Error: unable to add tag for 'root@test': invalid reference: missing registry or repository
```

Should tag `test` in folder `root` to `test1`

### How can we reproduce it?

```console
$ oras tag root@test --oci-layout test1
```

oras v1.2.0-beta.1

Ubuntu 20.04

wangxiaoxuan273:

Part of #1275 
Fixes #1304 

### DIFF
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -120,7 +120,6 @@ func (opts *Target) parseOCILayoutReference() error {
 	raw := opts.RawReference
 	var path string
 	var ref string
-
 	if idx := strings.LastIndex(raw, "@"); idx != -1 {
 		// `digest` found
 		path = raw[:idx]
@@ -130,8 +129,7 @@ func (opts *Target) parseOCILayoutReference() error {
 		var err error
 		path, ref, err = fileref.Parse(raw, "")
 		if err != nil {
-			err = errors.Join(err, errdef.ErrInvalidReference)
-			return err
+			return errors.Join(err, errdef.ErrInvalidReference)
 		}
 	}
 	opts.Path = path
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -98,13 +98,11 @@ func (opts *Target) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description 
 func (opts *Target) Parse() error {
 	switch {
 	case opts.IsOCILayout:
-		var err error
 		opts.Type = TargetTypeOCILayout
 		if len(opts.headerFlags) != 0 {
 			return errors.New("custom header flags cannot be used on an OCI image layout target")
 		}
-		opts.Path, opts.Reference, err = opts.parseOCILayoutReference()
-		return err
+		return opts.parseOCILayoutReference()
 	default:
 		opts.Type = TargetTypeRemote
 		if _, err := registry.ParseReference(opts.RawReference); err != nil {
@@ -118,8 +116,12 @@ func (opts *Target) Parse() error {
 }
 
 // parseOCILayoutReference parses the raw in format of <path>[:<tag>|@<digest>]
-func (opts *Target) parseOCILayoutReference() (path string, ref string, err error) {
+func (opts *Target) parseOCILayoutReference() error {
 	raw := opts.RawReference
+	var path string
+	var ref string
+	var err error
+
 	if idx := strings.LastIndex(raw, "@"); idx != -1 {
 		// `digest` found
 		path = raw[:idx]
@@ -129,9 +131,12 @@ func (opts *Target) parseOCILayoutReference() (path string, ref string, err erro
 		path, ref, err = fileref.Parse(raw, "")
 		if err != nil {
 			err = errors.Join(err, errdef.ErrInvalidReference)
+			return err
 		}
 	}
-	return
+	opts.Path = path
+	opts.Reference = ref
+	return nil
 }
 
 func (opts *Target) newOCIStore() (*oci.Store, error) {
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -125,7 +125,6 @@ func (opts *Target) parseOCILayoutReference() (path string, ref string, err erro
 		ref = raw[idx+1:]
 	} else {
 		// find `tag`
-		// // use Errors.Join with invalidReference, only when error != nil
 		path, ref, err = fileref.Parse(raw, "")
 		if err != nil {
 			err = errors.Join(err, errdef.ErrInvalidReference)
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -107,7 +107,7 @@ func (opts *Target) Parse() error {
 		opts.Type = TargetTypeRemote
 		if _, err := registry.ParseReference(opts.RawReference); err != nil {
 			return &oerrors.Error{
-				Err:            fmt.Errorf("%w: %q", err, opts.RawReference),
+				Err:            fmt.Errorf("%q: %w", opts.RawReference, err),
 				Recommendation: "Please make sure the provided reference is in the form of <registry>/<repo>[:tag|@digest]",
 			}
 		}
@@ -120,7 +120,6 @@ func (opts *Target) parseOCILayoutReference() error {
 	raw := opts.RawReference
 	var path string
 	var ref string
-	var err error
 
 	if idx := strings.LastIndex(raw, "@"); idx != -1 {
 		// `digest` found
@@ -128,6 +127,7 @@ func (opts *Target) parseOCILayoutReference() error {
 		ref = raw[idx+1:]
 	} else {
 		// find `tag`
+		var err error
 		path, ref, err = fileref.Parse(raw, "")
 		if err != nil {
 			err = errors.Join(err, errdef.ErrInvalidReference)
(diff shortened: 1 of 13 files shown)


shizhMSFT (changes requested):
The build and the lint failed.

The build failure is due to the E2E test failure caused by a bug in checking the reference for oci-layout (check line 77).

https://github.com/oras-project/oras/blob/e8da6e53ec3a679c966c948e1428ab6a72eee04f/cmd/oras/root/tag.go#L75-L82

qweeah:
@shizhMSFT is right,  you need to 1) handle all invalid reference error in `tag list` reflecting https://github.com/oras-project/oras/pull/1058; 2) improve target.Parse to accommodate changes in oras-go.

qweeah:
Also need to refactor OCI reference parsing and do it in target.Parse, so `unable to add tag for '%s'` can be printed for OCI layout

shizhMSFT (changes requested):
This PR covers #1307 but does not cover #1306

cmd/oras/internal/option/target.go
```diff
@@ -102,14 +102,13 @@ func (opts *Target) Parse() error {
 		if len(opts.headerFlags) != 0 {
 			return errors.New("custom header flags cannot be used on an OCI image layout target")
 		}
-		return nil
+		_, _, err := opts.parseOCILayoutReference()
```
qweeah, line 105: Is it possible to save the path and reference here and remove the `parseOCILayoutReference` call in `newXXX` here?

```suggestion
		opts.Path, opts.Reference, err := opts.parseOCILayoutReference()
```

cmd/oras/internal/option/target.go
```diff
@@ -102,14 +102,13 @@ func (opts *Target) Parse() error {
...
+		return err
 	default:
 		opts.Type = TargetTypeRemote
-		if ref, err := registry.ParseReference(opts.RawReference); err != nil {
-			return err
-		} else if ref.Registry == "" || ref.Repository == "" {
+		if _, err := registry.ParseReference(opts.RawReference); err != nil {
 			return &oerrors.Error{
-				Err:            fmt.Errorf("%q is an invalid reference", opts.RawReference),
+				Err:            fmt.Errorf("%w: %q is an invalid reference", err, opts.RawReference),
```
qweeah, line 111: Since `ErrInvalidReference` is already `errors.New("invalid reference")`, how about 
```suggestion
				Err:            fmt.Errorf("%w: %q", err, opts.RawReference),
```

qweeah (changes requested):
Since this PR also fixes #1304, we also need to enhance E2E tests.

cmd/oras/internal/option/target.go
wangxiaoxuan273: saved the values here and removed the references of `parseOCILayoutReference`

wangxiaoxuan273:
> Since this PR also fixes #1304, we also need to enhance E2E tests.

Added a test case for #1304

cmd/oras/internal/option/target.go
```diff
@@ -118,24 +118,23 @@ func (opts *Target) Parse() error {
 }
 
 // parseOCILayoutReference parses the raw in format of <path>[:<tag>|@<digest>]
-func parseOCILayoutReference(raw string) (path string, ref string, err error) {
+func (opts *Target) parseOCILayoutReference() (path string, ref string, err error) {
```
qweeah, line 121:
```suggestion
func (opts *Target) parseOCILayoutReference() error {
```

cmd/oras/internal/option/target.go
```diff
@@ -118,24 +118,23 @@ func (opts *Target) Parse() error {
...
 	if idx := strings.LastIndex(raw, "@"); idx != -1 {
 		// `digest` found
 		path = raw[:idx]
 		ref = raw[idx+1:]
 	} else {
 		// find `tag`
 		path, ref, err = fileref.Parse(raw, "")
+		if err != nil {
+			err = errors.Join(err, errdef.ErrInvalidReference)
+		}
```
qweeah, line 132: Also setup target option for digest code path
```suggestion
		if err != nil {
			err = errors.Join(err, errdef.ErrInvalidReference)
		} else {
		        opts.Path = path
		        opts.Reference = ref
		}
```

cmd/oras/internal/option/target.go
```diff
@@ -98,18 +98,18 @@ func (opts *Target) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
 func (opts *Target) Parse() error {
 	switch {
 	case opts.IsOCILayout:
+		var err error
```
qweeah, line 101: NIT: Better move this line closer to its first usage.

cmd/oras/internal/option/target.go
```diff
@@ -98,18 +98,18 @@ func (opts *Target) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
...
 		opts.Type = TargetTypeOCILayout
 		if len(opts.headerFlags) != 0 {
 			return errors.New("custom header flags cannot be used on an OCI image layout target")
 		}
-		return nil
+		opts.Path, opts.Reference, err = opts.parseOCILayoutReference()
```
qweeah, line 106: Can we move the path and reference setup into the parse function?
```suggestion
		return opts.parseOCILayoutReference()
```

qweeah (changes requested):

qweeah (approved):
LGTM

cmd/oras/internal/option/target.go
```diff
@@ -102,14 +102,12 @@ func (opts *Target) Parse() error {
...
+		return opts.parseOCILayoutReference()
 	default:
 		opts.Type = TargetTypeRemote
-		if ref, err := registry.ParseReference(opts.RawReference); err != nil {
-			return err
-		} else if ref.Registry == "" || ref.Repository == "" {
+		if _, err := registry.ParseReference(opts.RawReference); err != nil {
 			return &oerrors.Error{
-				Err:            fmt.Errorf("%q is an invalid reference", opts.RawReference),
+				Err:            fmt.Errorf("%w: %q", err, opts.RawReference),
```
shizhMSFT, line 110: By convention, the `err` is always at the end.
```suggestion
				Err:            fmt.Errorf("%q: %w", opts.RawReference, err),
```

cmd/oras/internal/option/target.go
```diff
@@ -118,24 +116,30 @@ func (opts *Target) Parse() error {
 }
 
 // parseOCILayoutReference parses the raw in format of <path>[:<tag>|@<digest>]
-func parseOCILayoutReference(raw string) (path string, ref string, err error) {
+func (opts *Target) parseOCILayoutReference() error {
+	raw := opts.RawReference
+	var path string
+	var ref string
+	var err error
```
shizhMSFT, line 123: This line should be in between of line 130 and line 131 since it is only used in the `else` block.

cmd/oras/root/tag.go
```diff
@@ -74,11 +74,16 @@ Example - Tag the manifest 'v1.0.1' to 'v1.0.2' in an OCI image layout folder 'l
 		},
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			opts.RawReference = args[0]
-			if _, err := registry.ParseReference(opts.RawReference); err != nil {
-				return fmt.Errorf("unable to add tag for '%s': %w", opts.RawReference, err)
-			}
 			opts.targetRefs = args[1:]
-			return option.Parse(&opts)
+			if err := option.Parse(&opts); err != nil {
+				if inner, ok := err.(*oerrors.Error); ok {
```
shizhMSFT, line 79: I'm curious about the reason why this type assertion is required?

shizhMSFT: `errors.Is()` does `Unwarp` the inner errors.

See https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/errors/wrap.go;l=44-78

shizhMSFT (changes requested):

cmd/oras/internal/option/target.go
wangxiaoxuan273: moved

cmd/oras/root/tag.go
shizhMSFT: I see the `inner` is used later.

cmd/oras/internal/option/target.go
```diff
@@ -118,24 +116,30 @@ func (opts *Target) Parse() error {
 }
 
 // parseOCILayoutReference parses the raw in format of <path>[:<tag>|@<digest>]
-func parseOCILayoutReference(raw string) (path string, ref string, err error) {
+func (opts *Target) parseOCILayoutReference() error {
+	raw := opts.RawReference
+	var path string
+	var ref string
+
```
shizhMSFT, line 123: nit: remove empty line

cmd/oras/internal/option/target.go
```diff
@@ -118,24 +116,30 @@ func (opts *Target) Parse() error {
...
 		// `digest` found
 		path = raw[:idx]
 		ref = raw[idx+1:]
 	} else {
 		// find `tag`
+		var err error
 		path, ref, err = fileref.Parse(raw, "")
+		if err != nil {
+			err = errors.Join(err, errdef.ErrInvalidReference)
+			return err
```
shizhMSFT, line 134:
```suggestion
			return errors.Join(err, errdef.ErrInvalidReference)
```

shizhMSFT (approved):
LGTM with suggestions

shizhMSFT (approved):
LGTM

qweeah (approved):
LGTM

PR #1312, opened 2024-03-27
refactor: add stdout to metadata handlers

qweeah:
This PR adds stdout to the metadata handlers.


shizhMSFT (approved):
LGTM. I'd expect the next PR for add out to status handlers.

PR #1314, opened 2024-03-28
refactor: use command output in status output

qweeah:
This PR makes status output handlers use command output.


shizhMSFT (approved):
LGTM

TerryHowe:
Yeh, that is an improvement, thanks

PR #1315, opened 2024-03-28
refactor: make command accessible in option parser

### LINKED ISSUE #1301: refactor: make command accessible in option parser
Currently options are parsed without any argument

https://github.com/oras-project/oras/blob/e8da6e53ec3a679c966c948e1428ab6a72eee04f/cmd/oras/internal/option/parser.go#L22-L25

All option parser implementations are called in [command.PreRunE](https://pkg.go.dev/github.com/spf13/cobra#Command.PreRunE), where command is actually accessible. Should pass the command to parser to facilitate cmd-specific parsing, e.g. processing the flag set.

wangxiaoxuan273:

Fixes #1301 

### DIFF
--- a/cmd/oras/internal/option/common.go
+++ b/cmd/oras/internal/option/common.go
@@ -50,7 +50,7 @@ func (opts *Common) WithContext(ctx context.Context) (context.Context, logrus.Fi
 }
 
 // Parse gets target options from user input.
-func (opts *Common) Parse(_ *cobra.Command) error {
+func (opts *Common) Parse(*cobra.Command) error {
 	// use STDERR as TTY output since STDOUT is reserved for pipeable output
 	return opts.parseTTY(os.Stderr)
 }
--- a/cmd/oras/internal/option/common.go
+++ b/cmd/oras/internal/option/common.go
@@ -20,6 +20,7 @@ import (
 	"os"
 
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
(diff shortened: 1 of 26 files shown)


cmd/oras/internal/option/parser.go
```diff
@@ -17,18 +17,20 @@ package option
...
 // FlagParser parses flags in an option.
 type FlagParser interface {
-	Parse() error
+	Parse(cmd *cobra.Command) error
 }
 
 // Parse parses applicable fields of the passed-in option pointer and returns
 // error during parsing.
-func Parse(optsPtr interface{}) error {
+func Parse(optsPtr interface{}, cmd *cobra.Command) error {
```
shizhMSFT, line 31: Can we put `cmd` as the first argument for consistency?
```suggestion
func Parse(cmd *cobra.Command, optsPtr any) error {
```

cmd/oras/internal/option/common.go
```diff
@@ -49,7 +50,7 @@ func (opts *Common) WithContext(ctx context.Context) (context.Context, logrus.Fi
 }
 
 // Parse gets target options from user input.
-func (opts *Common) Parse() error {
+func (opts *Common) Parse(_ *cobra.Command) error {
```
qweeah, line 53: How about
```suggestion
func (opts *Common) Parse(*cobra.Command) error {
```

qweeah (approved):
LGTM with suggestion.

PR #1319, opened 2024-04-01
feat: add mTLS options

### LINKED ISSUE #491: X509 Authentication with OCI registry
I previously opened [this](https://github.com/helm/helm/issues/10798) issue in helm and was directed to this repo as it is upstream to helm's OCI registry features. 

However, I wasn't able to find a related issue here. Would anyone be able to point me in the right direction or link me to the updates to ORAS facilitating x509 mTLS authentication?

sabre1041:
Adds support for mTLS to remote registries

Resolves #491


cmd/oras/internal/option/remote.go
```diff
@@ -50,6 +50,8 @@ import (
 type Remote struct {
 	DistributionSpec
 	CACertFilePath    string
+	CertFilePath      string
+	KeyFilePath       string
```
FeynmanZhou, line 54: Should users pass the key and cert file to ORAS from local filesystem with `--CertFilePath` and `KeyFilePath`?

sabre1041: @FeynmanZhou similar to the `--ca-file` option,  its the location of the file(s) containing the client certificate and associated private key

qweeah: > Should users pass the cert and key file to ORAS from local filesystem with --CertFilePath and KeyFilePath?

User should pass `--cert-file <CertFilePath>` and `--key-file <KeyFilePath>` to ORAS

cmd/oras/internal/option/remote.go
```diff
@@ -191,6 +195,14 @@ func (opts *Remote) tlsConfig() (*tls.Config, error) {
 			return nil, err
 		}
 	}
+	if opts.CertFilePath != "" && opts.KeyFilePath != "" {
```
qweeah, line 198: What if user only provide partial input? Like only one cert or one key file is provided? Should we add error check to suggest user provide both?

Helm seems silently ignore partial input, which I think ORAS should enhance.

qweeah: @FeynmanZhou

qweeah (changes requested):

cmd/oras/internal/option/remote.go
```diff
@@ -191,6 +195,14 @@ func (opts *Remote) tlsConfig() (*tls.Config, error) {
...
+		var err error
```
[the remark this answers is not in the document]
sabre1041: Resolved

cmd/oras/internal/option/remote.go
TerryHowe:
```suggestion
	if opts.CertFilePath != "" || opts.KeyFilePath != "" {
```

I would expect it to work like that^, but I'm not sure how the LoadX509KeyPair works

sabre1041: @TerryHowe this option that we are enabling is for client/mtls auth. so we would expect to have both a certificate and key.

@TerryHowe @qweeah i did look into ways of enforcing both options are provided and we can look to leverage the `MarkFlagsMutuallyExclusive` of Cobra.

https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#flag-groups

The downside is that it would require a large number of changes since these are global options. I could emit an error in the `tlsConfig` function, but i would rather perform the validation early

Let me know what path I should go down.

qweeah: `MarkFlagsMutuallyExclusive` returns error when `opts.CertFilePath != "" && opts.KeyFilePath != ""`

Here we need to return error when `opts.CertFilePath != "" XOR opts.KeyFilePath != ""`

> it would require a large number of changes since these are global options.

You just need to add check into the `remote.Parse` and the check will be applied to all commands before execution.

https://github.com/oras-project/oras/blob/202bcf44e210ff51a1b2acf4383f80420ddaa485/cmd/oras/internal/option/remote.go#L142-L158

sabre1041: Perfect timing as this was added recently! Will add the check

sabre1041: @qweeah Updated to account for the requested changes. Please take a look whenever you have a moment

qweeah: LGTM

qweeah (approved):
LGTM

cmd/oras/internal/option/remote.go
```diff
@@ -120,7 +125,9 @@ func (opts *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
 	opts.plainHTTP = func() (bool, bool) {
 		return *plainHTTP, fs.Changed(plainHTTPFlagName)
 	}
-	fs.StringVar(&opts.CACertFilePath, opts.flagPrefix+"ca-file", "", "server certificate authority file for the remote "+notePrefix+"registry")
+	fs.StringVar(&opts.CACertFilePath, opts.flagPrefix+caFileFlag, "", "server certificate authority file for the remote "+notePrefix+"registry")
+	fs.StringVarP(&opts.CertFilePath, opts.flagPrefix+certFileFlag, "", "", "client certificate authority file for the remote "+notePrefix+"registry")
```
shizhMSFT, line 129: client should have a leaf cert instead of a CA cert.
```suggestion
	fs.StringVarP(&opts.CertFilePath, opts.flagPrefix+certFileFlag, "", "", "client certificate file for the remote "+notePrefix+"registry")
```

cmd/oras/internal/option/remote_test.go
```diff
@@ -43,9 +45,213 @@ var testTagList = struct {
 	Tags: []string{"tag"},
 }
 
+// localhostServerCert is a PEM-encoded TLS cert with SAN IPs
+// "127.0.0.1" and "[::1]", expiring at Jan 29 16:00:00 2084 GMT.
+// adapted from src/crypto/tls:
+// go run generate_cert.go  --rsa-bits 4096 --host 127.0.0.1,::1,oras.land --ca --start-date "Jan 1 00:00:00 1970" --duration=1000000h
```
shizhMSFT, line 51: Where can I find `generate_cert.go`?

shizhMSFT: Will it be better to replace it with an `openssl` command?

cmd/oras/internal/option/remote_test.go
```diff
@@ -43,9 +45,213 @@ var testTagList = struct {
...
+var localhostServerCert = []byte(`-----BEGIN CERTIFICATE-----
```
shizhMSFT, line 52: nit: Will it be better to store those testing key pairs as files in the `testdata` folder, and then reference using [`//go:embed`](https://pkg.go.dev/embed)?

sabre1041: Great suggestion! learned something new in the process

cmd/oras/internal/option/remote_test.go
sabre1041: Clarified in comment

cmd/oras/internal/option/remote.go
sabre1041: done

cmd/oras/internal/option/remote_test.go
```diff
@@ -43,9 +46,63 @@ var testTagList = struct {
...
+}
+
+func loadTestingCert(certificate, key []byte) tls.Certificate {
+	cert, err := tls.X509KeyPair(certificate, key)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to load testing certificate: %v", err))
+	}
+
+	return cert
+
```
shizhMSFT, line 100: nit: remove empty line

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1343, opened 2024-04-15
feat: make all go-template key lower-cased

### LINKED ISSUE #1342: unify the naming of all metadata fields lower cased

v1.2.0

ORAS users should be able to filter the attributes of the command metadata output via lower cased key names.

It's a widely applied and proven user-experience of `kubectl`.

qweeah:
This PR makes all formattable names of the metadata output  lower cased.

Fixes #1342

### DIFF
--- a/cmd/oras/internal/display/utils/json.go
+++ b/cmd/oras/internal/display/utils/json.go
@@ -42,3 +42,17 @@ func PrintJSON(out io.Writer, data []byte, pretty bool) error {
 	_, err := out.Write(data)
 	return err
 }
+
+// ToMap converts the data to a map[string]any with json tag as key.
+func ToMap(data any) (map[string]any, error) {
+	// slow but easy
+	content, err := json.Marshal(data)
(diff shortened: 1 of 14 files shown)


cmd/oras/internal/display/utils/json.go
```diff
@@ -42,3 +42,17 @@ func PrintJSON(out io.Writer, data []byte, pretty bool) error {
...
+	// slow but easy
+	content, err := json.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+	var ret map[string]any
+	if err = json.Unmarshal(content, &ret); err != nil {
+		ret = nil
+	}
+	return ret, err
```
shizhMSFT, line 57: nit: code style
```suggestion
	if err = json.Unmarshal(content, &ret); err != nil {
		return nil, err
	}
	return ret, nil
```

shizhMSFT (approved):
LGTM

PR #1345, opened 2024-04-15
feat: improve error message when detecting conflicting flags

### LINKED ISSUE #1340: Need to improve error message when basic auth flags are used with identity token flags
After #1294 is completed, `cmd.MarkFlagsMutuallyExclusive` is used to prevent `--identity-token` / `--identity-token-stdin` to be used with `--username`/ `--password` / `--password-stdin`. However, the default error message is confusing for the users. We need to improve the error messages.

wangxiaoxuan273:
Improve error message when detecting conflicting flags. 
Current display:
image

Fixes #1340 

### DIFF
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"oras.land/oras-go/v2/registry/remote/auth"
 	"oras.land/oras-go/v2/registry/remote/errcode"
 )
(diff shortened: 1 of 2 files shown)


qweeah (changes requested):

cmd/oras/internal/option/remote.go
```diff
@@ -149,14 +149,36 @@ func (opts *Remote) Parse(cmd *cobra.Command) error {
...
+		for _, flagName := range exclusiveFlagSet {
+			if cmd.Flags().Changed(flagName) {
+				changedFlags = append(changedFlags, flagName)
+			}
+		}
+		return changedFlags
+	}
+	for _, set := range exclusiveFlagSets {
+		changedFlags := checkConflict(set)
+		if len(changedFlags) >= 2 {
```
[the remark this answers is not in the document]
qweeah: Return all conflicted flags in a flag set with error like `--%s, --%s, --%s cannot be used at the same time`

cmd/oras/internal/option/remote.go
```diff
@@ -149,14 +149,36 @@ func (opts *Remote) Parse(cmd *cobra.Command) error {
...
+	}
 	if err := opts.parseCustomHeaders(); err != nil {
 		return err
 	}
 	return opts.readSecret(cmd)
 }
 
+// CheckMutuallyExclusiveFlags checks if any mutually exclusive flags are both
+// set, returns an error when detecting set exclusive flags.
+func CheckMutuallyExclusiveFlags(cmd *cobra.Command, exclusiveFlagSets ...[]string) error {
```
qweeah, line 163: Add a comment on how the argument `...[]string` will be used, e.g. it detects the first set of conflict and returns the error without checking all the flag sets.

cmd/oras/internal/errors/errors.go
```diff
@@ -156,3 +157,27 @@ func NewErrEmptyTagOrDigest(ref string, cmd *cobra.Command, needsTag bool) error
...
+			if fs.Changed(flagName) {
+				changedFlags = append(changedFlags, fmt.Sprintf("--%s", flagName))
+			}
+		}
+		return changedFlags
+	}
+	for _, set := range exclusiveFlagSets {
+		changedFlags := checkConflict(set)
+		if len(changedFlags) >= 2 {
+			flags := strings.Join(changedFlags, ", ")
```
qweeah, line 178: need extra loop to add `--` as prefix to **ALL** the changed flag。

cmd/oras/internal/errors/errors.go
```diff
@@ -156,3 +157,27 @@ func NewErrEmptyTagOrDigest(ref string, cmd *cobra.Command, needsTag bool) error
...
+			return fmt.Errorf("--%s cannot be used at the same time", flags)
```
qweeah, line 179: Do we still need `--`?
```suggestion
			return fmt.Errorf("%s cannot be used at the same time", flags)
```

qweeah, line 179:
```suggestion
			return fmt.Errorf("%s cannot be used at the same time", strings.Join(changedFlags, ", "))
```

wangxiaoxuan273: My mistake

qweeah (approved):
LGTM

PR #1349, opened 2024-04-16
chore(ux): optimize progress bar

qweeah:
This PR
1) Adds defensive code to avoid showing negative speed in progress bar.
2) Calculates the download speed based on the size of transferred data in the past last seconds.


cmd/oras/internal/display/status/progress/speed.go
```diff
@@ -0,0 +1,50 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package progress
+
+type speedWindow struct {
+	point []int64
```
shizhMSFT, line 19: According to your algorithm for accuracy, a point should be a pair of `(time, offset)`. Then the average time is $$\text{Average Time} = \frac{\text{head}.offset - \text{tail}.offset}{\text{head}.time - \text{tail}.time}$$.

shizhMSFT: We can discuss the algorithm offline if needed

shizhMSFT (changes requested):

cmd/oras/internal/display/status/progress/speed.go
qweeah: yes you are right, will change accordingly.

shizhMSFT (approved):
LGTM

PR #1355, opened 2024-04-17
refactor: move logger method to command

TerryHowe:

I think it makes a lot more sense that a command has-a logger rather than an option.


qweeah (approved):
LGTM, need to add doc comment for the public function.

qweeah (approved):
LGTM

PR #1358, opened 2024-04-18
feat: improve error message of when a set of flags is required to use together

### LINKED ISSUE #1357: improve error message of when a set of flags is required to use together
I am using ORAS built from https://github.com/oras-project/oras/commit/2edebad6a91eed27fac639d0b51f3e8c5f1aba3c and trying to mTLS to a registry running at `localhost:5000`.

If only a cert file is provided without key, I will get error like:
```console
> oras login localhost:5000 --cert-file test   
Error: if any flags in the group [cert-file key-file] are set they must all be set; missing [key-file]
```
The error message is hard to read since it does not explicitly tells what should be done to user, suggest to change it to:
```console
> oras login localhost:5000 --cert-file test   
Error: --key-file is required but not provided
```

wangxiaoxuan273:
improve error message of when a set of flags is required to use together

Fixes #1357 


cmd/oras/internal/errors/errors.go
```diff
@@ -173,3 +173,20 @@ func CheckMutuallyExclusiveFlags(fs *pflag.FlagSet, exclusiveFlagSet ...string)
...
+// returns an error when detecting any flags not used while other flags have been used.
+func CheckRequiredTogetherFlags(fs *pflag.FlagSet, requiredTogetherFlags ...string) error {
+	var unchangedFlags []string
+	for _, flagName := range requiredTogetherFlags {
+		if !fs.Changed(flagName) {
+			unchangedFlags = append(unchangedFlags, fmt.Sprintf("--%s", flagName))
+		}
+	}
+	nUnchangedFlags := len(unchangedFlags)
+	if nUnchangedFlags > 0 && nUnchangedFlags != len(requiredTogetherFlags) {
```
qweeah, line 187: Below condition is more aligned to non-or-all checking.
```suggestion
	if nUnchangedFlags != 0 && nUnchangedFlags != len(requiredTogetherFlags) {
```

cmd/oras/internal/errors/errors.go
```diff
@@ -173,3 +173,20 @@ func CheckMutuallyExclusiveFlags(fs *pflag.FlagSet, exclusiveFlagSet ...string)
...
+	var unchangedFlags []string
+	for _, flagName := range requiredTogetherFlags {
+		if !fs.Changed(flagName) {
+			unchangedFlags = append(unchangedFlags, fmt.Sprintf("--%s", flagName))
+		}
+	}
+	nUnchangedFlags := len(unchangedFlags)
+	if nUnchangedFlags != 0 && nUnchangedFlags != len(requiredTogetherFlags) {
+		flags := strings.Join(unchangedFlags, ", ")
+		return fmt.Errorf("%s required but not provided", flags)
```
qweeah, line 189: The original error is sth like:
> Error: if any flags in the group [cert-file key-file] are set they must all be set; missing [key-file]

Both set and unset flags are all mention, suggestion to follow the design and provide complete error message like:
```suggestion
		return fmt.Errorf("since %s are set, %s required but missing", changed, unchanged)
```

cmd/oras/internal/errors/errors.go
```diff
@@ -173,3 +173,20 @@ func CheckMutuallyExclusiveFlags(fs *pflag.FlagSet, exclusiveFlagSet ...string)
...
+
+// CheckRequiredTogetherFlags checks if any flags required together are all used,
+// returns an error when detecting any flags not used while other flags have been used.
+func CheckRequiredTogetherFlags(fs *pflag.FlagSet, requiredTogetherFlags ...string) error {
+	var unchangedFlags []string
+	for _, flagName := range requiredTogetherFlags {
+		if !fs.Changed(flagName) {
+			unchangedFlags = append(unchangedFlags, fmt.Sprintf("--%s", flagName))
+		}
+	}
```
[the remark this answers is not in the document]
qweeah: The implementation can returns two arrays to represent changed and unchanged flags respectively.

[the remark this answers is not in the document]
wangxiaoxuan273: Refactored this as a function

cmd/oras/internal/errors/errors.go
```diff
@@ -161,15 +161,34 @@ func NewErrEmptyTagOrDigest(ref string, cmd *cobra.Command, needsTag bool) error
...
 		return fmt.Errorf("%s cannot be used at the same time", flags)
 	}
 	return nil
 }
+
+// CheckRequiredTogetherFlags checks if any flags required together are all used,
+// returns an error when detecting any flags not used while other flags have been used.
+func CheckRequiredTogetherFlags(fs *pflag.FlagSet, requiredTogetherFlags ...string) error {
+	changed, unchanged := checkChangedFlags(fs, requiredTogetherFlags...)
+	nUnchangedFlags := len(unchanged)
```
qweeah, line 176: naming is hard, `unchangedCnt` sounds more like a better one to me.
```suggestion
	unchangedCnt := len(unchanged)
```

cmd/oras/internal/errors/errors.go
```diff
@@ -161,15 +161,34 @@ func NewErrEmptyTagOrDigest(ref string, cmd *cobra.Command, needsTag bool) error
...
+	if nUnchangedFlags != 0 && nUnchangedFlags != len(requiredTogetherFlags) {
+		changed := strings.Join(changed, ", ")
+		unchanged := strings.Join(unchanged, ", ")
+		return fmt.Errorf("%s set, %s required but not provided", changed, unchanged)
```
qweeah, line 180: Since we are going to skip verbs, how about:
```suggestion
		return fmt.Errorf("%s must be used in conjunction with %s.", changed, unchanged)
```

qweeah (approved):
LGTM

cmd/oras/internal/errors/errors_test.go
```diff
@@ -53,3 +53,36 @@ func TestCheckMutuallyExclusiveFlags(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckRequiredTogetherFlags(t *testing.T) {
+	fs := &pflag.FlagSet{}
```
qweeah, line 58: Should add a case to check the case when none required flags are used. E.g. `--hello` and `--world` are required to use together but none of it is changed, `CheckRequiredTogetherFlags` should not return error.

wangxiaoxuan273: added a test

PR #1362, opened 2024-04-19
chore: project selected fields from OCI descriptor

qweeah:
This PR limits the selectable fields of a descriptor in json and template output.


shizhMSFT:
> Would it be easier to exclude fields you don't want?

I think OCI image-spec may add more fields in the future. Therefore, an allow list would be easier.

shizhMSFT (approved):
LGTM

PR #1372, opened 2024-05-05
fix: Replace obsolete html element

TerryHowe:

The align element was made obsolete in html 5


qweeah (approved):
LGTM

PR #1373, opened 2024-05-05
refactor: Move ShortDigest out of display package

TerryHowe:

The ShortDigest method ideally would be a method of descriptor, but that is not part of this project, so it should live in some package related to that object.


qweeah (approved):
LGTM

PR #1375, opened 2024-05-06
refactor: No error return for print

TerryHowe:

I don't think we should consider print errors a problem most of the time and we should manage the output of the error to print only once.

### DIFF
--- a/cmd/oras/internal/display/status/print.go
+++ b/cmd/oras/internal/display/status/print.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"oras.land/oras/internal/descriptor"
+	"os"
 	"sync"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -45,7 +46,12 @@ func (p *Printer) Println(a ...any) error {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 	_, err := fmt.Fprintln(p.out, a...)
-	return err
+	if err != nil {
+		err = fmt.Errorf("display output error: %w", err)
+		_, _ = fmt.Fprint(os.Stderr, err)
+	}
+	// Errors are handled above, so return nil
+	return nil
 }
 
 // PrintStatus prints transfer status.
--- a/cmd/oras/internal/display/status/print_test.go
+++ b/cmd/oras/internal/display/status/print_test.go
@@ -1,0 +1,65 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package status
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+type mockWriter struct {
+	errorCount int
+	written    string
+}
+
+func (mw *mockWriter) Write(p []byte) (n int, err error) {
+	mw.written += string(p)
+	if strings.TrimSpace(string(p)) != "boom" {
+		return len(string(p)), nil
+	}
+	mw.errorCount++
+	return 0, fmt.Errorf("Boom: " + string(p))
+}
+
+func (mw *mockWriter) String() string {
+	return mw.written
+}
+
+func TestPrint_Error(t *testing.T) {
+	mockWriter := &mockWriter{}
+	printer := NewPrinter(mockWriter)
+	printer.Println("boom")
+	if mockWriter.errorCount != 1 {
+		t.Error("Expected one errors actual <" + strconv.Itoa(mockWriter.errorCount) + ">")
+	}
+}
+
+func TestPrint_NoError(t *testing.T) {
+	mockWriter := &mockWriter{}
+	printer := NewPrinter(mockWriter)
+
+	expected := "blah blah"
+	printer.Println(expected)
+	actual := strings.TrimSpace(mockWriter.String())
+	if expected != actual {
+		t.Error("Expected <" + expected + "> not equal to actual <" + actual + ">")
+	}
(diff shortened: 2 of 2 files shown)


qweeah (changes requested):

cmd/oras/internal/display/status/deprecated.go
```diff
@@ -58,7 +58,8 @@ var printer = NewPrinter(os.Stdout)
 
 // Print objects to display concurrent-safely.
 func Print(a ...any) error {
-	return printer.Println(a...)
+	printer.Println(a...)
```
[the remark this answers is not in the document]
TerryHowe: Some errors might resolve themselves like ENOSPACE could be ephemeral

cmd/oras/internal/display/status/print.go
```diff
@@ -41,11 +43,16 @@ func NewPrinter(out io.Writer) *Printer {
...
 
 // Println prints objects concurrent-safely with newline.
-func (p *Printer) Println(a ...any) error {
+func (p *Printer) Println(a ...any) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 	_, err := fmt.Fprintln(p.out, a...)
-	return err
+	if err != nil {
+		if p.errored == nil {
```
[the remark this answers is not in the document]
TerryHowe: Done

cmd/oras/internal/display/status/deprecated.go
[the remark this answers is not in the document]
qweeah: I will resolve this comment and continue the discussion in another comment.

cmd/oras/internal/display/status/print.go
```diff
@@ -41,11 +43,14 @@ func NewPrinter(out io.Writer) *Printer {
 }
 
 // Println prints objects concurrent-safely with newline.
-func (p *Printer) Println(a ...any) error {
+func (p *Printer) Println(a ...any) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 	_, err := fmt.Fprintln(p.out, a...)
-	return err
+	if err != nil && p.errored == nil {
```
qweeah, line 50: Say we have an error `A`, which auto resolved and we have another error `B/

Should `A` and `B` be output to stderr? If so then we need to change it to
```suggestion
	if err != nil {
```

qweeah: If `&& p.errored == nil` is used then `errored` will always be wrapped `A` but not changed to `B`

qweeah:
If `&& p.errored == nil` is not required, then `errored` can be removed from the printer like:
```golang
// Println prints objects concurrent-safely with newline.
func (p *Printer) Println(a ...any) error {
	p.lock.Lock()
	defer p.lock.Unlock()
	_, err := fmt.Fprintln(p.out, a...)
	if err != nil  {
		err = fmt.Errorf("display output error: %w", err)
		_, _ = fmt.Fprint(os.Stderr, err)
	}
        return err
}
```

cmd/oras/internal/display/status/print.go
TerryHowe: It be very unusual to have A not the same as B and probably not worth distinguishing. I'm okay with last error wins, but I'd like to report it only once and it is easier to just report the first.

TerryHowe:
Last commit only contains improved test errors

cmd/oras/internal/display/status/print.go
qweeah: To me the most reasonable solution is to report all the errors.

TerryHowe: Done

qweeah: If we agree to report all the errors, should the error be returned directly via `printer.Println`?

qweeah: like
```golang
// Println prints objects concurrent-safely with newline.
func (p *Printer) Println(a ...any) error {
	p.lock.Lock()
	defer p.lock.Unlock()
	_, err := fmt.Fprintln(p.out, a...)
	if err != nil  {
		err = fmt.Errorf("display output error: %w", err)
		_, _ = fmt.Fprint(os.Stderr, err)
	}
        return err
}
```

qweeah: so `errored` is no longer needed.

TerryHowe: Sounds good to me

cmd/oras/internal/display/status/deprecated.go
```diff
@@ -58,7 +58,8 @@ var printer = NewPrinter(os.Stdout)
...
+	return nil
```
qweeah, line 62: Need documentation here to explain why and how the print errors are ignored.

cmd/oras/internal/display/status/print.go
```diff
@@ -58,7 +62,8 @@ func (p *Printer) PrintStatus(desc ocispec.Descriptor, status string, verbose bo
 		}
 		name = desc.MediaType
 	}
-	return p.Println(status, descriptor.ShortDigest(desc), name)
+	p.Println(status, descriptor.ShortDigest(desc), name)
+	return nil
```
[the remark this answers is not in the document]
TerryHowe: `p.Println` no longer returns anything.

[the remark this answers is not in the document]
qweeah: Sorry I missed that. In my mind we should:
1) make `printer.Println` always return `nil`
2) use `return printer.Println` in `printer.PrintStatus` and `Print(a ...any)`

So the error ignoring in only implemented in `printer.Println`. In the future, if we want to change this behavior we only need to change `printer.Println`

cmd/oras/internal/display/status/print.go
```diff
@@ -41,11 +42,14 @@ func NewPrinter(out io.Writer) *Printer {
...
+func (p *Printer) Println(a ...any) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 	_, err := fmt.Fprintln(p.out, a...)
-	return err
+	if err != nil {
+		err = fmt.Errorf("display output error: %w", err)
+		_, _ = fmt.Fprint(os.Stderr, err)
+	}
 }
```
qweeah, line 53: Need to return nil here and add comment to explain why and how the print errors are ignored.

TerryHowe: I think I've handled all your concerns

qweeah (approved):
LGTM

PR #1377, opened 2024-05-08
feat: add `--template` and support `--format go-template=TEMPLATE` experience

qweeah:
This PR implements latest change in #1199


cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+		Recommendation: fmt.Sprintf("supported types: %s", strings.Join(optionalTypes, ", ")),
+	}
+}
+
+func (opts *Format) parseFlag() error {
+	opts.Type = opts.FormatFlag
+	if opts.Template != "" {
+		// template explicitly set
+		if opts.Type != FormatTypeGoTemplate {
+			return fmt.Errorf("--template must be used with --format %s", FormatTypeGoTemplate)
```
FeynmanZhou, line 107: There is a case that we may also consider add error catch and error message:

If I use `--format go-template` only (without `--template`), the default output and verbose log are empty. See examples below.

It would be reasonable to print an error message `Error: you should use "--format go-template=GO_TEMPLATE"` to prompt users that this is a wrong usage and how to use this flag.

```
$ oras manifest fetch localhost:5000/my-ubuntu:latest --format go-template

$ oras manifest fetch localhost:5000/my-ubuntu:latest --format go-template -v
```

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+	// generate usage string
+	maxLength := 0
+	for _, option := range opts.types {
+		if len(option.Name) > maxLength {
+			maxLength = len(option.Name)
+		}
+	}
+	for _, option := range opts.types {
+		usage += fmt.Sprintf("\n'%s':%s%s", option.Name, strings.Repeat(" ", maxLength-len(option.Name)+2), option.Usage)
+	}
```
shizhMSFT, line 70: Will [`text/tabwriter`](https://pkg.go.dev/text/tabwriter) be better?

cmd/oras/internal/option/format.go
qweeah: Will align the error reporting to `kubectl`
```console
> kubectl get all --output go-template
error: template format specified but no template given
> kubectl get all --output go-template=""
error: template format specified but no template given
```

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+
+// FormatError generates the error message for an invalid type.
+func (opts *Format) TypeError() error {
+	return fmt.Errorf("unsupported format type: %s", opts.Type)
+}
+
+// RawFormatFlag returns raw input of --format flag.
+func (opts *Format) RawFormatFlag() string {
+	return opts.FormatFlag
 }
```
shizhMSFT, line 141: This method seems redundant.

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+// Caller should make sure that this function is used before applying flags.
+func (opts *Format) SetTypesAndDefault(defaultType string, types []FormatType) {
+	opts.FormatFlag = defaultType
+	opts.types = types
+}
+
+// FormatError generates the error message for an invalid type.
+func (opts *Format) TypeError() error {
+	return fmt.Errorf("unsupported format type: %s", opts.Type)
+}
```
shizhMSFT, line 136: This error generation should be in the `display` package

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+func (opts *Format) SetTypes(types []FormatType) {
+	opts.types = types
+}
+
+// SetTypesAndDefault resets the format options and default value.
+// Caller should make sure that this function is used before applying flags.
+func (opts *Format) SetTypesAndDefault(defaultType string, types []FormatType) {
+	opts.FormatFlag = defaultType
+	opts.types = types
+}
```
shizhMSFT, line 131: Those APIs seems redundant.

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+		return nil
+	}
+
+	goTemplatePrefix := FormatTypeGoTemplate + "="
+	if strings.HasPrefix(opts.FormatFlag, goTemplatePrefix) {
+		// add parameter to template
+		opts.Type = FormatTypeGoTemplate
+		opts.Template = opts.FormatFlag[len(goTemplatePrefix):]
+	}
+	return nil
```
shizhMSFT, line 118: This part should be checked first

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,127 @@ limitations under the License.
...
+	oerrors "oras.land/oras/cmd/oras/internal/errors"
+)
+
+const (
+	// format types
+	FormatTypeJSON       = "json"
+	FormatTypeTree       = "tree"
+	FormatTypeTable      = "table"
+	FormatTypeGoTemplate = "go-template"
+)
```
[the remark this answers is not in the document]
qweeah: Yes, same style with `TargetTypeRemote` and `TargetTypeOCILayout`

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,120 @@ limitations under the License.
...
-	if fs.Lookup(name) != nil {
-		// allow command to overwrite the flag
-		return
-	}
-	fs.StringVar(&opts.Template, name, "", `[Experimental] Format output using a custom template:
-'json':       Print in JSON format
-'$TEMPLATE':  Print output using the given Go template.`)
+	var buf bytes.Buffer
+	_, _ = buf.WriteString("[Experimental] Format output using a custom template:")
+	w := tabwriter.NewWriter(&buf, 0, 0, 2, ' ', 0)
```
shizhMSFT, line 75: nit:
```suggestion
	buf := bytes.NewBufferString("[Experimental] Format output using a custom template:")
	w := tabwriter.NewWriter(buf, 0, 0, 2, ' ', 0)
```

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,120 @@ limitations under the License.
...
+	}
+
+	goTemplatePrefix := FormatTypeGoTemplate.Name + "="
+	if strings.HasPrefix(opts.FormatFlag, goTemplatePrefix) {
+		// add parameter to template
+		opts.Type = FormatTypeGoTemplate.Name
+		opts.Template = opts.FormatFlag[len(goTemplatePrefix):]
+	}
+	return nil
 }
```
shizhMSFT, line 134: nit: this is accepted for now. It requires refactoring if we want to support more type-value pairs in the future like `--format go-template-file=<path>`.

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,120 @@ limitations under the License.
...
+
+	if opts.Type == "" {
+		// flag not specified
+		return nil
+	}
+
+	if opts.Type == FormatTypeGoTemplate.Name && opts.Template == "" {
+		return &oerrors.Error{
+			Err:            fmt.Errorf("%q format specified but no template given", opts.Type),
+			Recommendation: fmt.Sprintf("use `--format %s=TEMPLATE` or `--format %s --template TEMPLATE` to specify the template", opts.Type, opts.Type),
```
shizhMSFT, line 99: We should only recommend the primary usage
```suggestion
			Recommendation: fmt.Sprintf("use `--format %s=TEMPLATE` to specify the template", opts.Type, opts.Type),
```

cmd/oras/internal/option/format.go
```diff
@@ -15,21 +15,120 @@ limitations under the License.
...
+
+// format types
+var (
+	FormatTypeJSON = &FormatType{
+		Name:  "json",
+		Usage: "Print in JSON format",
+	}
+	FormatTypeGoTemplate = &FormatType{
+		Name:  "go-template",
+		Usage: "Print in JSON format",
```
shizhMSFT, line 51: This usage seems wrong. It is not for go template.

cmd/oras/root/discover.go
```diff
@@ -78,29 +78,34 @@ Example - Discover referrers of the manifest tagged 'v1' in an OCI image layout
...
 			}
+			opts.RawReference = args[0]
+			if err := option.Parse(cmd, &opts); err != nil {
+				return err
+			}
 			if cmd.Flags().Changed("output") {
-				switch opts.Template {
+				switch opts.Format.Type {
 				case "tree", "json", "table":
 					fmt.Fprintf(cmd.ErrOrStderr(), "[DEPRECATED] --output is deprecated, try `--format %s` instead\n", opts.Template)
```
shizhMSFT, line 88: The error message is not updated
```suggestion
					fmt.Fprintf(cmd.ErrOrStderr(), "[DEPRECATED] --output is deprecated, try `--format %s` instead\n", opts.Format.Type)
```

cmd/oras/internal/option/format.go
qweeah: Added todo comments

shizhMSFT (approved):
LGTM

Issue #1226, opened 2024-01-03
Support formatted JSON output for oras copy


FeynmanZhou:
### What is the version of your ORAS CLI

ORAS v1.1.0

### What would you like to be added?

When writing the feature spec [ORAS formatted output](https://github.com/oras-project/oras/pull/1199), the `--format` flag is not considered in `oras copy` because the scenario is not common and clear enough.

We do see users want to find out all the referrers that did get copied and their `digest` and `artifactType` that got copied over in the output. But this feature request has not been justified as a common scenario in the community and can even be achieved by using formatted JSON output of `oras discover`.

This issue is created for gathering more potential use cases of using formatted JSON output with `oras copy`. If you have any requirements or scenarios related to this feature request, feel free to comment on this issue.

### Why is this needed for ORAS?

N/A

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

PR #1378, opened 2024-05-08
refactor: Get rid of deprecated PrintStatus method

TerryHowe:

Remove last reference to deprecated PrintStatus method


TerryHowe:
Uggh, those tests aren't run on osx

qweeah (approved):
LGTM. BTW `cp.go` is also deprecated and should be refactored in #1226.

PR #1380, opened 2024-05-14
bump: update Golang to v1.22.3

qweeah:
This PR bumps all used Golang to v1.22.3


go.mod
```diff
@@ -1,6 +1,6 @@
 module oras.land/oras
 
-go 1.22.0
```
shizhMSFT, line 3: We don't need to update the minimum go version in go mod, do we?

qweeah: You mean patch version? Tried `go mod init test` and it gives me patch version go.mod
```go.mod
module test

go 1.22.1
```

qweeah: So yes we should update go patch version.

shizhMSFT (approved):
LGTM

PR #1383, opened 2024-05-17
fix: remove non-classic snap plugins

qweeah:
This PR added a fix to remove non-classic snap plugins in the snap build configuration. Will resolve [publish errors](https://github.com/oras-project/oras/actions/runs/9124299668) in the snap release


shizhMSFT (approved):
LGTM

PR #1384, opened 2024-05-18
fix: oras cp documentation

TerryHowe:
This was the only difference I thought should be corrected in the CLI as far as the documentation


qweeah (approved):
LGTM

Issue #1368, opened 2024-04-24
Have an option to verify the locally pulled files


stmlange:
### What is the version of your ORAS CLI

1.1.0

### What would you like to be added?

Maybe I'm missing it, but assume I have a local copy of some pulled data. Is there an option to run some sha256check or something to verify that the local copy matches with the artifacts listed in the manifest?

### Why is this needed for ORAS?

The manifest can have multiple digest encoding's which can make it very tricky to manually verify if the local copy is what is equal to the remote artifact.

### Are you willing to submit PRs to contribute to this feature?

- [X] Yes, I am willing to implement it.

qweeah:
@stmlange Can you kindly explain your scenario and why the verification is needed?

stmlange:
@qweeah Thanks for reaching out. The main reason why I have created this ticket is that there doesn't seem to be an (easy) option to verify/check if the locally pulled artifacts are "equal" to what the remove artifact is.

Try to answer the question: is what I have locally really what was published to remote?

Consider for example maven/gradle that publish dedicated sha1 and md5 hashsums so one can download the hashsum and verify somehow that the published thing is "correct"/"equal" to the local variant.

With oras one can have multiple digest variants encoded (https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests).

A digest can be `sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b` or `sha512:401b09eab3c013d4ca54922bb802bec8fd5318192b0a75f201d8b372742`...or whatever ORAS supports. Due to multiple supported hashalgorithms it is therefore not trivial to manually check if the downloaded artifact is actually what was published.

Running oras pull multiple times actually seems to re-download the artifact. So it may be possible that oras pull secretly checks those digests (and potentially fails if the download was not successfull), but redownloading is a waste of network resources.

qweeah:
Thanks @stmlange for the detailed explanation. 
You can utilize `oras manifest fetch` generate a checksum file and use `checksum -a 256 -c $FILE` to check it.

/cc @FeynmanZhou To validate if it's something we should add to ORAS CLI.

qweeah:
@stmlange Also worth mentioning that if you want to copy an artifact in a trusted way, why not using an [OCI image layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)?

```console
[...]

PR #1386, opened 2024-05-21
doc: verify local files

qweeah:
This PR adds doc on how to use format output to verify local files without pulling the whole artifact. Related to #1368.


docs/proposals/formatted-output.md
```diff
@@ -75,6 +75,14 @@ jobs:
           cat $PATH
 ```
 
+### Verify local files
+
+Carol is a build engineer who needs to check whether some locally copied files are identical to the remote artifact content. she can utilize `oras manifest fetch` to generate a checksum file based on the manifest without pulling the whole artifact. (See related issue [here](https://github.com/oras-project/oras/issues/1368)).
+
+```bash
+oras manifest fetch $REGISTRY/$REPO:$TAG --format go-template='{{range .content.layers}}{{if index .annotations "org.opencontainers.image.title"}}{{(split ":" .digest)._1}}  {{index .annotations "org.opencontainers.image.title"}}{{println}}{{end}}{{end}}' | shasum -c
```
shizhMSFT, line 83: What's the default algorithm for `shasum`? Should we set it to `shasum -a 256`?

qweeah: From usage perspective, `shasum` detects algorithm based on the length of the sum, e.g. it can check a file containing both sha256 and sha512 check sum
```console
> echo 256 > 256
> echo 512 > 512
> shasum -a 256 256 >> checksum
> shasum -a 512 512 >> checksum
> cat checksum
25bfed1cc9e0857d2ba433f97e50d92204946c3deac5adfbb917afa39d5132b18ec8da90352dec41c089962f83a18cbfd7174f5fcabc5474d96c949b7d11e324  512
f16c302d5d30e1d3fbe955cf4f637f58a871adeba597922e3baad0aeeb13f656  256
> shasum -c checksum
512: OK
256: OK
```

shizhMSFT (approved):
LGTM

PR #1389, opened 2024-05-26
chore: Remove deprecated PrintStatus method

TerryHowe:

Removes deprecated PrintStatus method


qweeah (approved):
LGTM. How could the patch coverage be 73.33% with 0 line missed?

shizhMSFT (approved):
LGTM

PR #1391, opened 2024-05-27
chore: Remove new tag status hint printer

TerryHowe:

Removes the deprecated NewTagStatusHintPritner


qweeah (commented):
Ideally, `NewTagStatusHintPrinter` in the `deprecated.go` file should be refactored into tag handlers in `display` package. (And tag command should be built upon the refactored handler types)

In this case, I would suggest not merging this PR so all the to-be-refactored code are all in one place.

PR #1392, opened 2024-05-27
fix: remove call to deprecated print

TerryHowe:

I missed this call in the last pull request. The printer is already available for manifest push.


qweeah (approved):
LGTM

PR #1396, opened 2024-05-29
chore: Add complicated blob fetch example from docs

TerryHowe:

This example was in the docs, but not in the CLI


cmd/oras/root/blob/fetch.go
```diff
@@ -67,6 +67,11 @@ Example - Fetch and print a blob from OCI image layout folder 'layout-dir':
 
 Example - Fetch and print a blob from OCI image layout archive file 'layout.tar':
   oras blob fetch --oci-layout --output - layout.tar@sha256:9a201d228ebd966211f7d1131be19f152be428bd373a92071c71d8deaf83b3e5
+
+Example - Fetch a blob with a specific reference, determine its size using 'jq', and use 'pv' to display the progress while redirecting the output to a file named 'layer.bin':
```
qweeah, line 71: Do we still need it since blob fetch has already supported progress bar natively in v1.2.0?

shizhMSFT: No. You are right that v1.2.0 supports showing progress.

TerryHowe: Sounds good

PR #1397, opened 2024-05-29
chore: Add format type text

TerryHowe:

Introduce `text` as default format type for `pull`, `push`, `manifest fetch` and `attach` commands.

The only change in E2E experience is in the help document. Taking `oras attach` as an example:

**Before**
```console
> oras attach -h
...
--format string                               [Experimental] Format output using a custom template:
                                                    'json':         Print in JSON format
                                                    'go-template':  Print output using the given Go template
...
```

**After**
```console
> oras attach -h
...
      --format string                               [Experimental] Format output using a custom template:
                                                    'json':         Print in JSON format
                                                    'go-template':  Print output using the given Go template
                                                    'text':         Print in text format (default "text")
...
```

### DIFF
--- a/cmd/oras/internal/display/handler.go
+++ b/cmd/oras/internal/display/handler.go
@@ -39,7 +39,7 @@ func NewPushHandler(out io.Writer, format option.Format, tty *os.File, verbose b
 	var statusHandler status.PushHandler
 	if tty != nil {
 		statusHandler = status.NewTTYPushHandler(tty)
-	} else if format.Type == "" {
+	} else if format.Type == option.FormatTypeText.Name {
 		statusHandler = status.NewTextPushHandler(out, verbose)
 	} else {
 		statusHandler = status.NewDiscardHandler()
@@ -47,7 +47,7 @@ func NewPushHandler(out io.Writer, format option.Format, tty *os.File, verbose b
 
 	var metadataHandler metadata.PushHandler
 	switch format.Type {
-	case "":
+	case option.FormatTypeText.Name:
 		metadataHandler = text.NewPushHandler(out)
 	case option.FormatTypeJSON.Name:
 		metadataHandler = json.NewPushHandler(out)
@@ -64,7 +64,7 @@ func NewAttachHandler(out io.Writer, format option.Format, tty *os.File, verbose
 	var statusHandler status.AttachHandler
 	if tty != nil {
 		statusHandler = status.NewTTYAttachHandler(tty)
-	} else if format.Type == "" {
+	} else if format.Type == option.FormatTypeText.Name {
 		statusHandler = status.NewTextAttachHandler(out, verbose)
 	} else {
 		statusHandler = status.NewDiscardHandler()
@@ -72,7 +72,7 @@ func NewAttachHandler(out io.Writer, format option.Format, tty *os.File, verbose
 
 	var metadataHandler metadata.AttachHandler
 	switch format.Type {
-	case "":
+	case option.FormatTypeText.Name:
 		metadataHandler = text.NewAttachHandler(out)
 	case option.FormatTypeJSON.Name:
 		metadataHandler = json.NewAttachHandler(out)
@@ -89,7 +89,7 @@ func NewPullHandler(out io.Writer, format option.Format, path string, tty *os.Fi
 	var statusHandler status.PullHandler
 	if tty != nil {
 		statusHandler = status.NewTTYPullHandler(tty)
-	} else if format.Type == "" {
+	} else if format.Type == option.FormatTypeText.Name {
 		statusHandler = status.NewTextPullHandler(out, verbose)
 	} else {
 		statusHandler = status.NewDiscardHandler()
@@ -97,7 +97,7 @@ func NewPullHandler(out io.Writer, format option.Format, path string, tty *os.Fi
 
 	var metadataHandler metadata.PullHandler
 	switch format.Type {
-	case "":
+	case option.FormatTypeText.Name:
 		metadataHandler = text.NewPullHandler(out)
 	case option.FormatTypeJSON.Name:
 		metadataHandler = json.NewPullHandler(out, path)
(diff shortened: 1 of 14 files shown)


cmd/oras/internal/option/format.go
```diff
@@ -64,6 +64,10 @@ var (
 		Name:  "tree",
 		Usage: "Get referrers recursively and print in tree format",
 	}
+	FormatTypeText = &FormatType{
+		Name:  "",
+		Usage: "Print in text format",
```
qweeah, line 69: This `Usage` is not used, are you intended to print it in the help doc?

qweeah (changes requested):
Just to clarify: For this PR, which of the following is matches your expectation:
1) Make the format selection code more readable, or
2) Define the behavior of text output in command help doc.  E.g. for v.1.2.0 the help doc of `oras attach` is 
```console
> oras discover -h
...
--format string                               [Experimental] Format output using a custom template:
                                                    'json':         Print in JSON format
                                                    'go-template':  Print output using the given Go template
...
```

And you are trying to change it to:
```console
> oras discover -h
...
--format string                               [Experimental] Format output using a custom template:
                                                    '':             Print in text format
                                                    'json':         Print in JSON format
                                                    'go-template':  Print output using the given Go template
...
```

cmd/oras/internal/option/format.go
```diff
@@ -74,16 +78,16 @@ type Format struct {
...
 	buf := bytes.NewBufferString("[Experimental] Format output using a custom template:")
 	w := tabwriter.NewWriter(buf, 0, 0, 2, ' ', 0)
 	for _, t := range opts.AllowedTypes {
 		_, _ = fmt.Fprintf(w, "\n'%s':\t%s", t.Name, t.Usage)
 	}
-	w.Flush()
+	_ = w.Flush()
 	// apply flags
-	fs.StringVar(&opts.FormatFlag, "format", opts.FormatFlag, buf.String())
+	fs.StringVar(&opts.FormatFlag, "format", FormatTypeText.Name, buf.String())
```
qweeah, line 90: This is wrong, some command doesn't have text output, e.g. `oras discover`. Should only set the default value to `FormatTypeText.Name` only when `opts.FormatFlag` is not changed.

cmd/oras/internal/display/handler.go
```diff
@@ -113,7 +113,7 @@ func NewPullHandler(out io.Writer, format option.Format, path string, tty *os.Fi
 func NewDiscoverHandler(out io.Writer, format option.Format, path string, rawReference string, desc ocispec.Descriptor, verbose bool) (metadata.DiscoverHandler, error) {
 	var handler metadata.DiscoverHandler
 	switch format.Type {
-	case option.FormatTypeTree.Name, "":
```
qweeah, line 116: I think we can remove `""` since it's just to be defensive on a non-existent code path.
```suggestion
	case option.FormatTypeTree.Name:
```

qweeah (changes requested):
Should exclude `FormatTypeText` from `oras discover`

cmd/oras/root/discover.go
```diff
@@ -105,6 +105,7 @@ Example - Discover referrers of the manifest tagged 'v1' in an OCI image layout
 		option.FormatTypeTable,
 		option.FormatTypeJSON.WithUsage("Get direct referrers and output in JSON format"),
 		option.FormatTypeGoTemplate.WithUsage("Print direct referrers using the given Go template"),
+		option.FormatTypeText,
```
qweeah, line 108:
```suggestion
```

qweeah:
Since this PR unifies default value for format type, I am thinking maybe we should take a step further and add a new function to set default and optional allowed types, e.g.
```golang
// Format contains input and parsed options for formatted output flags.
type Format struct {
	FormatFlag   string
	Type         string
	Template     string
	defaultType  *FormatType
	allowedTypes []*FormatType
}

// SetTypes sets the default format type and allowed format types.
func (opts *Format) SetTypes(defaultType *FormatType, allowedTypes ...*FormatType) {
	opts.defaultType = defaultType
	opts.allowedTypes = append(allowedTypes, defaultType)
}
```

So taking `oras attach` as an example, code changes in command with default text output would be like:
```diff
- opts.AllowedTypes = []*option.FormatType{option.FormatTypeJSON, option.FormatTypeGoTemplate}
+ opts.SetTypes(option.FormatTypeText, option.FormatTypeJSON, option.FormatTypeGoTemplate)
```

TerryHowe (commented):
/lgtm

Nice, thanks!

cmd/oras/internal/option/format.go
```diff
@@ -64,26 +64,41 @@ var (
...
 	FormatFlag   string
 	Type         string
 	Template     string
-	AllowedTypes []*FormatType
+	allowedTypes []*FormatType
+	defaultType  *FormatType
+}
+
+// SetTypes sets the default format type and allowed format types.
+func (f *Format) SetTypes(defaultType *FormatType, allowedTypes ...*FormatType) {
```
shizhMSFT, line 83: nit:
```suggestion
func (f *Format) SetTypes(defaultType *FormatType, otherTypes ...*FormatType) {
```

cmd/oras/internal/option/format.go
```diff
@@ -64,26 +64,41 @@ var (
...
-	for _, t := range opts.AllowedTypes {
+	for _, t := range opts.allowedTypes {
 		_, _ = fmt.Fprintf(w, "\n'%s':\t%s", t.Name, t.Usage)
 	}
-	w.Flush()
+	_ = w.Flush()
+	defaultTypeName := FormatTypeText.Name
+	if opts.defaultType != nil {
+		defaultTypeName = opts.defaultType.Name
+	}
```
shizhMSFT, line 99: Just curious when will `defaultType` be `nil`?

shizhMSFT (approved):
LGTM with suggestions and questions.

cmd/oras/internal/option/format.go
TerryHowe: Done.

cmd/oras/internal/option/format.go
TerryHowe: Only if someone didn't call SetTypes is all I can guess.

qweeah: Terry is right. Double checked the implementation and it seems that we can remove the type and store default value into `FormatFlag`. Refactored and repushed.

qweeah (approved):
LGTM

Issue #1041, opened 2023-08-02
Consider removing the --skip-delete-referrers   


sajayantony:
### What is the version of your ORAS CLI

1.1.0-rc.1+unreleased

### What would you like to be added?

The '--skip-delete-referrers' flag in the ORAS CLI currently serves as a fallback mechanism to avoid deleting referrers until all registries fully support the referrers API. However, with the increasing adoption of the referrers API among registries, the recommendation here is to remove this flag to streamline the CLI.

### Why is this needed for ORAS?

Eliminating Fallback Mechanism: As registries widely adopt the referrers API, using a fallback mechanism becomes unnecessary. Removing the flag will align the ORAS CLI with modern registry standards and encourage users to embrace the benefits of the referrers API, such as improved storage management and performance.

Non-Zero Exit Code Issue: The current behavior of the ORAS CLI, where deleting referrers using the tag schema results in a non-zero exit code, can cause problems in CI/CD pipelines. By removing the flag, we ensure a consistent zero exit code, providing clearer feedback to users and preventing false pipeline failures.

Alignment with Registry Behavior: Registries already handle tag updates by attaching new manifests without deleting old ones. Removing the '--skip-delete-referrers' flag will allow the ORAS CLI to align seamlessly with this standard registry behavior, promoting consistency and reducing potential disruptions.

Recommended Solution:
To enhance the ORAS CLI's usability and ensure it stays coordinated with modern registry practices, we propose removing the '--skip-delete-referrers' flag. This will simplify the CLI, avoid non-zero exit code issues, and align it with how registries handle tag updates. Additionally, we can update the documentation to inform users about this change during the change release. 

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

qweeah:
LGTM. BTW this issue actually proposes two changes:
1) remove `--skip-delete-referrers`
2) set default behavior to `--skip-delete-referrers=true`, which means old referrers will not be cleaned by oras

shizhMSFT:
After discussing with @sajayantony, this proposal is for `v1.1.0`. The patch release `v1.0.1` for `v1.0.0` remains the previous behavior.

qweeah:
We agreed to exclude `--skip-delete-referrers` in https://github.com/oras-project/oras/issues/1038, so originally the flag won't be added to `v1.0.1`, right?

FeynmanZhou:
[...]

PR #1402, opened 2024-06-11
chore: Update goreleaser to v6 and configuration to v2

TerryHowe:

Config changes to update `goreleaser`

### DIFF
--- a/.github/workflows/release-github.yml
+++ b/.github/workflows/release-github.yml
@@ -31,9 +31,9 @@ jobs:
         with:
           go-version: '1.22.3'
       - name: run goreleaser
-        uses: goreleaser/goreleaser-action@v5
+        uses: goreleaser/goreleaser-action@v6
         with:
-          version: latest
-          args: release --rm-dist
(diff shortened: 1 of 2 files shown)


qweeah:
Looks like this PR and #1041 are depending each other. It's better to close #1402 and include all changes in this PR.

E.g. Error try releasing without #1042: https://github.com/qweeah/oras/actions/runs/9459022967/job/26055464835

.github/workflows/release-github.yml
```diff
@@ -34,6 +34,6 @@ jobs:
         uses: goreleaser/goreleaser-action@v5
         with:
           version: latest
```
qweeah, line 36: Better change this to:
```suggestion
          version: '~> v2'
```

To avoid warning like:
```
Warning: You are using 'latest' as default version. Will lock to '~> v2'.
```

TerryHowe: Done

qweeah (approved):
LGTM

PR #1406, opened 2024-06-14
refactor: Move verbose into printer object

TerryHowe:

The verbose option should be owned by the printer.

### DIFF
--- a/cmd/oras/root/cp_test.go
+++ b/cmd/oras/root/cp_test.go
@@ -132,7 +132,7 @@ func Test_doCopy(t *testing.T) {
 	opts.From.Reference = memDesc.Digest.String()
 	dst := memory.New()
 	builder := &strings.Builder{}
-	printer := output.NewPrinter(builder)
+	printer := output.NewPrinter(builder, opts.Verbose)
 	// test
 	_, err = doCopy(context.Background(), printer, memStore, dst, &opts)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/oras/root/cp_test.go
```diff
@@ -132,7 +132,7 @@ func Test_doCopy(t *testing.T) {
 	opts.From.Reference = memDesc.Digest.String()
 	dst := memory.New()
 	builder := &strings.Builder{}
-	printer := output.NewPrinter(builder)
+	printer := output.NewPrinter(builder, false)
```
qweeah, line 135: Although it's not used, it's better to set the value to the command option.
```suggestion
	printer := output.NewPrinter(builder, opts.Verbose)
```

cmd/oras/root/cp_test.go
```diff
@@ -156,7 +156,7 @@ func Test_doCopy_skipped(t *testing.T) {
 	opts.Verbose = true
 	opts.From.Reference = memDesc.Digest.String()
 	builder := &strings.Builder{}
-	printer := output.NewPrinter(builder)
+	printer := output.NewPrinter(builder, false)
```
qweeah, line 159: printer := output.NewPrinter(builder, opts.Verbose)

cmd/oras/root/cp_test.go
```diff
@@ -191,7 +191,7 @@ func Test_doCopy_mounted(t *testing.T) {
 	}
 	to.PlainHTTP = true
 	builder := &strings.Builder{}
-	printer := output.NewPrinter(builder)
+	printer := output.NewPrinter(builder, false)
```
qweeah, line 194: printer := output.NewPrinter(builder, opts.Verbose)

qweeah (approved):
LGTM with minor suggestions.

cmd/oras/root/cp_test.go
TerryHowe: Yes, done.

PR #1407, opened 2024-06-14
refactor: Clean up pull handlers

TerryHowe:
**What this PR does / why we need it**:

Just some clean up in the pull handlers.

PR #1409, opened 2024-06-17
refactor: clean status prompts

qweeah:
As a follow up to #1407, this PR aligns the prompts naming for status handlers.


TerryHowe (approved):
/lgtm

PR #1411, opened 2024-06-17
chore: Increase print test coverage

TerryHowe:

Add more print test coverage

### DIFF
--- a/cmd/oras/internal/output/print_test.go
+++ b/cmd/oras/internal/output/print_test.go
@@ -43,23 +43,49 @@ func (mw *mockWriter) String() string {
 func TestPrint_Error(t *testing.T) {
 	mockWriter := &mockWriter{}
 	printer := NewPrinter(mockWriter, false)
-	printer.Println("boom")
+	err := printer.Println("boom")
 	if mockWriter.errorCount != 1 {
 		t.Error("Expected one errors actual <" + strconv.Itoa(mockWriter.errorCount) + ">")
+	}
+	if err != nil {
+		t.Error("Expected error to be ignored")
 	}
 }
 
 func TestPrint_NoError(t *testing.T) {
-	mockWriter := &mockWriter{}
-	printer := NewPrinter(mockWriter, false)
+	builder := &strings.Builder{}
+	printer := NewPrinter(builder, false)
 
-	expected := "blah blah"
-	printer.Println(expected)
-	actual := strings.TrimSpace(mockWriter.String())
+	expected := "normal\n"
+	err := printer.Println("normal")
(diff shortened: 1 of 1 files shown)


cmd/oras/internal/output/print_test.go
```diff
@@ -43,23 +43,49 @@ func (mw *mockWriter) String() string {
...
+	}
+	actual := builder.String()
 	if expected != actual {
 		t.Error("Expected <" + expected + "> not equal to actual <" + actual + ">")
 	}
-	if mockWriter.errorCount != 0 {
-		t.Error("Expected no errors actual <" + strconv.Itoa(mockWriter.errorCount) + ">")
+}
+
+func TestPrint_Verbose(t *testing.T) {
```
qweeah, line 74:
```suggestion
func TestPrinter_PrintVerbose(t *testing.T) {
```

cmd/oras/internal/output/print_test.go
```diff
@@ -43,23 +43,49 @@ func (mw *mockWriter) String() string {
...
+	err := printer.Println("boom")
 	if mockWriter.errorCount != 1 {
 		t.Error("Expected one errors actual <" + strconv.Itoa(mockWriter.errorCount) + ">")
 	}
+	if err != nil {
+		t.Error("Expected error to be ignored")
+	}
 }
 
 func TestPrint_NoError(t *testing.T) {
```
qweeah, line 55: According to the [testing doc](https://pkg.go.dev/testing#hdr-Examples), this function should be named as:
```suggestion
func TestPrinter_PrintVerbose_noError(t *testing.T) {
```

cmd/oras/internal/output/print_test.go
```diff
@@ -43,23 +43,49 @@ func (mw *mockWriter) String() string {
 func TestPrint_Error(t *testing.T) {
```
qweeah, line 43:
```suggestion
func TestPrinter_Println(t *testing.T) {
```

qweeah (changes requested):
Need to rename test function to be compliant with [suggested practices ](https://pkg.go.dev/testing#hdr-Examples)in Golang docs:


E.g. the naming convention to declare examples for the package, a function F, a type T and method M on type T are:
```
func Example() { ... }
func ExampleF() { ... }
func ExampleT() { ... }
func ExampleT_M() { ... }
```

Multiple example functions for a package/type/function/method may be provided by appending a distinct suffix to the name. **The suffix must start with a lower-case letter**.

qweeah (approved):
LGTM

PR #1412, opened 2024-06-17
Add e2e test go work sum file

TerryHowe:

Add this file or ignore it?


qweeah:
I would vote to ignore it, since the module sum of E2E test workspace should be tracked by `test/e2e/go.sum`

sajayantony:
Is there precedence in how other projects do this. I'm ok ignoring it but if this has some odd side effect of keeping or bringing unknown packages might move towards checkin this in.

qweeah:
> Is there precedence in how other projects do this

I do see [kubernetes](https://github.com/kubernetes/kubernetes) checks in the work sum file. But the usage is different. k8s' workspace is created on the root module level but ORAS' workspace is created on e2e test module level. **So if we check go work sum file in, we are doing it for ORAS E2E test module, not for ORAS CLI module.**

Besides from testing packages (Gomega and Ginkgo), ORAS E2E test module should rely same version of packages used by ORAS CLI module. So go work sum file won't be needed for cross-module package version control.

qweeah:
IMHO, the only advantage we don't ignore go work sum file is: we can clearly see a package version mismatch between CLI and E2E test module.

PR #1413, opened 2024-06-17
refactor: Move descriptor logic to descriptor

TerryHowe:

Move the logic of descriptor title vs media type to the descriptor package.

### DIFF
--- a/internal/descriptor/descriptor.go
+++ b/internal/descriptor/descriptor.go
@@ -37,3 +37,12 @@ func ShortDigest(desc ocispec.Descriptor) (digestString string) {
 	}
 	return digestString
 }
+
+// GetTitleOrMediaType get a descriptor name using either title or media type
+func GetTitleOrMediaType(desc ocispec.Descriptor) (name string, isTitle bool) {
+	name, ok := desc.Annotations[ocispec.AnnotationTitle]
+	if !ok {
+		return desc.MediaType, false
+	}
+	return name, true
+}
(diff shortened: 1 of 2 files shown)


internal/descriptor/descriptor.go
```diff
@@ -37,3 +37,11 @@ func ShortDigest(desc ocispec.Descriptor) (digestString string) {
 	}
 	return digestString
 }
+
+func GetTileOrMediaType(desc ocispec.Descriptor) (name string, isTitle bool) {
```
qweeah, line 41: Also requires since it's public function

```suggestion
func GetTitleOrMediaType(desc ocispec.Descriptor) (name string, isTitle bool) {
```

qweeah (changes requested):

internal/descriptor/descriptor.go
TerryHowe: Done

internal/descriptor/descriptor.go
```diff
@@ -37,3 +37,12 @@ func ShortDigest(desc ocispec.Descriptor) (digestString string) {
 	}
 	return digestString
 }
+
+// GetTitleOrMediaType get a descriptor name using either title or media type
```
qweeah, line 41:
```suggestion
// GetTitleOrMediaType gets a descriptor name using either title or media type.
```

qweeah (approved):
LGTM with minor suggestion.

PR #1415, opened 2024-06-18
refactor: Move generate content key method to descriptor

TerryHowe:

Move `GeneerateContentKey` to descriptor package and increase test coverage.

### DIFF
--- a/internal/descriptor/descriptor.go
+++ b/internal/descriptor/descriptor.go
@@ -46,3 +46,8 @@ func GetTitleOrMediaType(desc ocispec.Descriptor) (name string, isTitle bool) {
 	}
 	return name, true
 }
+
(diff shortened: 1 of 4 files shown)


internal/descriptor/descriptor.go
```diff
@@ -46,3 +46,8 @@ func GetTitleOrMediaType(desc ocispec.Descriptor) (name string, isTitle bool) {
 	}
 	return name, true
 }
+
+// GenerateContentKey generates a unique key for each content descriptor using digest and name.
```
qweeah, line 50: NIT: better keep comment length less than 80

```suggestion
// GenerateContentKey generates a unique key for each content descriptor using
// digest and name.
```

qweeah (approved):
LGTM with minor suggestions.

PR #1417, opened 2024-06-19
refactor: Move two output functions to the output package

TerryHowe:

Move the json and template output methods to the output package.


qweeah (approved):
LGTM.

PR #1420, opened 2024-06-20
refactor: Pass printer to text handler

TerryHowe:

Pass printer object to text handler.
* Fewer arguments to the handlers
* Handlers use the same mutex used elsewhere

I'm going to explore the possibility of passing the printer to the json and template handler, but trying to keep this small.


qweeah:
> I'm going to explore the possibility of passing the printer to the json and template handler, but trying to keep this small.

Making printer implement `io.Writer` should help.

qweeah (approved):
LGTM, still need other reviewer approval on the printer's implementation of `io.Writer`

PR #1426, opened 2024-06-21
refactor: Use printer object for output

TerryHowe:

Change the last calls to fmt.Fprint to use the printer object.


qweeah (approved):
LGTM, I think we can add a dedicated option to reduce duplicated  `output.NewPrinter(cmd.OutOrStdout(), opts.Verbose)` though

TerryHowe:
> I think we can add a dedicated option to reduce duplicated `output.NewPrinter(cmd.OutOrStdout(), opts.Verbose)` though

I should probably step back and look at where this has gone. Generally, I'm pretty happy, but I'm sure there is room for improvement somewhere.

PR #1427, opened 2024-06-22
refactor: Handle stderr similar to stdout

TerryHowe:

Use the `cmd.ErrOrStderr()` for error output. This is more consistent and go scripting will be able to capture stderr.

This should be the final PR for the printer although it may be worth looking at where things are and to see what could be improved.

### DIFF
--- a/cmd/oras/internal/output/print.go
+++ b/cmd/oras/internal/output/print.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
 	"sync"
 
 	"oras.land/oras/internal/descriptor"
@@ -34,13 +33,14 @@ type PrintFunc func(ocispec.Descriptor) error
 // Printer prints for status handlers.
 type Printer struct {
 	out     io.Writer
+	err     io.Writer
 	verbose bool
 	lock    sync.Mutex
 }
 
 // NewPrinter creates a new Printer.
-func NewPrinter(out io.Writer, verbose bool) *Printer {
-	return &Printer{out: out, verbose: verbose}
+func NewPrinter(out io.Writer, err io.Writer, verbose bool) *Printer {
+	return &Printer{out: out, err: err, verbose: verbose}
 }
 
 // Write implements the io.Writer interface.
@@ -57,7 +57,7 @@ func (p *Printer) Println(a ...any) error {
 	_, err := fmt.Fprintln(p.out, a...)
 	if err != nil {
 		err = fmt.Errorf("display output error: %w", err)
-		_, _ = fmt.Fprint(os.Stderr, err)
+		_, _ = fmt.Fprint(p.err, err)
 	}
 	// Errors are handled above, so return nil
 	return nil
@@ -70,7 +70,7 @@ func (p *Printer) Printf(format string, a ...any) error {
 	_, err := fmt.Fprintf(p.out, format, a...)
 	if err != nil {
 		err = fmt.Errorf("display output error: %w", err)
-		_, _ = fmt.Fprint(os.Stderr, err)
+		_, _ = fmt.Fprint(p.err, err)
 	}
 	// Errors are handled above, so return nil
 	return nil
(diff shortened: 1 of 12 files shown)


cmd/oras/internal/output/print.go
```diff
@@ -57,7 +57,7 @@ func (p *Printer) Println(a ...any) error {
 	_, err := fmt.Fprintln(p.out, a...)
 	if err != nil {
 		err = fmt.Errorf("display output error: %w", err)
-		_, _ = fmt.Fprint(os.Stderr, err)
+		_, _ = fmt.Fprint(p.err, err)
```
qweeah, line 60: What if a user redirects stderr into a malformed output? In that case they won't be able to see the error message anyway.

IMO here we should always output to os.Stderr to make sure the `err` is delivered in a best-effort way.

TerryHowe: If we don't pass `cmd.ErrOrStderr()` to the printer, the user cannot capture it for go scripting. That is the point of this.

The user can always mess things up, redirect stderr to /dev/null or whatever they want to do.

qweeah: Okay there are two kinds of redirection. Well maybe I am not making it clear, by redirection I mean: during go scripting, changing the command stderr via https://pkg.go.dev/github.com/spf13/cobra@v1.8.1#Command.SetErr.

qweeah: In that case should we sent `err` here to os.Stderr or the user specified error writer? The former makes more sense to me.

qweeah: The key point here is: if a user runs a binary that uses ORAS go scripting, should the `err` here be consider as:

1) ORAS' error output? (Should sent it to `cmd.ErrOrStderr()`), or
2) the binary's error output?  (Should sent it to `os.Stdout`)

qweeah:
It just came to me: If user resets command output (via [SetErr](https://pkg.go.dev/github.com/spf13/cobra@v1.8.1#Command.SetErr.)) should we also set the [err output stream for logging](https://github.com/oras-project/oras/blob/main/internal/trace/context.go#L40)?

TerryHowe:
I created and issue to investigate the logging idea. I'm not sure.

As to go scripting, I think it should work just like shell scripting. By default stdout and stderr go to the current file descriptors. If the user wants to redirect stdout or stderr, they should be allowed to do that. Currently, they can only redirect stdout.

qweeah:
> By default stdout and stderr go to the current file descriptors. 

One tricky point is that go scripting allows running several commands in parallel, and assign each run different files for stdout and stderr. Logging should go to the user-assigned stderr is need (but not need to be done in this PR).

cmd/oras/internal/output/print.go
qweeah: I am okay with sending stdout output failure to user-specified command stderr, but that may change in the future. Is there any way we can declare that ORAS go scripting(using ORAS CLI as golang SDK) is experimental and subject to changes?

TerryHowe:
Created issue https://github.com/oras-project/oras-www/issues/357

cmd/oras/internal/output/print.go
qweeah: We can follow it up in https://github.com/oras-project/oras/issues/1432 and https://github.com/oras-project/oras-www/issues/357

cmd/oras/root/discover.go
```diff
@@ -85,7 +85,7 @@ Example - Discover referrers of the manifest tagged 'v1' in an OCI image layout
 			if cmd.Flags().Changed("output") {
 				switch opts.Format.Type {
 				case "tree", "json", "table":
-					fmt.Fprintf(cmd.ErrOrStderr(), "[DEPRECATED] --output is deprecated, try `--format %s` instead\n", opts.Template)
```
qweeah, line 88: Should use `cmd.PrintErrf("[DEPRECATED] --output is deprecated, try `--format %s` instead\n", opts.Template)` instead?

qweeah (approved):
LGTM though some `printer` are already imbedded in command options and no need to recreate. Will change that in another PR.

PR #1428, opened 2024-06-23
refactor: assembles printer in the common output option

qweeah:
Follow up of #1426, de-duplicates `printer` initialization during command run(mentioned in https://github.com/oras-project/oras/pull/1426#pullrequestreview-2133693570)


TerryHowe (approved):
/lgtm

Issue #1404, opened 2024-06-13
When subject artifact reference is not specified, error output of `oras attach` is confusing


qweeah:
### What happened in your environment?

Running below command with empty variable `$BASE_IMAGE` (and thus actually not specified):
```console
> oras attach --artifact-type "application/vnd.openvex.vex+json" $BASE_IMAGE ./vex/python3.12.2.vex.json
Error: neither file nor annotation provided in the command
Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<layer_media_type>] [...]
To attach to an existing artifact, please provide files via argument or annotations via flag "--annotation". Run "oras attach -h" for more options and examples
```

The error message still "[makes me think](https://a.co/d/0SBdPYc)" because I specified a file and I don't want to add annotations. 

### What did you expect to happen?

The output error message should tell that the problem is with the image (or subject artifact).

### How can we reproduce it?

Run `oras attach` without specifying any subject artifact reference.

### What is the version of your ORAS CLI?

Version:        1.2.0
Go version:     go1.22.3
Git commit:     dcef719e208a9b226b15bc6512ad729a7dd93270
Git tree state: clean

### What is your OS environment?

20.04

### Are you willing to submit PRs to fix it?

- [ ] Yes, I am willing to fix it.

qweeah:
Technically, a valid artifact reference can also be a valid file path. When a user only provides one argument, it's hard to which of the below usages matches the real case:
- The subject reference is missing
- The to-be-attach file path is missing

Current error message handles the latter but the real use case is the former.

qweeah:
IMHO, it's better to follow the practice suggested by [CLI Guideline](https://clig.dev/#arguments-and-flags) and make file reference specified by flags. E.g. The usage of `oras attach` should changed to
```bash
oras attach --artifact-type "application/vnd.openvex.vex+json" $BASE_IMAGE --file ./vex/python3.12.2.vex.json
```

Since it's breaking change, specifying file paths via argument is still allowed but will trigger deprecation warning.

PR #1430, opened 2024-06-24
chore(ux): improve error message when attaching without subject artifact 

### LINKED ISSUE #1424: improve error message of `oras attach` when subject artifact is not specified
When using `oras attach`, if a user forgot to provide the subject artifact reference and only specifies one to-be-uploaded file, ORAS CLI will output (tested in bash, different shell might have different output):

```console
> ARTIFACT=
> oras attach --artifact-type test/example $ARTIFACT./test.json
Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<layer_media_type>] [...]
To attach to an existing artifact, please provide files via argument or annotations via flag "--annotation". Run "oras attach -h" for more options and examplesdigest]
```

Input was parsed under a wrong consumption: user does specify subject artifact reference `./layers`, which actually is a file reference to be attached. This error can be avoided by parsing the only input argument. If the reference is valid, continue with current output. If it's not, give corresponding hint like:

```console
Error: "./test.json" is an invalid artifact reference
Did you forget to specify a subject artifact?
```

qweeah:
This PR improves error message when the `oras attach` is run with only one argument, which might be an invalid artifact subject reference or mistakenly used as a to-be-uploaded file path. 

**Before**
```console
> oras attach --artifact-type test/example $ARTIFACT ./layers 
Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<layer_media_type>] [...]
To attach to an existing artifact, please provide files via argument or annotations via flag "--annotation". Run "oras attach -h" for more options and examples
```
**After**
```console
> oras attach --artifact-type test/example $ARTIFACT ./layers 
Error: "./layers": no tag or digest specified
Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<layer_media_type>] [...]
Have you specified an artifact reference to attach to? Please specify a reference in the form of "<name>:<tag>" or "<name>@<digest>". Run "oras attach -h" for more options and examples
```

Fixes #1424, #1404


wangxiaoxuan273:
LGTM but I'm not a maintainer

cmd/oras/internal/errors/errors.go
```diff
@@ -26,6 +26,15 @@ import (
 	"oras.land/oras-go/v2/registry/remote/errcode"
 )
 
+// OperationType reperesnts the type of operation.
+type OperationType string
+
+const (
+	// ParseArtifactReference is the operation type parsing artifact
+	// reference.
+	ParseArtifactReference = OperationType("")
```
TerryHowe, line 35: Set this to `ParseArtifactReference` maybe?

qweeah: Sorry I don't get it.

qweeah: Do you mean change the string value to `"ParseArtifactReference "`

TerryHowe: Yes or maybe use iota instead?

TerryHowe (approved):
/lgtm

cmd/oras/root/attach.go
```diff
@@ -87,10 +87,19 @@ Example - Attach file to the manifest tagged 'v1' in an OCI image layout folder
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			opts.RawReference = args[0]
 			opts.FileRefs = args[1:]
-			if err := option.Parse(cmd, &opts); err != nil {
-				return err
+			err := option.Parse(cmd, &opts)
+			if err == nil && opts.Reference == "" {
+				// ensure reference is not empty
+				err = oerrors.NewErrEmptyTagOrDigest(opts.RawReference, cmd, true)
 			}
```
shizhMSFT, line 94:
```suggestion
			if err == nil {
				// ensure reference is not empty
				if opts.Reference != "" {
					return nil
				}
				err = oerrors.NewErrEmptyTagOrDigest(opts.RawReference, cmd, true)
			}
```

cmd/oras/root/attach.go
```diff
@@ -87,10 +87,19 @@ Example - Attach file to the manifest tagged 'v1' in an OCI image layout folder
...
-			return nil
+			if len(opts.FileRefs) == 0 {
+				// no file argument provided
+				if err, ok := err.(*oerrors.Error); ok && err.OperationType == oerrors.ParseArtifactReference {
+					// invalid reference
+					err.Recommendation = fmt.Sprintf("Have you specified an artifact reference to attach to? %s", err.Recommendation)
```
shizhMSFT, line 99: Consider compiler style recommendations: "Are you missing an artifact reference to attach to"?

Reference:
- https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/assembly-references
- https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1061

shizhMSFT (changes requested):

cmd/oras/root/attach.go
FeynmanZhou: For this case, how about improving the error message as follows: 

```
> oras attach --artifact-type test/example $ARTIFACT ./layers 
Error: "./layers" is an invalid artifact reference.
Usage: oras attach [flags] --artifact-type=<type> <name>{:<tag>|@<digest>} <file>[:<layer_media_type>] | --annotation "key=val"
To attach files to an object, please specify a valid artifact reference and files via arguments in sequence". Run "oras attach -h" for more options and examples.
```

FeynmanZhou (commented):
Just one suggested error message for this case

cmd/oras/root/attach.go
qweeah: @FeynmanZhou What if a user is trying to attach end-of-life annotations and input an invalid artifact reference? E.g. forget to include the subject's tag. Will the error message `please specify a valid artifact reference and files via arguments in sequence` makes he/she wondering what are the `files` for?

cmd/oras/root/attach.go
```diff
@@ -87,10 +87,22 @@ Example - Attach file to the manifest tagged 'v1' in an OCI image layout folder
...
 			opts.FileRefs = args[1:]
-			if err := option.Parse(cmd, &opts); err != nil {
-				return err
+			err := option.Parse(cmd, &opts)
+			if err == nil {
+				if opts.Reference != "" {
+					return nil
+				}
+				// ensure reference is not empty
+				err = oerrors.NewErrEmptyTagOrDigest(opts.RawReference, cmd, true)
```
shizhMSFT, line 96: nit: I just noticed this block of code is `EnsureReferenceNotEmpty` that you have removed. Consider
```suggestion
				err = opts.EnsureReferenceNotEmpty(cmd, true)
				if err == nil {
					return nil
				}
```
The method name is self-explainable.

shizhMSFT (approved):
LGTM with suggestions

FeynmanZhou (approved):
LGTM

PR #1433, opened 2024-06-26
refactor: reduce printer instantiation & unify stdout/stderr usage

qweeah:

Follow-up of #1427.


TerryHowe (approved):
/lgtm

PR #1438, opened 2024-07-01
chore: Remove last deprecated method

TerryHowe:
**What this PR does / why we need it**:

Remove the last deprecated method

PR #1439, opened 2024-07-03
refactor: add display handler for `cp`, `tag` and `manifest push`

qweeah:
This PR unifies display handler usage by adding dedicated handlers to relevant commands, as a follow-up of #1438.


TerryHowe (approved):
/lgtm

PR #1447, opened 2024-07-17
refactor: Isolate progress channel in Messenger

TerryHowe:

Isolate details about the progress channel from users

### DIFF
--- a/cmd/oras/internal/display/status/progress/messenger.go
+++ b/cmd/oras/internal/display/status/progress/messenger.go
@@ -1,0 +1,95 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package progress
+
+import (
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras/cmd/oras/internal/display/status/progress/humanize"
+	"time"
+)
+
+// Messenger is progress message channel.
+type Messenger struct {
+	ch     chan *status
+	closed bool
+}
+
+// Start initializes the messenger.
+func (sm *Messenger) Start() {
+	if sm.ch == nil {
+		return
+	}
+	sm.ch <- startTiming()
+}
+
+// Send a status message for the specified descriptor.
+func (sm *Messenger) Send(prompt string, descriptor ocispec.Descriptor, offset int64) {
+	for {
+		select {
+		case sm.ch <- newStatusMessage(prompt, descriptor, offset):
+			return
+		case <-sm.ch:
+			// purge the channel until successfully pushed
+		default:
+			// ch is nil
+			return
+		}
+	}
+}
+
+// Stop the messenger after sending a end message.
+func (sm *Messenger) Stop() {
+	if sm.closed {
+		return
+	}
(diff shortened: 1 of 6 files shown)


cmd/oras/internal/display/status/progress/messenger.go
```diff
@@ -0,0 +1,55 @@
...
+limitations under the License.
+*/
+
+package progress
+
+import (
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Messenger is progress message channel
```
qweeah, line 22:
```suggestion
// Messenger is progress message channel.
```

cmd/oras/internal/display/status/progress/status.go
```diff
@@ -65,25 +65,25 @@ func newStatus() *status {
...
 	}
 }
 
-// EndTiming ends timing and set status to done.
-func EndTiming() *status {
+// endTiming ends timing and set status to done.
+func endTiming() *status {
 	return &status{
 		offset:  -1,
 		endTime: time.Now(),
```
qweeah, line 89: Other than making those private, I think we can get rid of those functional. I wrote those code and it sucks :)

cmd/oras/internal/display/status/progress/messenger.go
```diff
@@ -0,0 +1,55 @@
...
+	for {
+		if sm.ch == nil {
+			return
+		}
+		select {
+		case sm.ch <- newStatusMessage(prompt, descriptor, offset):
+			return
+		case <-sm.ch:
+			// purge the channel until successfully pushed
+		}
```
qweeah, line 44: `nil` can be ignored via:
```suggestion
		select {
		case sm.ch <- newStatusMessage(prompt, descriptor, offset):
			return
		case <-sm.ch:
			// purge the channel until successfully pushed
		}
		default:
			// ch is nil
			return
```

cmd/oras/internal/display/status/progress/messenger.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+func (sm *Messenger) Stop() {
+	if sm.ch == nil {
+		return
+	}
+	sm.ch <- endTiming()
+	close(sm.ch)
+	sm.ch = nil
```
qweeah, line 54: I would suggest to leave sm.ch there. Sending message to a closed status channel is a code-level defect and it's easier to be identified from panic(send to closed channel) to sleep(send to nil channel). Only when all goroutines are sleeping, there will be a panic.

qweeah (changes requested):

cmd/oras/internal/display/status/progress/status.go
qweeah: I mean move function content to `messenger.go`

cmd/oras/internal/display/status/progress/messenger.go
qweeah, line 1: Don't forget to add comment for public functions in this file.

qweeah:
BTW @TerryHowe Good job, this is an elegant refactoring!

cmd/oras/internal/display/status/progress/status.go
TerryHowe: I'm going to move them to `messenger.go`. They don't do much, but I think the code reads better with them because of the select/case statement.

cmd/oras/internal/display/status/progress/messenger.go
TerryHowe: Done.

cmd/oras/internal/display/status/progress/messenger.go
TerryHowe: Done.

cmd/oras/internal/display/status/progress/messenger.go
TerryHowe: Done

cmd/oras/internal/display/status/progress/status.go
TerryHowe: Moved

qweeah (approved):
LGTM

PR #1448, opened 2024-07-18
fix: variable name conflicts with built in function

TerryHowe:

The variable name `len` conflicts with the built in. This will simplify my other PR a bit as well.

### DIFF
--- a/cmd/oras/internal/display/status/progress/manager.go
+++ b/cmd/oras/internal/display/status/progress/manager.go
@@ -92,17 +92,17 @@ func (m *manager) render() {
 	defer m.statusLock.RUnlock()
 	// todo: update size in another routine
 	width, height := m.console.Size()
-	len := len(m.status) * 2
+	statusLength := len(m.status) * 2
 	offset := 0
-	if len > height {
+	if statusLength > height {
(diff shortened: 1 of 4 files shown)


cmd/oras/internal/display/status/progress/manager.go
```diff
@@ -92,17 +92,17 @@ func (m *manager) render() {
 	defer m.statusLock.RUnlock()
 	// todo: update size in another routine
 	width, height := m.console.Size()
-	len := len(m.status) * 2
+	statusLength := len(m.status) * 2
```
qweeah, line 95: How about `lineCount`, one status output take two lines.
```suggestion
	lineCount := len(m.status) * 2
```

qweeah (approved):
LGTM with minor suggestions.

qweeah:
Added a unit test to cover the missing line, but I have wrongly pushed to #1447. Guess this should be merged first and merging #1447 will save that fault.

cmd/oras/internal/display/status/progress/manager.go
TerryHowe: Done

PR #1451, opened 2024-07-22
chore: Bring back the print error

TerryHowe:

I'm not a big fan of the print error, but other things I am a lot more concerned with.


qweeah (approved):
LGTM

PR #1452, opened 2024-07-22
fix: A couple small fixes to the descriptor test

TerryHowe:

Minor fixes to the descriptor test.


qweeah (approved):
LGTM

PR #1453, opened 2024-07-22
refactor: Take private memory store out of test function

TerryHowe:

Take private functions out of test function so that they can be accessed by other tests. Also, changes it so  the return values are declared rather than array indexes.


qweeah (approved):
LGTM, thanks @TerryHowe

PR #1455, opened 2024-07-24
refactor: Make mock fetcher generally available

TerryHowe:

I wanted to get a feel for how people felt about making a mock fetcher as a test utility, so I created this small PR.

### DIFF
--- a/internal/testutils/fetcher.go
+++ b/internal/testutils/fetcher.go
@@ -1,0 +1,101 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testutils
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


internal/testutils/fetcher.go
```diff
@@ -0,0 +1,96 @@
...
+func (mf *MockFetcher) pushImage(subject *ocispec.Descriptor, mediaType string, config ocispec.Descriptor, layers ...ocispec.Descriptor) ocispec.Descriptor {
+	manifest := ocispec.Manifest{
+		MediaType: mediaType,
+		Subject:   subject,
+		Config:    config,
+		Layers:    layers,
+	}
+	manifestJSON, err := json.Marshal(manifest)
+	if err != nil {
+		mf.t.Fatal(err)
```
qweeah, line 74: Anyway we can exclude this file or this package? Test code should not be calculated into coverage counting.

qweeah (changes requested):
Well, I got to confess mocking a fetcher is toll and the test code is that bad, but you haven't met the mocked registry service yet :)

internal/testutils/fetcher.go
qweeah, line 1: Need comment doc for all public functions.

TerryHowe: Done.

internal/testutils/fetcher.go
TerryHowe: I'm not sure how exclude this from coverage. All the ways I'm seeing to do that are not great.

TerryHowe: Okay, that seemed to work. No coverage on this.

qweeah (approved):
LGTM

PR #1459, opened 2024-07-27
refactor: Move testutils to general location

TerryHowe:

The successor PR is big and still lacks coverage without more changes. This PR is for a small part of that change, just moving the console testutils to internal/testutils so we can have one testutils that is ignored by coverage.


internal/testutils/testutils.go
qweeah, line 1: How about renaming this file to `console.go`?

Makefile
```diff
@@ -33,7 +33,8 @@ LDFLAGS += -X $(PROJECT_PKG)/internal/version.GitTreeState=${GIT_DIRTY}
 
 .PHONY: test
 test: tidy vendor check-encoding  ## tidy and run tests
-	$(GO_EXE) test -race -v -coverprofile=coverage.txt -covermode=atomic -coverpkg=./... ./...
+	$(GO_EXE) test -race -v -coverprofile=coverage-all.txt -covermode=atomic -coverpkg=./... ./...
```
qweeah, line 36: How about modifying `-coverpkg`? Like

> $(GO_EXE) test -race -v -coverprofile=coverage-all.txt -covermode=atomic -coverpkg=$(go list ./... | grep -v internal/testutils | tr '\n' ',') ./...

qweeah (changes requested):

internal/testutils/testutils.go
TerryHowe: Done

Makefile
TerryHowe: Done.

.gitignore
```diff
@@ -36,6 +36,7 @@ debug
 
 # Custom
 coverage.txt
+coverage-all.txt
```
qweeah, line 39: No need to ignore this file

qweeah (approved):
LGTM with minor suggestion.

.gitignore
TerryHowe: Done

PR #1465, opened 2024-07-30
feature: Add manifest fetch documentation for digest

TerryHowe:

This PR adds an example for `--format` and it is a common task (for me at least) to get the digest.


sajayantony (approved):

sajayantony:
Funny I do the same thing instead of `resolve` https://oras.land/docs/commands/oras_resolve

PR #1466, opened 2024-08-01
fix: Mention the resolve command in manifest fetch help

TerryHowe:

A reminder to use the resolve command.


qweeah (approved):
LGTM, the suggested usage is not as fast as `oras resolve`

PR #1473, opened 2024-08-09
refactor: DRY up text_test

TerryHowe:

The text_test was a little wordy and I though the error fetcher would be easier to find in its own file.


qweeah (approved):
LGTM

Issue #1461, opened 2024-07-28
Create a mock console to increase test coverage on mac


TerryHowe:
### What is the version of your ORAS CLI

1.2.0

### What would you like to be added?

Increase test coverage on mac with mock console

### Why is this needed for ORAS?

These tests do not run on mac because of X-code issue:
```
 % grep -r solaris cmd internal 
cmd/oras/internal/option/common_unix_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/internal/display/status/track/target_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/internal/display/status/progress/manager_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/internal/display/status/progress/status_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/internal/display/status/tty_console_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/internal/display/status/console/console_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/internal/display/status/console/testutils/testutils.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/root/cp_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/root/blob/push_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
cmd/oras/root/blob/fetch_test.go://go:build freebsd || linux || netbsd || openbsd || solaris
```

### Are you willing to submit PRs to contribute to this feature?

- [X] Yes, I am willing to implement it.

qweeah:
1) Where should be mac test be run?
2) How to merge the test coverage of mac with linux's?
3) What SDK should be used to mock the console test on mac?

TerryHowe:
1. I'm hoping that everything outside of cmd/oras/internal/display/status/progress would be tested on mac
2. I was hoping that a layer would be created at `progress` package that could be mocked with dependency injection
3. I was not planning on bringing in a package for mock, just go code.

TerryHowe:
Close to done. Related:

https://github.com/oras-project/oras/pull/1481
https://github.com/oras-project/oras/pull/1474
https://github.com/oras-project/oras/pull/1511

PR #1474, opened 2024-08-11
refactor: simplify track reader manager and messenger use

TerryHowe:

If we just pass the messenger to reader, it will be a lot easier to test these packages.

Partial: https://github.com/oras-project/oras/issues/1461


TerryHowe:
I'm going to make a couple PRs out of this until it is focused only one the reader/messenger refactor again.

TerryHowe:
I think I still want to pull something out of this.

TerryHowe:
Partially handled by https://github.com/oras-project/oras/pull/1624 and https://github.com/oras-project/oras/pull/1524

Issue #1053, opened 2023-08-08
Support multi-arch image/artifact push


sajayantony:
### What is the version of your ORAS CLI

1.0.0

### What would you like to be added?

As artifacts become increasingly accepted, it would be useful for ORAS to support multi-arch artifacts as well. 
There are already some standard conventions of tools like homebrew implementing this.  

Multi-arch manifest below - https://explore.ggcr.dev/?image=ghcr.io%2Fhomebrew%2Fcore%2Foras:1.0.1

```json 
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d0e42e6f6d564a9d1e178e72b42ff12b7e2e3c36cd32b44b150f7465c171db41",
      "size": 1914,
      "platform": {
        "architecture": "arm64",
        "os": "darwin",
        "os.version": "macOS 12.6"
      },
      "annotations": {
        "org.opencontainers.image.ref.name": "1.0.1.arm64_monterey",
        "sh.brew.bottle.digest": "8690763d0a9d0be6574ee4cfdb8edbab302f232c8056bae2d0083f170719afed",
        "sh.brew.bottle.size": "3436422",
        "sh.brew.tab": "{\"homebrew_version\":\"4.1.4-6-ga8022e6\",\"changed_files\":[],\"source_modified_time\":1691075089,\"compiler\":\"clang\",\"runtime_dependencies\":[],\"arch\":\"arm64\",\"built_on\":{\"os\":\"Macintosh\",\"os_version\":\"macOS 12.6\",\"cpu_family\":\"dunno\",\"xcode\":\"14.2\",\"clt\":\"14.2.0.0.1.1668646533\",\"preferred_perl\":\"5.30\"}}"
      }
    },
```

Each binary is uploaded with the following manifest 

```json 
 oras manifest get --pretty ghcr.io/homebrew/core/oras@sha256:d0e42e6f6d564a9d1e178e72b42ff12b7e2e3c36cd32b44b150f7465c171db41
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:0e7e738a321110bc395c82868cd93c6dbd0e4afa7b8931f08a6039ed59544249",
    "size": 222
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
[...]

PR #1475, opened 2024-08-12
feat: create index from existing manifests (index create)

wangxiaoxuan273:
`oras manifest index create` command

Current output:
```
~# oras manifest index create wxxdevreg.azurecr.io/test:tag0,tag1,tag2 amd64 armv7
Fetching  amd64
Fetched   amd64
Fetching  armv7
Fetched   armv7
Packed    5123c2f12c03  application/vnd.oci.image.index.v1+json
Pushed    [registry] wxxdevreg.azurecr.io/test:tag0
Tagged    tag2
Tagged    tag1
Digest: sha256:5123c2f12c03cf1af0fc7b27dde0e5c363fa61ecff7885a0fe6c09d1af4e039f
```

Part of #1053 


cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,181 @@
...
+			if err != nil {
+				return nil, err
+			}
+		}
+		resolved = append(resolved, desc)
+	}
+	return resolved, nil
+}
+
+func getPlatform(ctx context.Context, target oras.ReadOnlyTarget, manifestBytes []byte) (*ocispec.Platform, error) {
```
wangxiaoxuan273, line 122: We may need to handle it if the platform is empty.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,208 @@
...
+
+Example - create an index and push it with multiple tags:
+  oras manifest index create localhost:5000/tag1, tag2, tag3 amd64 arm64 sha256:xxx
+`,
+		Args: cobra.MinimumNArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			refs := strings.Split(args[0], ",")
+			indexRef := refs[0]
+			// if a digest is given as the index reference, we need to ignore it
+			indexRef, _, _ = strings.Cut(indexRef, "@")
```
wangxiaoxuan273, line 77: If a user gives a digest for the index to create (i.e. `oras manifest index create localhost:5000/hello@sha256:xxx arm64`), should we silently ignore it, or gives an error and fails?

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func createIndex(cmd *cobra.Command, opts createOptions) error {
+	ctx, logger := command.GetLogger(cmd, &opts.Common)
+	target, err := opts.NewTarget(opts.Common, logger)
+	if err != nil {
+		return err
+	}
+	// if a digest is given as the index reference, we need to ignore it
+	opts.RawReference, _, _ = strings.Cut(opts.RawReference, "@")
```
wangxiaoxuan273, line 96: If the user gives a digest for index reference, should we silently ignore it, or return an error and fail?

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+	// tag the index if extra tags are provided
+	extraRefs = removeDigests(extraRefs)
+	if len(extraRefs) != 0 {
+		handler := display.NewManifestPushHandler(printer)
+		tagListener := listener.NewTaggedListener(target, handler.OnTagged)
+		if _, err = oras.TagBytesN(ctx, tagListener, desc.MediaType, content, extraRefs, oras.DefaultTagBytesNOptions); err != nil {
+			return err
+		}
+	}
+	return printer.Println("Digest:", desc.Digest)
```
qweeah, line 192: Looks like this should be output before tagging.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+		return ocispec.Descriptor{}, nil, err
+	}
+	desc := content.NewDescriptorFromBytes(ocispec.MediaTypeImageIndex, indexBytes)
+	return desc, indexBytes, nil
+}
+
+func pushIndex(ctx context.Context, target oras.Target, desc ocispec.Descriptor, content []byte, ref string, extraRefs []string, path string, printer *output.Printer) error {
+	// push the index
+	var err error
+	if ref == "" || contentutil.IsDigest(ref) {
```
qweeah, line 173: `contentutil.IsDigest(ref)` is a bad path and should be avoided (Image a user can predict the digest of packed index, they don't need `oras` at all)

qweeah (changes requested):

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, err
+	}
+	// fetch config content
+	contentBytes, err := content.FetchAll(ctx, target, manifest.Config)
+	if err != nil {
+		return nil, err
+	}
+	var platform ocispec.Platform
+	if err := json.Unmarshal(contentBytes, &platform); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
qweeah: E.g. I can have a manifest with config `abcdefg` but still that doesn't mean I cannot create an index based on the manifest.

shizhMSFT (commented):
Please mark `oras manifest index` command set as "experimental".

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+	sources   []string
+	extraRefs []string
+}
+
+func createCmd() *cobra.Command {
+	var opts createOptions
+	cmd := &cobra.Command{
+		Use:   "create [flags] <name>[:<tag[,<tag>][...]] [{<tag>|<digest>}...]",
+		Short: "Create and push an index from provided manifests",
+		Long: `Create and push an index to a repository or an OCI image layout
```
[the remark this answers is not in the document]
sajayantony: Shorten the text to maybe `same repository` to avoid really verbose help.

cmd/oras/root/manifest/index/create.go
sajayantony: Nested indexes are supported in OCI and should be a test case as well.

wangxiaoxuan273: added a test case for creating nested indexes.

cmd/oras/root/manifest/index/create.go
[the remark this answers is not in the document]
wangxiaoxuan273: Thanks Billy, this is good wording.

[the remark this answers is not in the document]
wangxiaoxuan273: changed to `Create and push an index from provided manifests. All manifests should be in the same repository`

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+  oras manifest index create localhost:5000/hello:latest amd64 arm64 darwin
+
+Example - create an index from source manifests using both tags and digests, 
+ and push the index with tag 'latest':
+  oras manifest index create localhost:5000/hello latest amd64 sha256:xxx darwin
+
+Example - create an index and push it with multiple tags:
+  oras manifest index create localhost:5000/tag1, tag2, tag3 amd64 arm64 sha256:xxx
+`,
+		Args: cobra.MinimumNArgs(1),
```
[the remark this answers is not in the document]
wangxiaoxuan273: changed.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+Example - create an index from source manifests tagged amd64, arm64, darwin in the repository
+ localhost:5000/hello, and push the index with tag 'latest':
+  oras manifest index create localhost:5000/hello:latest amd64 arm64 darwin
+
+Example - create an index from source manifests using both tags and digests, 
+ and push the index with tag 'latest':
+  oras manifest index create localhost:5000/hello latest amd64 sha256:xxx darwin
+
+Example - create an index and push it with multiple tags:
+  oras manifest index create localhost:5000/tag1, tag2, tag3 amd64 arm64 sha256:xxx
```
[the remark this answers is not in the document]
wangxiaoxuan273: removed the space

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+func createIndex(cmd *cobra.Command, opts createOptions) error {
+	ctx, logger := command.GetLogger(cmd, &opts.Common)
+	target, err := opts.NewTarget(opts.Common, logger)
+	if err != nil {
+		return err
+	}
+	// if a digest is given as the index reference, we need to ignore it
+	opts.RawReference, _, _ = strings.Cut(opts.RawReference, "@")
+	// we assume that the sources and the to be created index are all in the same
+	// repository, so no copy is needed
```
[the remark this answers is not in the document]
wangxiaoxuan273: removed this comment

cmd/oras/root/manifest/index/create.go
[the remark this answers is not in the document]
wangxiaoxuan273: good point

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: Don't add to the resulted descriptor if the platform information is empty.

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: If user gives an invalid tag, we should return an error.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+
+func createCmd() *cobra.Command {
+	var opts createOptions
+	cmd := &cobra.Command{
+		Use:   "create [flags] <name>[:<tag[,<tag>][...]] [{<tag>|<digest>}...]",
+		Short: "Create and push an index from provided manifests",
+		Long: `Create and push an index to a repository or an OCI image layout
+
+Example - create an index from source manifests tagged amd64, arm64, darwin in the repository
+ localhost:5000/hello, and push the index without tagging it:
```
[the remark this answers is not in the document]
wangxiaoxuan273: change the tags to `linux-amd64`, `linux-arm64`

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+	return pushIndex(ctx, target, desc, content, opts.Reference, opts.extraRefs, opts.AnnotatedReference(), opts.Printer)
+}
+
+func resolveSourceManifests(ctx context.Context, target oras.ReadOnlyTarget, opts createOptions) ([]ocispec.Descriptor, error) {
+	var resolved []ocispec.Descriptor
+	for _, source := range opts.sources {
+		desc, content, err := oras.FetchBytes(ctx, target, source, oras.DefaultFetchBytesOptions)
+		if err != nil {
+			if errors.Is(err, errdef.ErrNotFound) {
+				return nil, fmt.Errorf("could not resolve %s. Make sure the manifest exists in %s", source, opts.Path)
```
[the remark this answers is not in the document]
wangxiaoxuan273: change the error message to `cannot find the manifest %s`, source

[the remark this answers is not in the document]
wangxiaoxuan273: should not separate ErrNotFound from other error handlings

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			return nil, err
+		}
+		opts.Println("Resolved manifest", source)
+		if descriptor.IsImageManifest(desc) {
+			desc.Platform, err = getPlatform(ctx, target, content)
+			if err != nil {
+				return nil, err
+			}
+		}
```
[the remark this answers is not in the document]
wangxiaoxuan273: should add a check to only allow manifests. Should return an error if it is a blob.

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: Has a point. If this is printed before tagging and tagging fails, user knows what's the digest. But I wrote it to align with existing status output of `oras manifest push`.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := json.Unmarshal(contentBytes, &platform); err != nil {
+		return nil, nil
+	}
+	if platform.Architecture == "" && platform.OS == "" {
+		return nil, nil
+	}
+	return &platform, nil
+}
+
+func packIndex(oldIndex *ocispec.Index, manifests []ocispec.Descriptor) (ocispec.Descriptor, []byte, error) {
```
Wwwsylvia, line 161: nit: is this more like a `baseIndex`?

internal/descriptor/descriptor_test.go
```diff
@@ -56,6 +56,18 @@ func TestDescriptor_IsImageManifest(t *testing.T) {
 	}
 }
 
+func TestDescriptor_IsManifest(t *testing.T) {
+	got := descriptor.IsManifest(imageDesc)
+	if !reflect.DeepEqual(got, true) {
```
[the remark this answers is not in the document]
wangxiaoxuan273: I just copied it from other tests in the file.

internal/descriptor/descriptor.go
```diff
@@ -21,6 +21,20 @@ import (
...
 )
 
+// IsManifest checks if a descriptor describes a manifest.
+func IsManifest(desc ocispec.Descriptor) bool {
+	switch desc.MediaType {
+	case docker.MediaTypeManifest,
+		docker.MediaTypeManifestList,
+		ocispec.MediaTypeImageManifest,
+		ocispec.MediaTypeImageIndex,
+		"application/vnd.oci.artifact.manifest.v1+json":
```
qweeah, line 31: Should add docker manifest list in but remove `application/vnd.oci.artifact.manifest.v1+json`.
@Wwwsylvia What do you think? Also any chance to promote `IsManifest` and `IsImageManifest` to oras-go?

internal/descriptor/descriptor_test.go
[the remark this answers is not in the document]
qweeah: how about
```suggestion
	if got {
```

qweeah (changes requested):

internal/descriptor/descriptor.go
wangxiaoxuan273: I copied this from oras-go. It's an internal oras-go function.

cmd/oras/root/manifest/index/create.go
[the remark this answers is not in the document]
wangxiaoxuan273: It will be used in `oras manifest index update`.

cmd/oras/internal/display/metadata/text/manifest_push.go
```diff
@@ -37,3 +38,20 @@ func NewManifestPushHandler(printer *output.Printer) metadata.ManifestPushHandle
...
+func NewIndexPushHandler(printer *output.Printer) metadata.ManifestPushHandler {
+	return &IndexPushHandler{
+		printer: printer,
+	}
+}
+
+// OnTagged implements metadata.TaggedHandler.
+func (h *IndexPushHandler) OnTagged(_ ocispec.Descriptor, tag string) error {
+	return h.printer.Printf("%s  %s\n", status.IndexPromptTagged, tag)
+}
```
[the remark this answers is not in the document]
wangxiaoxuan273: This index handler is only used in `pushIndex` method, and `pushIndex` will be used by both `index create` and `index update`. So I think `IndexPushHandler` is a better name.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,201 @@
...
+		Short: "[Experimental] Create and push an index from provided manifests",
+		Long: `[Experimental] Create and push an index from provided manifests. All manifests should be in the same repository
+
+Example - create an index from source manifests tagged 'linux-amd64' and 'linux-arm64', and push without tagging:
+  oras manifest index create localhost:5000/hello linux-amd64 linux-arm64
+
+Example - create an index from source manifests tagged 'linux-amd64' and 'linux-arm64', and push with the tag 'latest':
+  oras manifest index create localhost:5000/hello:latest linux-amd64 linux-arm64
+
+Example - create an index from source manifests using both tags and digests, and push with tag 'latest':
```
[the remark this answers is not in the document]
wangxiaoxuan273: good catch. I forgot the colon.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, err
+	}
+	var platform ocispec.Platform
+	// ignore if the manifest does not have platform information
+	if err := json.Unmarshal(contentBytes, &platform); err != nil {
+		return nil, nil
+	}
+	if platform.Architecture == "" && platform.OS == "" {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
wangxiaoxuan273: I think it's better to separate these two cases for readability.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+			// check that all tags given by the user are valid tags
+			if opts.Reference != "" && !contentutil.IsValidTag(opts.Reference) {
+				return fmt.Errorf("%s is not a valid tag", opts.Reference)
+			}
+			for _, ref := range opts.extraRefs {
+				if !contentutil.IsValidTag(ref) {
+					return fmt.Errorf("%s is not a valid tag", ref)
+				}
+			}
```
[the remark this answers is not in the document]
wangxiaoxuan273: Then it's weird to separate the check for `opts.Reference` and `opts.extraRefs` before and after `Parse`.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+	indexBytes, _ := json.Marshal(index)
+	desc := content.NewDescriptorFromBytes(ocispec.MediaTypeImageIndex, indexBytes)
+	opts.Printf("%s  %s  %s\n", status.IndexPromptPacked, descriptor.ShortDigest(desc), ocispec.MediaTypeImageIndex)
+	return pushIndex(ctx, target, desc, indexBytes, opts.Reference, opts.extraRefs, opts.AnnotatedReference(), opts.Printer)
+}
+
+func resolveSourceManifests(ctx context.Context, target oras.ReadOnlyTarget, opts createOptions) ([]ocispec.Descriptor, error) {
+	resolved := []ocispec.Descriptor{}
+	for _, source := range opts.sources {
+		opts.Printf("%s  %s\n", status.IndexPromptFetching, source)
```
qweeah, line 127: No need to add extra space.
```suggestion
		opts.Println(status.IndexPromptFetching, source)
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+		desc, content, err := oras.FetchBytes(ctx, target, source, oras.DefaultFetchBytesOptions)
+		if err != nil {
+			return nil, fmt.Errorf("could not find the manifest %s: %w", source, err)
+		}
+		if !descriptor.IsManifest(desc) {
+			return nil, fmt.Errorf("%s is not a manifest", source)
+		}
+		opts.Printf("%s  %s\n", status.IndexPromptFetched, source)
```
qweeah, line 135:
```suggestion
		opts.Println(status.IndexPromptFetched, source)
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	// fetch config content
+	contentBytes, err := content.FetchAll(ctx, target, manifest.Config)
+	if err != nil {
+		return nil, err
+	}
+	var platform ocispec.Platform
+	// ignore if the manifest does not have platform information
+	if err := json.Unmarshal(contentBytes, &platform); err != nil || (platform.Architecture == "" && platform.OS == "") {
+		return nil, nil
```
qweeah, line 161:
```suggestion
	if err := json.Unmarshal(contentBytes, &platform); err != nil || (platform.Architecture == "" && platform.OS == "") {   
	        // ignore if the manifest does not contain valid platform information
		return nil, nil
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+	index := ocispec.Index{
+		Versioned: specs.Versioned{
+			SchemaVersion: 2,
+		},
+		MediaType: ocispec.MediaTypeImageIndex,
+		Manifests: manifests,
+	}
+	indexBytes, _ := json.Marshal(index)
+	desc := content.NewDescriptorFromBytes(ocispec.MediaTypeImageIndex, indexBytes)
+	opts.Printf("%s  %s  %s\n", status.IndexPromptPacked, descriptor.ShortDigest(desc), ocispec.MediaTypeImageIndex)
```
qweeah, line 120:
```suggestion
        opts.Println(status.IndexPromptPacked, descriptor.ShortDigest(desc), ocispec.MediaTypeImageIndex)
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+	var err error
+	if ref == "" {
+		err = target.Push(ctx, desc, bytes.NewReader(content))
+	} else {
+		_, err = oras.TagBytes(ctx, target, desc.MediaType, content, ref)
+	}
+	if err != nil {
+		return err
+	}
+	printer.Printf("%s  %s\n", status.IndexPromptPushed, path)
```
qweeah, line 177:
```suggestion
        opts.Println(status.IndexPromptPushed, opts.AnnotatedReference())
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	var platform ocispec.Platform
+	// ignore if the manifest does not have platform information
+	if err := json.Unmarshal(contentBytes, &platform); err != nil || (platform.Architecture == "" && platform.OS == "") {
+		return nil, nil
+	}
+	return &platform, nil
+}
+
+func pushIndex(ctx context.Context, target oras.Target, desc ocispec.Descriptor, content []byte, ref string, extraRefs []string, path string, printer *output.Printer) error {
```
qweeah, line 166:
```suggestion
func pushIndex(ctx context.Context, target oras.Target, desc ocispec.Descriptor, content []byte, opts createOptions) error {
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,186 @@
...
+		_, err = oras.TagBytes(ctx, target, desc.MediaType, content, ref)
+	}
+	if err != nil {
+		return err
+	}
+	printer.Printf("%s  %s\n", status.IndexPromptPushed, path)
+	if len(extraRefs) != 0 {
+		handler := display.NewIndexCreateHandler(printer)
+		tagListener := listener.NewTaggedListener(target, handler.OnTagged)
+		if _, err = oras.TagBytesN(ctx, tagListener, desc.MediaType, content, extraRefs, oras.DefaultTagBytesNOptions); err != nil {
```
qweeah, line 181:
```suggestion
		if _, err = oras.TagBytesN(ctx, tagListener, desc.MediaType, content, opts.extraRefs, oras.DefaultTagBytesNOptions); err != nil {
```

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: I'd like to reuse this method in `index update`, so cannot change the input argument to `createOptions`

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+			It("should show help doc with alias", func() {
+				ORAS("manifest", "index", "create", "--help").MatchKeyWords("Aliases", "pack").Exec()
+			})
+		})
+	})
+})
+
+var _ = Describe("1.1 registry users:", func() {
+	When("running `manifest index create`", func() {
+
```
qweeah, line 38: NIT: extra line

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+			// verify
+			ORAS("manifest", "fetch", RegistryRef(ZOTHost, ImageRepo, "index-create-by-digest")).
+				MatchKeyWords("amd64", "arm64").Exec()
+		})
+
+		It("should create index by using source manifest tags", func() {
+			// tag manifests for testing purpose
+			ORAS("tag", RegistryRef(ZOTHost, ImageRepo, string(multi_arch.LinuxAMD64.Digest)), "amd64").Exec()
+			ORAS("tag", RegistryRef(ZOTHost, ImageRepo, string(multi_arch.LinuxARM64.Digest)), "arm64").Exec()
+			ORAS("manifest", "index", "create", RegistryRef(ZOTHost, ImageRepo, "index-create-by-tag"),
```
qweeah, line 53: If those tags are needed, it's better to modify the OCI layout rather than changing test data without copying.

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+					"sha256:5c98cfc90e390c575679370a5dc5e37b52e854bbb7b9cb80cc1f30b56b8d183e").Exec()
+			// verify
+			ORAS("manifest", "fetch", RegistryRef(ZOTHost, ImageRepo, "index-create-by-tag")).
+				MatchKeyWords("arm64", "amd64").Exec()
+		})
+
+		It("should create index without tagging it", func() {
+			// tag manifests for testing purpose
+			ORAS("tag", RegistryRef(ZOTHost, ImageRepo, string(multi_arch.LinuxAMD64.Digest)), "amd64").Exec()
+			ORAS("tag", RegistryRef(ZOTHost, ImageRepo, string(multi_arch.LinuxARM64.Digest)), "arm64").Exec()
```
qweeah, line 65: Same as before. Better use `linux-amd64` rather than `amd64`

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+var _ = Describe("1.1 registry users:", func() {
+	When("running `manifest index create`", func() {
+
+		It("should create index by using source manifest digests", func() {
+			ORAS("manifest", "index", "create", RegistryRef(ZOTHost, ImageRepo, "index-create-by-digest"),
+				string(multi_arch.LinuxAMD64.Digest), string(multi_arch.LinuxARM64.Digest)).
+				MatchKeyWords("Fetched", "Pushed",
+					"sha256:cce9590b1193d8bcb70467e2381dc81e77869be4801c09abe9bc274b6a1d2001").Exec()
+			// verify
+			ORAS("manifest", "fetch", RegistryRef(ZOTHost, ImageRepo, "index-create-by-digest")).
```
qweeah, line 45: Validation is too simple here, need to be improved with below checks

1) can be unmarshalled to a index
2) `manifests` field should contain certain number of manifest entries
3) manifest with certain digest should match known platform

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+	})
+})
+
+var _ = Describe("1.1 registry users:", func() {
+	When("running `manifest index create`", func() {
+
+		It("should create index by using source manifest digests", func() {
+			ORAS("manifest", "index", "create", RegistryRef(ZOTHost, ImageRepo, "index-create-by-digest"),
+				string(multi_arch.LinuxAMD64.Digest), string(multi_arch.LinuxARM64.Digest)).
+				MatchKeyWords("Fetched", "Pushed",
```
qweeah, line 42: Can you add digest check into the status output check?

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+					"sha256:cce9590b1193d8bcb70467e2381dc81e77869be4801c09abe9bc274b6a1d2001").Exec()
```
qweeah, line 43: Is this digest of the crafted index? Better add metadata prmopt like `Digest: xxxx` so the spec is more readable.

test/e2e/suite/command/manifest_index.go
```diff
@@ -0,0 +1,118 @@
...
+	When("running manifest index command", func() {
+		When("running `manifest index create`", func() {
+			It("should show help doc with alias", func() {
+				ORAS("manifest", "index", "create", "--help").MatchKeyWords("Aliases", "pack").Exec()
+			})
+		})
+	})
+})
+
+var _ = Describe("1.1 registry users:", func() {
```
qweeah, line 36: Need a spec to test when the source image config is not a well-formed JSON.

qweeah (changes requested):

test/e2e/suite/command/manifest_index.go
wangxiaoxuan273: Added a test case of manifest with non conventional config

test/e2e/suite/command/manifest_index.go
wangxiaoxuan273: Added digest check

test/e2e/suite/command/manifest_index.go
wangxiaoxuan273: Added a validation function

qweeah (changes requested):

test/e2e/README.md
```diff
@@ -163,11 +163,18 @@ graph TD;
...
+        A9>tag: linux-amd64]-..->A2
+        A10>tag: linux-arm64]-..->A3
+        A11>tag: linux-armv7]-..->A4
 
         B0>tag: foobar]-..->B1[oci image]
         B1-- foo1 -->B2(blob1)
         B1-- foo2 -->B2(blob1)
         B1-- bar -->B3(blob2)
+
+        C0>tag: unusual-config]-..->C1[oci image]
```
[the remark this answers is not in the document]
wangxiaoxuan273: Nice catch. Thanks.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,184 @@
...
+	for _, source := range opts.sources {
+		opts.Println(status.IndexPromptFetching, source)
+		desc, content, err := oras.FetchBytes(ctx, target, source, oras.DefaultFetchBytesOptions)
+		if err != nil {
+			return nil, fmt.Errorf("could not find the manifest %s: %w", source, err)
+		}
+		if !descriptor.IsManifest(desc) {
+			return nil, fmt.Errorf("%s is not a manifest", source)
+		}
+		opts.Println(status.IndexPromptFetched, source)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Added a function `PlainDescriptor()` that shrinks a descriptor to the minimal.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,185 @@
...
+		return err
+	}
+	index := ocispec.Index{
+		Versioned: specs.Versioned{
+			SchemaVersion: 2,
+		},
+		MediaType: ocispec.MediaTypeImageIndex,
+		Manifests: manifests,
+	}
+	indexBytes, _ := json.Marshal(index)
```
Wwwsylvia, line 116: Curious why do we ignore the marshal error?

internal/descriptor/descriptor.go
```diff
@@ -37,6 +50,16 @@ func ShortDigest(desc ocispec.Descriptor) (digestString string) {
 	return digestString
 }
 
+// PlainDescriptor returns a plain descriptor that contains only MediaType, Digest and
+// Size.
+func PlainDescriptor(desc ocispec.Descriptor) ocispec.Descriptor {
```
Wwwsylvia, line 55: nit: We can simply call this function `Plain` as it is in the package `descriptor`, so it can be used like `descriptor.Plain(desc)`.
BTW should we have a unit test for this?

wangxiaoxuan273: Renamed the function

internal/contentutil/reference.go
```diff
@@ -15,10 +15,18 @@ limitations under the License.
...
 
 // IsDigest checks if the given string is a valid digest.
 func IsDigest(tagOrDigest string) bool {
 	_, err := digest.Parse(tagOrDigest)
 	return err == nil
 }
+
+// IsValidTag checks if the given string is a valid tag.
+func IsValidTag(tag string) bool {
+	return registry.Reference{Reference: tag}.ValidateReferenceAsTag() == nil
```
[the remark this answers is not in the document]
wangxiaoxuan273: Thanks Billy!

cmd/oras/internal/display/handler.go
```diff
@@ -174,6 +174,11 @@ func NewManifestPushHandler(printer *output.Printer) metadata.ManifestPushHandle
 	return text.NewManifestPushHandler(printer)
 }
 
+// NewIndexCreateHandler returns an index create handler.
+func NewIndexCreateHandler(printer *output.Printer) metadata.IndexCreateHandler {
```
shizhMSFT, line 178: Should it be named as `NewManifestIndexCreateHandler`?

cmd/oras/internal/display/metadata/interface.go
```diff
@@ -78,6 +78,11 @@ type ManifestPushHandler interface {
 	TaggedHandler
 }
 
+// IndexCreateHandler handles metadata output for index create events.
+type IndexCreateHandler interface {
```
shizhMSFT, line 82: Should it be named as `ManifestIndexCreateHandler`?

cmd/oras/internal/display/metadata/text/manifest_index_create.go
```diff
@@ -0,0 +1,39 @@
...
+package text
+
+import (
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras/cmd/oras/internal/display/metadata"
+	"oras.land/oras/cmd/oras/internal/output"
+)
+
+// IndexCreateHandler handles text metadata output for index create events.
+type IndexCreateHandler struct {
```
shizhMSFT, line 25: Should it be named as `ManifestIndexCreateHandler`?

cmd/oras/internal/display/handler.go
shizhMSFT: The name should be consistent with the command name.

cmd/oras/internal/display/metadata/interface.go
shizhMSFT: The name should be consistent with the command name.

cmd/oras/internal/display/metadata/text/manifest_index_create.go
shizhMSFT: The name should be consistent with the command name.

internal/descriptor/descriptor.go
```diff
@@ -37,6 +50,16 @@ func ShortDigest(desc ocispec.Descriptor) (digestString string) {
 	return digestString
 }
 
+// Plain returns a plain descriptor that contains only MediaType, Digest and
+// Size.
+func Plain(desc ocispec.Descriptor) ocispec.Descriptor {
```
shizhMSFT, line 55: This function seems to be copied from `oras-go`. Can you add a reference in the comment block?

internal/descriptor/descriptor.go
```diff
@@ -21,6 +21,19 @@ import (
 	"oras.land/oras/internal/docker"
 )
 
+// IsManifest checks if a descriptor describes a manifest.
+func IsManifest(desc ocispec.Descriptor) bool {
```
shizhMSFT, line 25: This function seems to be copied from `oras-go` with slight modification. Can you add a reference in the comment block?

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,185 @@
...
+  oras manifest index create localhost:5000/hello linux-amd64 linux-arm64
+
+Example - create an index from source manifests tagged 'linux-amd64' and 'linux-arm64', and push with the tag 'latest':
+  oras manifest index create localhost:5000/hello:latest linux-amd64 linux-arm64
+
+Example - create an index from source manifests using both tags and digests, and push with tag 'latest':
+  oras manifest index create localhost:5000/hello:latest linux-amd64 sha256:xxx
+
+Example - create an index and push it with multiple tags:
+  oras manifest index create localhost:5000/hello:tag1,tag2,tag3 linux-amd64 linux-arm64 sha256:xxx
```
shizhMSFT, line 67: Try not to use `latest` as the example tag as it is a bad practice. You may follow `oras pull --help` to use `v1` as the example tag.

shizhMSFT, line 67: Try not to use `sha256:xxx` in the example. Use a real digest instead.

shizhMSFT: See `oras manifest delete --help`.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,185 @@
...
+			allRefs := opts.extraRefs
+			if opts.Reference != "" {
+				allRefs = append(opts.extraRefs, opts.Reference)
+			}
+			for _, ref := range allRefs {
+				if !contentutil.IsValidTag(ref) {
+					return fmt.Errorf("%s is not a valid tag", ref)
+				}
+			}
+			return nil
```
shizhMSFT, line 88: Let's do this check in another PR along with other commands.

shizhMSFT: Please also open an issue to track this if there isn't one.

internal/contentutil/reference.go
```diff
@@ -15,10 +15,18 @@ limitations under the License.
...
 // IsDigest checks if the given string is a valid digest.
 func IsDigest(tagOrDigest string) bool {
 	_, err := digest.Parse(tagOrDigest)
 	return err == nil
 }
+
+// IsValidTag checks if the given string is a valid tag.
+func IsValidTag(tag string) bool {
+	return registry.Reference{Reference: tag}.ValidateReferenceAsDigest() != nil
+}
```
shizhMSFT, line 32: Do you mean
```suggestion
func IsTag(tagOrDigest string) bool {
	return registry.Reference{Reference: tagOrDigest}.ValidateReferenceAsTag() == nil
}
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return err
+	}
+	opts.Printf("%s  %s  %s\n", status.IndexPromptPacked, descriptor.ShortDigest(desc), ocispec.MediaTypeImageIndex)
+	return pushIndex(ctx, target, desc, content, opts.Reference, opts.extraRefs, opts.AnnotatedReference(), opts.Printer)
+}
+
+func resolveSourceManifests(ctx context.Context, target oras.ReadOnlyTarget, opts createOptions) ([]ocispec.Descriptor, error) {
+	resolved := []ocispec.Descriptor{}
+	for _, source := range opts.sources {
```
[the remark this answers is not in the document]
shizhMSFT: Let's do a simple version here. Later, we improve it and implement it in `oras-go`.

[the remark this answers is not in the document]
shizhMSFT: Please create an issue in `oras` for perf improvement

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,185 @@
...
+	desc := content.NewDescriptorFromBytes(ocispec.MediaTypeImageIndex, indexBytes)
+	opts.Println(status.IndexPromptPacked, descriptor.ShortDigest(desc), ocispec.MediaTypeImageIndex)
+	return pushIndex(ctx, target, desc, indexBytes, opts.Reference, opts.extraRefs, opts.AnnotatedReference(), opts.Printer)
+}
+
+func fetchSourceManifests(ctx context.Context, target oras.ReadOnlyTarget, opts createOptions) ([]ocispec.Descriptor, error) {
+	resolved := []ocispec.Descriptor{}
+	for _, source := range opts.sources {
+		opts.Println(status.IndexPromptFetching, source)
+		desc, content, err := oras.FetchBytes(ctx, target, source, oras.DefaultFetchBytesOptions)
```
shizhMSFT, line 126: note: this is a bit dangerous if the target is an OCI layout when someone is trying to fetch a large blob, dumping everything into memory, which is doable. The good thing is that `oras-go` protects you at 4 MiB with `DefaultFetchBytesOptions`.

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func getPlatform(ctx context.Context, target oras.ReadOnlyTarget, manifestBytes []byte) (*ocispec.Platform, error) {
+	// extract config descriptor
+	var manifest ocispec.Manifest
+	if err := json.Unmarshal(manifestBytes, &manifest); err != nil {
+		return nil, err
+	}
+	// fetch config content
+	contentBytes, err := content.FetchAll(ctx, target, manifest.Config)
```
shizhMSFT, line 153: security concerns: do you want to check the config size before fetching? We probably don't want to fetch a huge config blob into memory.

internal/contentutil/reference.go
wangxiaoxuan273: See: https://github.com/oras-project/oras/pull/1475#discussion_r1732233944

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: See https://github.com/oras-project/oras/pull/1475#discussion_r1732233944 and #1479

internal/contentutil/reference.go
wangxiaoxuan273: Removed this part of code.

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: added a descriptor size check before `FetchAll`.

internal/descriptor/descriptor.go
```diff
@@ -21,6 +21,20 @@ import (
 	"oras.land/oras/internal/docker"
 )
 
+// IsManifest checks if a descriptor describes a manifest.
+// Copied from `oras-go`: https://github.com/oras-project/oras-go/blob/d6c837e439f4c567f8003eab6e423c22900452a8/internal/descriptor/descriptor.go#L67
```
shizhMSFT, line 25:
```suggestion
// Adapted from `oras-go`: https://github.com/oras-project/oras-go/blob/d6c837e439f4c567f8003eab6e423c22900452a8/internal/descriptor/descriptor.go#L67
```

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,180 @@
...
+
+func getPlatform(ctx context.Context, target oras.ReadOnlyTarget, manifestBytes []byte) (*ocispec.Platform, error) {
+	// extract config descriptor
+	var manifest ocispec.Manifest
+	if err := json.Unmarshal(manifestBytes, &manifest); err != nil {
+		return nil, err
+	}
+	// if config size is larger than 4 MiB, discontinue the fetch
+	if manifest.Config.Size >= maxConfigSize {
+		return nil, fmt.Errorf("config is too large")
```
shizhMSFT, line 145: requires a more meaningful error message

wangxiaoxuan273: updated the error message

shizhMSFT (approved):
LGTM

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func getPlatform(ctx context.Context, target oras.ReadOnlyTarget, manifestBytes []byte) (*ocispec.Platform, error) {
+	// extract config descriptor
+	var manifest ocispec.Manifest
+	if err := json.Unmarshal(manifestBytes, &manifest); err != nil {
+		return nil, err
+	}
+	// if config size is larger than 4 MiB, discontinue the fetch
+	if manifest.Config.Size >= maxConfigSize {
```
Wwwsylvia, line 145: Do you mean `>`?

cmd/oras/root/manifest/index/create.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, fmt.Errorf("config size %v exceeds MaxBytes %v: %w", manifest.Config.Size, maxConfigSize, errdef.ErrSizeExceedsLimit)
+	}
+	// fetch config content
+	contentBytes, err := content.FetchAll(ctx, target, manifest.Config)
```
Wwwsylvia, line 149: Just a thought, wondering if it would be more convenient to use `oras.FetchBytes` here

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: Good catch! thanks

cmd/oras/root/manifest/index/create.go
Wwwsylvia: Oh I see, it's because we need to pass a `Repository.Blobs()` to `FetchBytes`, which is not avaialble in the current context.

Wwwsylvia (approved):
LGTM

qweeah (approved):
LGTM

PR #1481, opened 2024-08-16
refactor: create an interface for status console

TerryHowe:

Create a interface for console so that alternate implementations can exist. In doing this, I found the `Size()` method was not properly implemented, so I created a `GetHeightWidth()` method.

Partial: https://github.com/oras-project/oras/issues/1461

### DIFF
--- a/cmd/oras/internal/display/status/console/console.go
+++ b/cmd/oras/internal/display/status/console/console.go
@@ -16,10 +16,9 @@ limitations under the License.
 package console
 
 import (
-	"os"
-
-	"github.com/containerd/console"
+	containerd "github.com/containerd/console"
 	"github.com/morikuni/aec"
+	"os"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


sabre1041 (approved):
LGTM

sabre1041 (approved):

cmd/oras/internal/display/status/console/console.go
```diff
@@ -33,60 +33,75 @@ const (
...
-	c, err := console.ConsoleFromFile(f)
-	if err != nil {
-		return nil, err
-	}
-	return &Console{c}, nil
+// GetHeightWidth returns the width and height of the console.
+func (c *console) GetHeightWidth() (height, width int) {
+	windowSize, _ := c.Size()
+	return int(windowSize.Height), int(windowSize.Width)
 }
```
qweeah, line 81: Can we merge the implementation to those two functions? It's strange to make `Size` public but `console` is private.

qweeah (changes requested):

qweeah (approved):
Thanks Terry for contributing. LGTM with minor sugggestions.

qweeah (approved):
LGTM with minor suggestions.

qweeah (approved):
LGTM

PR #1482, opened 2024-08-18
chore: reset test coverage package list

qweeah:
Before this change, the nested $() in Makefile `$(GO_EXE) list ./... | grep -v internal/testutils | tr '\n' ','` generates empty return value so `-coverpkg` is actually set to `''`


TerryHowe (approved):
/lgtm

PR #1484, opened 2024-08-19
fix: copy prompt skipped was missing

TerryHowe:

This is a bit confusing, but previously, OnCopySkipped prompted with "Exists" and PostCopy prompted with "Skipped". I think this was changed accidentally.

https://github.com/oras-project/oras/blob/release-1.2/cmd/oras/root/cp.go#L180:L189


qweeah (approved):
Nice catch, added unit test for that.

cmd/oras/internal/display/status/text.go
```diff
@@ -161,12 +162,12 @@ func (ch *TextCopyHandler) PreCopy(_ context.Context, desc ocispec.Descriptor) e
...
-	successors, err := graph.FilteredSuccessors(ctx, desc, ch.fetcher, DeduplicatedFilter(ch.committed))
+	deduplicated, err := graph.FilteredSuccessors(ctx, desc, ch.fetcher, DeduplicatedFilter(ch.committed))
 	if err != nil {
 		return err
 	}
-	for _, successor := range successors {
-		if err = ch.printer.PrintStatus(successor, copyPromptExists); err != nil {
+	for _, successor := range deduplicated {
+		if err = ch.printer.PrintStatus(successor, copyPromptSkipped); err != nil {
 			return err
```
qweeah, line 171: BTW, we need to create an interface for `output.Printer` otherwise it's impossible to cover this line like tty prompter.

FeynmanZhou:
The CodeCov scan seems failed in the CI. Should we fix it before merging this PR?

TerryHowe:
Coverage on this file is 90% from my reading, but it just happens that renaming `successors` happens to touch a part of the code that is not tested. I think it makes sense to merge anyway here.

TerryHowe:
I'm going to ignore the bogus coverage report.

Issue #1066, opened 2023-08-18
support `--platform` in `oras push` and `oras attach`


qweeah:
### What is the version of your ORAS CLI

oras v1.1.0-rc.1

### What would you like to be added?

Add a new flag `--platform`(flag name TBD) to allow baking platform information into config data like
```shell
oras push localhost:5000/oras:v1.0.1 --artifact-type "application/vnd.example+type" --platform linux/amd64 oras--1.0.1.arm64_monterey.bottle.tar.gz
oras attach localhost:5000/oras:v1.0.1 --artifact-type "application/vnd.sbom" --platform linux/amd64 oras--1.0.1.arm64_monterey.bottle.tar.gz
```

The specified platform information will be stored in the config blob like https://oras.land/docs/concepts/artifact#artifacts-with-config

### Why is this needed for ORAS?

User can easily build artifact with platform information without composing config blob on their own.

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

amands98:
Should I include support for a platform flag ? // @qweeah

qweeah:
@amands98 Thanks for volunteering! Please make sure the baked config is following https://github.com/opencontainers/image-spec/blob/main/config.md#properties

shizhMSFT:
What's the media type of the config? Is it artifact-type or the standard config type? How is it supported by multiple cloud providers?

wangxiaoxuan273:
Looks like `--platform` for `oras attach` is already completed: #813 #1309

qweeah:
Still need to apply this to `oras attach`

TerryHowe:
`oras atach` has a platform option although I haven't tried it

Wwwsylvia:
> Still need to apply this to oras attach

Do you mean we can add a `--artifact-platform` flag to the manifest of the config referrers, just like we do for `oras push`?

> oras atach has a platform option although I haven't tried it

That is to control which arch-specific manifest to attach on.

---
But what are the use cases of adding platform to a referrer?🤔

FeynmanZhou:
> Still need to apply this to `oras attach`

Since we lack of the scenario for `oras attach`, we will keep it in ORAS future milestone.

TerryHowe:
I think a new issue for attach would make it easier

Wwwsylvia:
I just opened a new issue for `oras attach`: https://github.com/oras-project/oras/issues/1537
I think we are good to close this one.

PR #1500, opened 2024-09-16
feat: add platform option to push command

TerryHowe:

Add the platform option to the push command for example:
```
oras push --plain-http localhost:15000/oras:darwin,arm64  --artifact-platform darwin/arm64 --artifact-type 'application/vnd.oci.image.config.v1+json'  bin/darwin/arm64/oras:application/x-elf
```

Create the manifest for example:
```
oras manifest index create localhost:15000/oras:v1 sha256:ed3691788db69623790c46256dd11d8d6edc31be6e4762b53c74dfc3bb6792cf
```

The results
```
% oras manifest fetch --pretty localhost:15000/oras@sha256:c9e54542075ad502900d11b9b5d5bdb4b974ad580842515684d3fbb4849e2bb1
{
  "schemaVersion": 2,
  "artifactType": "application/vnd.oci.image.config.v1+json",
  "config": {
    "digest": "sha256:e9302bbb2fb8f6c2df866d3c4e41917849442f89a575f36f43366a7279804f70",
    "size": 38
  },
  "layers": [
    {
      "mediaType": "application/x-elf",
      "digest": "sha256:69c16b8386b7949fe37ff2356296b6331c1e6d5fbbc3e616192cc10ece55ea8d",
      "size": 11673840,
      "annotations": {
        "org.opencontainers.image.title": "bin/darwin/amd64/oras"
      }
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2024-09-16T16:54:48Z"
  }
}
% oras manifest fetch-config --pretty localhost:15000/oras@sha256:c9e54542075ad502900d11b9b5d5bdb4b974ad580842515684d3fbb4849e2bb1
{
  "architecture": "amd64",
  "os": "darwin"
}
```
Closes https://github.com/oras-project/oras/issues/1066
Partial https://github.com/oras-project/oras/issues/1053

### DIFF
--- a/cmd/oras/internal/option/platform.go
+++ b/cmd/oras/internal/option/platform.go
@@ -40,7 +40,7 @@ func (opts *Platform) ApplyFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&opts.platform, "platform", "", "", opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
 }
 
-// parse parses the input platform flag to an oci platform type.
+// Parse parses the input platform flag to an oci platform type.
 func (opts *Platform) Parse(*cobra.Command) error {
 	if opts.platform == "" {
 		return nil
@@ -72,4 +72,20 @@ func (opts *Platform) Parse(*cobra.Command) error {
 	}
 	opts.Platform = &p
 	return nil
+}
+
+// ArtifactPlatform option struct.
+type ArtifactPlatform struct {
+	Platform
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *ArtifactPlatform) ApplyFlags(fs *pflag.FlagSet) {
+	opts.FlagDescription = "set artifact platform"
+	fs.StringVarP(&opts.platform, "artifact-platform", "", "", "[Experimental] "+opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
+}
+
+// Parse parses the input platform flag to an oci platform type.
+func (opts *ArtifactPlatform) Parse(cmd *cobra.Command) error {
+	return opts.Platform.Parse(cmd)
 }
--- a/cmd/oras/root/push.go
+++ b/cmd/oras/root/push.go
@@ -16,6 +16,8 @@ limitations under the License.
 package root
 
 import (
+	"bytes"
+	"encoding/json"
 	"errors"
 	"strings"
 
@@ -41,6 +43,7 @@ import (
 type pushOptions struct {
 	option.Common
 	option.Packer
+	option.ArtifactPlatform
 	option.ImageSpec
 	option.Target
 	option.Format
@@ -100,6 +103,9 @@ Example - Push repository with manifest annotations:
 Example - Push repository with manifest annotation file:
   oras push --annotation-file annotation.json localhost:5000/hello:v1
 
+Example - Push artifact to repository with platform:
+  oras push --artifact-platform linux/arm/v5 localhost:5000/hello:v1
+
 Example - Push file "hi.txt" with multiple tags:
   oras push localhost:5000/hello:tag1,tag2,tag3 hi.txt
 
@@ -132,6 +138,10 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 						Recommendation: "set an artifact type via `--artifact-type` or consider image spec v1.0",
 					}
 				}
+			}
+			configAndPlatform := []string{"config", "artifact-platform"}
+			if err := oerrors.CheckMutuallyExclusiveFlags(cmd.Flags(), configAndPlatform...); err != nil {
+				return err
 			}
 
 			switch opts.PackVersion {
@@ -177,6 +187,22 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 			return err
 		}
 		desc, err := addFile(ctx, store, option.AnnotationConfig, cfgMediaType, path)
+		if err != nil {
+			return err
+		}
+		desc.Annotations = packOpts.ConfigAnnotations
+		packOpts.ConfigDescriptor = &desc
+	} else if opts.Platform.Platform != nil {
+		blob, err := json.Marshal(opts.Platform.Platform)
+		if err != nil {
+			return err
+		}
+		mediaType := oras.MediaTypeUnknownConfig
+		if opts.Flag == option.ImageSpecV1_0 && opts.artifactType != "" {
+			mediaType = opts.artifactType
+		}
+		desc := content.NewDescriptorFromBytes(mediaType, blob)
+		err = store.Push(ctx, desc, bytes.NewReader(blob))
 		if err != nil {
 			return err
 		}
--- a/test/e2e/internal/testdata/foobar/const.go
+++ b/test/e2e/internal/testdata/foobar/const.go
@@ -38,6 +38,21 @@ var (
 		Digest: "46b68ac1696c", Name: "application/vnd.unknown.config.v1+json",
 	}
 
+	PlatformConfigSize     = 38
+	PlatformConfigDigest   = digest.Digest("sha256:e94c0ba80a1157ffab5b5c6656fffc089c6446c7ed0604f3382910d1ef7dd40d")
+	PlatformConfigStateKey = match.StateKey{
+		Digest: "e94c0ba80a11", Name: "application/vnd.unknown.config.v1+json",
+	}
+
+	PlatformV10ConfigSize     = 38
+	PlatformV10ConfigDigest   = digest.Digest("sha256:e94c0ba80a1157ffab5b5c6656fffc089c6446c7ed0604f3382910d1ef7dd40d")
+	PlatformV10ConfigStateKey = match.StateKey{
+		Digest: "e94c0ba80a11", Name: "test/artifact+json",
+	}
+	PlatformV1DEfaultConfigStateKey = match.StateKey{
+		Digest: "e94c0ba80a11", Name: "application/vnd.unknown.config.v1+json",
+	}
+
 	FileBarName     = "foobar/bar"
 	FileBarStateKey = match.StateKey{Digest: "fcde2b2edba5", Name: FileLayerNames[2]}
 	FileStateKeys   = []match.StateKey{
--- a/test/e2e/suite/command/push.go
+++ b/test/e2e/suite/command/push.go
@@ -77,6 +77,14 @@ var _ = Describe("ORAS beginners:", func() {
 			ORAS("push", ref, "--config", foobar.FileConfigName, "--artifact-type", "test/artifact+json", "--image-spec", "v1.0").ExpectFailure().WithWorkDir(tempDir).Exec()
 		})
 
+		It("should fail to use --artifact-platform and --config at the same time", func() {
+			tempDir := PrepareTempFiles()
+			repo := pushTestRepo("no-mediatype")
+			ref := RegistryRef(ZOTHost, repo, "")
+
+			ORAS("push", ref, "--artifact-platform", "linux/amd64", "--config", foobar.FileConfigName).ExpectFailure().WithWorkDir(tempDir).Exec()
+		})
+
 		It("should fail if image spec is not valid", func() {
 			testRepo := attachTestRepo("invalid-image-spec")
 			subjectRef := RegistryRef(ZOTHost, testRepo, foobar.Tag)
@@ -609,6 +617,66 @@ var _ = Describe("OCI image layout users:", func() {
 				MediaType: configType,
 				Size:      int64(foobar.FileConfigSize),
 				Digest:    foobar.FileConfigDigest,
+			}))
+		})
+
+		It("should push files with platform", func() {
+			tempDir := PrepareTempFiles()
+			ref := LayoutRef(tempDir, tag)
+			ORAS("push", Flags.Layout, ref, "--artifact-platform", "darwin/arm64", foobar.FileBarName, "-v").
+				MatchStatus([]match.StateKey{
+					foobar.PlatformConfigStateKey,
+					foobar.FileBarStateKey,
+				}, true, 2).
+				WithWorkDir(tempDir).Exec()
+			// validate
+			fetched := ORAS("manifest", "fetch", Flags.Layout, ref).Exec().Out.Contents()
+			var manifest ocispec.Manifest
+			Expect(json.Unmarshal(fetched, &manifest)).ShouldNot(HaveOccurred())
+			Expect(manifest.Config).Should(Equal(ocispec.Descriptor{
+				MediaType: foobar.PlatformConfigStateKey.Name,
+				Size:      int64(foobar.PlatformConfigSize),
+				Digest:    foobar.PlatformConfigDigest,
+			}))
+		})
+
+		It("should push files with platform with mediaType as artifactType for v1.0", func() {
+			tempDir := PrepareTempFiles()
+			ref := LayoutRef(tempDir, tag)
+			ORAS("push", Flags.Layout, ref, "--image-spec", "v1.0", "--artifact-type", "test/artifact+json", "--artifact-platform", "darwin/arm64", foobar.FileBarName, "-v").
+				MatchStatus([]match.StateKey{
+					foobar.PlatformV10ConfigStateKey,
+					foobar.FileBarStateKey,
+				}, true, 2).
+				WithWorkDir(tempDir).Exec()
+			// validate
+			fetched := ORAS("manifest", "fetch", Flags.Layout, ref).Exec().Out.Contents()
+			var manifest ocispec.Manifest
+			Expect(json.Unmarshal(fetched, &manifest)).ShouldNot(HaveOccurred())
+			Expect(manifest.Config).Should(Equal(ocispec.Descriptor{
+				MediaType: "test/artifact+json",
+				Size:      int64(foobar.PlatformV10ConfigSize),
+				Digest:    foobar.PlatformV10ConfigDigest,
+			}))
+		})
+
+		It("should push files with platform with no artifactType for v1.0", func() {
+			tempDir := PrepareTempFiles()
+			ref := LayoutRef(tempDir, tag)
+			ORAS("push", Flags.Layout, ref, "--image-spec", "v1.0", "--artifact-platform", "darwin/arm64", foobar.FileBarName, "-v").
+				MatchStatus([]match.StateKey{
+					foobar.PlatformV1DEfaultConfigStateKey,
+					foobar.FileBarStateKey,
+				}, true, 2).
+				WithWorkDir(tempDir).Exec()
+			// validate
+			fetched := ORAS("manifest", "fetch", Flags.Layout, ref).Exec().Out.Contents()
+			var manifest ocispec.Manifest
+			Expect(json.Unmarshal(fetched, &manifest)).ShouldNot(HaveOccurred())
+			Expect(manifest.Config).Should(Equal(ocispec.Descriptor{
+				MediaType: "application/vnd.unknown.config.v1+json",
+				Size:      int64(foobar.PlatformV10ConfigSize),
+				Digest:    foobar.PlatformV10ConfigDigest,
 			}))
 		})
 


cmd/oras/root/push.go
```diff
@@ -97,6 +101,9 @@ Example - Push repository with manifest annotations:
 Example - Push repository with manifest annotation file:
   oras push --annotation-file annotation.json localhost:5000/hello:v1
 
+Example - Push repository with platform:
+  oras push --platform linux/arm/v5 localhost:5000/hello:v1
```
qweeah, line 105: CC @FeynmanZhou for the flag naming. `attach` already using it so better call the new one `--config-platform`

cmd/oras/root/push.go
```diff
@@ -41,6 +44,7 @@ import (
 type pushOptions struct {
 	option.Common
 	option.Packer
+	option.Platform
```
qweeah, line 47: Should add this to `Packer` options

cmd/oras/root/push.go
```diff
@@ -152,6 +162,7 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 	cmd.Flags().IntVarP(&opts.concurrency, "concurrency", "", 5, "concurrency level")
 	opts.SetTypes(option.FormatTypeText, option.FormatTypeJSON, option.FormatTypeGoTemplate)
 	option.ApplyFlags(&opts, cmd.Flags())
+	cmd.MarkFlagsMutuallyExclusive("config", "platform")
```
qweeah, line 165: Use `CheckMutuallyExclusiveFlags` instead. The output of this function is hard to read.

cmd/oras/root/push.go
```diff
@@ -183,6 +194,22 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
...
+	} else if opts.Platform.Platform != nil {
+		blob, err := json.Marshal(opts.Platform.Platform)
+		if err != nil {
+			return err
+		}
+		desc := ocispec.Descriptor{
+			MediaType: oras.MediaTypeUnknownConfig,
+			Digest:    digest.FromBytes(blob),
+			Size:      int64(len(blob)),
+		}
```
wangxiaoxuan273, line 206: nit: This part can be replaced by `func NewDescriptorFromBytes(mediaType string, content []byte) ocispec.Descriptor` of oras-go. Reference: [NewDescriptorFromBytes](https://pkg.go.dev/oras.land/oras-go/v2@v2.5.0/content#NewDescriptorFromBytes)

cmd/oras/root/push.go
wangxiaoxuan273: But if this is added to `Packer`, we need to change `oras attach` since it already has a `--platform` flag.

qweeah: 1. Both `attach` and `push` should support customized platform. 
2. It's inconsistent if we use `--platform` for customizing config platform in `push`, while `--platform` for selecting subject platform in `attach`

cmd/oras/root/push.go
shizhMSFT: It is a good point that `--platform` means differently in `oras push` and `oras attach`. We need a better flag name.

shizhMSFT: `--config-platform` does not sound like a good name. Would it be better to name it `--artifact-platform`? That will be consistent with `--artifact-type`.

cmd/oras/root/push.go
```diff
@@ -136,6 +143,9 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 				if opts.manifestConfigRef != "" && opts.artifactType != "" {
 					return errors.New("--artifact-type and --config cannot both be provided for 1.0 OCI image")
 				}
+				if opts.Platform.Platform != nil && opts.artifactType != "" {
+					return errors.New("--artifact-type and --platform cannot both be provided for 1.0 OCI image")
+				}
```
shizhMSFT, line 148: They can be used together, can't they?

cmd/oras/root/push.go
```diff
@@ -183,6 +194,22 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 		}
 		desc.Annotations = packOpts.ConfigAnnotations
 		packOpts.ConfigDescriptor = &desc
+	} else if opts.Platform.Platform != nil {
+		blob, err := json.Marshal(opts.Platform.Platform)
+		if err != nil {
+			return err
+		}
+		desc := ocispec.Descriptor{
+			MediaType: oras.MediaTypeUnknownConfig,
```
shizhMSFT, line 203: This value depends on the image-spec version.

cmd/oras/root/push.go
Wwwsylvia: Indeed they can be used together. We can populate the `artifactType` as `mediaType` into the config descriptor.

cmd/oras/root/push.go
Wwwsylvia: If the `image-spec` version is `v1.0`. The media type can also be the value of `--artifact-type`.

cmd/oras/root/push.go
TerryHowe: Done.

cmd/oras/root/push.go
TerryHowe: Done.

cmd/oras/root/push.go
TerryHowe: Done.

cmd/oras/root/push.go
qweeah: Let not merge it to `Packer` since this PR is all about `push`

cmd/oras/root/push.go
TerryHowe: Done

Wwwsylvia (approved):
LGTM with nit, but I am not a maintainer.

cmd/oras/root/push.go
```diff
@@ -183,6 +194,24 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
...
 		desc.Annotations = packOpts.ConfigAnnotations
 		packOpts.ConfigDescriptor = &desc
+	} else if opts.Platform.Platform != nil {
+		blob, err := json.Marshal(opts.Platform.Platform)
+		if err != nil {
+			return err
+		}
+		mediaType := oras.MediaTypeUnknownConfig
+		if opts.Flag == option.ImageSpecV1_0 {
+			if opts.artifactType != "" {
```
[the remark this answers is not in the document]
TerryHowe: Done

sabre1041 (approved):
LGTM

cmd/oras/root/push.go
TerryHowe: This does mean pull will use `--platform` and push `--artifact-platform` though which does seem a bit inconsistent to me.

qweeah: It's hard to choose a good flag name.

Some historical view on this: `--platform` was firstly introduced to `oras manifest to select an image with specific platform from a multi-arch image. Then the same UX was applied to `oras discover` and `oras attach` for selecting a platform-specific subject artifact.

cmd/oras/internal/option/platform.go
```diff
@@ -73,3 +73,21 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
+// ArtifactPlatform option struct.
+type ArtifactPlatform struct {
+	Platform
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *ArtifactPlatform) ApplyFlags(fs *pflag.FlagSet) {
+	if opts.FlagDescription == "" {
+		opts.FlagDescription = "artifact platform"
+	}
```
qweeah, line 86:
```suggestion
	opts.FlagDescription = "set artifact platform"
```

cmd/oras/root/push.go
```diff
@@ -156,6 +162,11 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 }
 
 func runPush(cmd *cobra.Command, opts *pushOptions) error {
+	configAndPlatform := []string{"config", "artifact-platform"}
+	if err := oerrors.CheckMutuallyExclusiveFlags(cmd.Flags(), configAndPlatform...); err != nil {
+		return err
+	}
```
qweeah, line 168: This can be done in `PreRunE` before `option.Parse`

cmd/oras/root/push.go
```diff
@@ -116,7 +122,7 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 				return err
 			}
 
-			if opts.manifestConfigRef != "" && opts.artifactType == "" {
+			if (opts.manifestConfigRef != "" || opts.Platform.Platform != nil) && opts.artifactType == "" {
```
qweeah, line 125: This block handles `oras push --config xxx --artifact-type xxx ...`: 

- For v1.1 artifacts, this usage invalid. 
- Otherwise best-effort try to pack v1.0 artifacts.


Since `--artifact-platform` cannot be used with `--config`, there is no need to change this code block.

cmd/oras/internal/option/platform.go
```diff
@@ -73,3 +73,21 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
+	fs.StringVarP(&opts.platform, "artifact-platform", "", "", opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
```
qweeah, line 87:
```suggestion
	opts.FlagDescription = "set artifact platform"
	opts.Platform.AppFlags(fs)
```

qweeah (changes requested):

cmd/oras/root/push.go
qweeah: Only need to make sure: `--artifact-platform` is not used for v1.0 artifact in the below `opts.PackVersion` switching block.

TerryHowe: Made the change I think you requested. I do need to think about this some more though.

cmd/oras/root/push.go
```diff
@@ -130,12 +136,19 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
...
+			}
 
 			switch opts.PackVersion {
 			case oras.PackManifestVersion1_0:
 				if opts.manifestConfigRef != "" && opts.artifactType != "" {
 					return errors.New("--artifact-type and --config cannot both be provided for 1.0 OCI image")
 				}
+				if opts.Platform.Platform != nil && opts.artifactType != "" {
+					return errors.New("--artifact-platform and --artifact-type cannot both be provided for 1.0 OCI image")
+				}
```
TerryHowe, line 151: I'm not sure what the v1.0 rule should be.

qweeah: `--artifact-platform` specifies content of config blob, `--artifact-type` specifies `mediaType` of config blob, so both flags can be used for OCI v1.0 image.

cmd/oras/root/push.go
qweeah: > Only need to make sure: --artifact-platform is not used for v1.0 artifact in the below opts.PackVersion switching block.

This is wrong actually. Sorry for the back-and-forth change.

qweeah (changes requested):

cmd/oras/root/push.go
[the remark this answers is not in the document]
qweeah: This is need if we are going to allow v1.1 `application/vnd.unknown.artifact.v1` artifact to have platform information in config blob.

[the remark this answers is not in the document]
qweeah: @shizhMSFT If `--platform` is used without `--artifact-type` and `image-spec` is not provided, should we pack v1.1 or v1.0 image?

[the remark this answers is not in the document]
qweeah: Currently if `--config` is not provided and `image-spec` is not provided, `oras` will pack v1.1 artifact
```console
➜  ~ oras push localhost:5000/test:type
✓ Uploaded  application/vnd.oci.empty.v1+json                                                                   2/2  B 100.00%   34ms
  └─ sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
✓ Uploaded  application/vnd.oci.image.manifest.v1+json                                                      535/535  B 100.00%     0s
  └─ sha256:5f56fc622e5f67d225a27be2c3637a606880c23aa98a2561b3797c4b54ac9fd5
Pushed [registry] localhost:5000/test:type
ArtifactType: application/vnd.unknown.artifact.v1
Digest: sha256:5f56fc622e5f67d225a27be2c3637a606880c23aa98a2561b3797c4b54ac9fd5
➜  ~ oras manifest fetch --pretty localhost:5000/test@sha256:5f56fc622e5f67d225a27be2c3637a606880c23aa98a2561b3797c4b54ac9fd5
{
  "schemaVersion": 2,
  "artifactType": "application/vnd.unknown.artifact.v1",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "size": 2,
    "data": "e30="
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.empty.v1+json",
      "size": 2,
      "data": "e30="
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2024-10-10T06:29:05Z"
  }
}
```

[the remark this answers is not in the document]
qweeah: If `--config` is used with `--artifact-type`, `oras` will also pack v1.1 artifact
```console
➜  ~ oras push localhost:5000/test:v1.1 --artifact-type test/123 --config 123
✓ Exists    application/vnd.oci.empty.v1+json                                                                   2/2  B 100.00%     0s
  └─ sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
✓ Uploaded  application/vnd.unknown.config.v1+json                                                              3/3  B 100.00%   39ms
  └─ sha256:ca3d163bab055381827226140568f3bef7eaac187cebd76878e0b63e9e442356
✓ Uploaded  application/vnd.oci.image.manifest.v1+json                                                      499/499  B 100.00%     0s
  └─ sha256:d80e321ea40ab36b661693ddf507f425801819481949d6bdbb6ec52f503139ae
Pushed [registry] localhost:5000/test:v1.1
ArtifactType: test/123
Digest: sha256:d80e321ea40ab36b661693ddf507f425801819481949d6bdbb6ec52f503139ae
➜  ~ oras manifest fetch --pretty localhost:5000/test@sha256:d80e321ea40ab36b661693ddf507f425801819481949d6bdbb6ec52f503139ae
{
  "schemaVersion": 2,
  "artifactType": "test/123",
  "config": {
    "digest": "sha256:ca3d163bab055381827226140568f3bef7eaac187cebd76878e0b63e9e442356",
    "size": 3
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.empty.v1+json",
      "size": 2,
      "data": "e30="
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2024-10-10T06:34:43Z"
  }
}
```

If config media type is not provided, it will be set to `application/vnd.unknown.config.v1+json`. We can follow this practice for `--platform` and bake v1.1 artifact for `oras push --platform xxx --artifact xxx` usage.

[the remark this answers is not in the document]
shizhMSFT: > @shizhMSFT If `--platform` is used without `--artifact-type` and `image-spec` is not provided, should we pack v1.1 or v1.0 artifact? (I would prefer v1.1)

Let's follow the default value of `image-spec` when this flag is not provided.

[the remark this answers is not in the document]
qweeah: We make `oras push --config xxx` craft v1.0 artifact to avoid breaking change. `oras push --platform xxx` should create v1.1 artifact. The suggested change should be applied.

cmd/oras/root/push.go
qweeah:
```suggestion
```

TerryHowe: Since the code above this does not allow specifying `--artifact-type` and `--config` at the same time I thought it made sense to not allow `--artifact-typ` and `--artifact-platform` since platform is just a shortcut for config.

qweeah: We do not allow specifying `--artifact-type` and `--config` at the same time because **those flags function are overlapped, user might specify different media types of config descriptor**, like `--artifact-type aaa --config xxx:bbb`, `aaa` and `bbb` conflicts.

But there is no such potential conflict between `--artifact-type` and `--artifact-platform`

cmd/oras/root/push.go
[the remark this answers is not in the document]
TerryHowe: v1.1 image spec is default except for that case you mention `--config` and no `--artifact-type`

https://github.com/oras-project/oras/blob/main/cmd/oras/root/push.go#L122-L128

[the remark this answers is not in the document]
qweeah: Yes, and we should make v1.1 default when `--artifact-platform xxx` is used regardless of `--artifact-type`.

Having `|| opts.Platform.Platform != nil` here will make v1.0 default when `--artifact-platform` used without `--artifact-type`.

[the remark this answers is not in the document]
TerryHowe: Done

cmd/oras/root/push.go
```diff
@@ -182,6 +192,22 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
...
+	} else if opts.Platform.Platform != nil {
+		blob, err := json.Marshal(opts.Platform.Platform)
+		if err != nil {
+			return err
+		}
+		mediaType := oras.MediaTypeUnknownConfig
+		if opts.Flag == option.ImageSpecV1_0 && opts.artifactType != "" {
+			mediaType = opts.artifactType
+		}
+		desc := content.NewDescriptorFromBytes(mediaType, blob)
```
qweeah, line 204:
```suggestion
		mediaType := opts.artifactType
		if mediaType == "" {
			mediaType = opts.artifactType
		}
		desc := content.NewDescriptorFromBytes(mediaType, blob)
```

qweeah (approved):
LGTM with minor suggestion

cmd/oras/root/push.go
TerryHowe: Manually made this change because oras.MediaTypeUnknownConfig.

TerryHowe: With this change, the e2e test fails because `oras push --oci-layout /tmp/ginkgo1771247066:e2e --artifact-platform darwin/arm64 foobar/bar -v`

Results in:
```
{
  "schemaVersion": 2,
  "artifactType": "application/vnd.unknown.artifact.v1",
  "config": {
    "mediaType": "application/vnd.unknown.artifact.v1",
    "digest": "sha256:e94c0ba80a1157ffab5b5c6656fffc089c6446c7ed0604f3382910d1ef7dd40d",
    "size": 38
  },
  "layers": [
    {
      "digest": "sha256:d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1",
      "size": 5,
      "annotations": {
        "org.opencontainers.image.title": "foobar/bar"
      }
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2024-10-15T16:14:58Z"
  }
}
```

instead of 
```
{
  "schemaVersion": 2,
  "artifactType": "application/vnd.unknown.artifact.v1",
  "config": {
    "digest": "sha256:e94c0ba80a1157ffab5b5c6656fffc089c6446c7ed0604f3382910d1ef7dd40d",
    "size": 38
  },
  "layers": [
    {
      "digest": "sha256:d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1",
      "size": 5,
      "annotations": {
        "org.opencontainers.image.title": "foobar/bar"
      }
    }
  ],
  "annotations": {
    "org.opencontainers.image.created": "2024-10-15T16:20:21Z"
  }
}
```

qweeah: I think the former is preferred since user doesn't specify `--artifact-spec v1.0` or `--config`?

TerryHowe: This is a problem tho 
```
  "artifactType": "application/vnd.unknown.artifact.v1",
  "config": {
    "mediaType": "application/vnd.unknown.artifact.v1",
```

qweeah:
Sorry merging https://github.com/oras-project/oras/pull/1500/commits/55cf4260c98bed4c34b3cd612e44d0d0867774f5 is not intended, I was playing with the PR locally and the upstream is accidentally set to this PR.

TerryHowe:
> Sorry merging [55cf426](https://github.com/oras-project/oras/commit/55cf4260c98bed4c34b3cd612e44d0d0867774f5) is not intended, I was playing with the PR locally and the upstream is accidentally set to this PR.

reverted

cmd/oras/root/push.go
qweeah: I think switch-casing on the default value is a better option than setting the value.
```suggestion
                mediaType := opts.artifactType
		if mediaType == "" {
			switch opts.PackVersion {
			case oras.PackManifestVersion1_0:
				mediaType = ocispec.MediaTypeImageConfig
			case oras.PackManifestVersion1_1:
				mediaType = oras.MediaTypeUnknownArtifact
			}
		}
		desc := content.NewDescriptorFromBytes(mediaType, blob)
```

qweeah: > This is a problem tho
> 
> ```
>   "artifactType": "application/vnd.unknown.artifact.v1",
>   "config": {
>     "mediaType": "application/vnd.unknown.artifact.v1",
> ```

This is valid for v1.1 though.

TerryHowe: It may be valid, but I don't think that was the intention for the user

qweeah: User doesn't specify `--artifact-type` so we cannot know what `config.mediaType` user is expecting.

If user what to specify `artifactType` and `config.mediaType` for v1.1 at the same time, they should use `--artifact-type <artifactType> --config <path>:<config.mediaType>`.

qweeah:
```
  "artifactType": "application/vnd.unknown.artifact.v1",
  "config": {
    "mediaType": "application/vnd.unknown.artifact.v1",
```
This LGTM and it is compliant to [v1.1 artifact with config](https://oras.land/docs/concepts/artifact#artifacts-with-config)

TerryHowe: The reason the test coverage failed was because https://github.com/oras-project/oras/blob/main/cmd/oras/root/push.go#L144 means `opts.artifactType` is never empty here, so I got rid of all the unnecessary logic.

cmd/oras/root/push.go
```diff
@@ -182,6 +192,22 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
...
 		packOpts.ConfigDescriptor = &desc
+	} else if opts.Platform.Platform != nil {
+		blob, err := json.Marshal(opts.Platform.Platform)
+		if err != nil {
+			return err
+		}
+		mediaType := opts.artifactType
+		if mediaType == "" {
+			mediaType = oras.MediaTypeUnknownConfig
+		}
```
shizhMSFT, line 203: This logic works for OCI spec v1.0. However, it does not work very well for OCI spec v1.1. And I agree with you w.r.t. https://github.com/oras-project/oras/pull/1500#discussion_r1801538753

shizhMSFT: Your original code looks better to me
```go
mediaType := oras.MediaTypeUnknownConfig
if opts.Flag == option.ImageSpecV1_0 && opts.artifactType != "" {
	mediaType = opts.artifactType
}
```

shizhMSFT (changes requested):
In fact, I have a more fundamental question: How do I specify both `artifactType` and `config.mediaType`? It seems not possible for image-spec 1.1 in the current PR.

We need the ability to set `config.mediaType` to `application/vnd.oci.image.config.v1+json` when `--artifact-platform` is set in order to support [OCI volumes](https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/). /cc @FeynmanZhou

TerryHowe:
> In fact, I have a more fundamental question: How do I specify both `artifactType` and `config.mediaType`? It seems not possible for image-spec 1.1 in the current PR.

This is out of the scope for this PR and I don't think that was possible before this PR.

cmd/oras/internal/option/platform.go
```diff
@@ -73,3 +73,19 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
+// ArtifactPlatform option struct.
+type ArtifactPlatform struct {
+	Platform
+}
+
+// ApplyFlags applies flags to a command flag set.
+func (opts *ArtifactPlatform) ApplyFlags(fs *pflag.FlagSet) {
+	opts.FlagDescription = "set artifact platform"
+	fs.StringVarP(&opts.platform, "artifact-platform", "", "", opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
+}
```
shizhMSFT, line 86: Can you make this flag as `[Experimental]` or `[Preview]`?

cmd/oras/root/push.go
TerryHowe: The original code was clearer at least.

shizhMSFT:
> This is out of the scope for this PR and I don't think that was possible before this PR.

Can you mark the flag `--artifact-platform` as `[Experimental]` or `[Preview]` so that we can complete it in next few iterations / PRs?

cmd/oras/root/push.go
shizhMSFT: Can you apply your original code above to this PR?

shizhMSFT (approved):
LGTM

qweeah:
> In fact, I have a more fundamental question: How do I specify both artifactType and config.mediaType? It seems not possible for image-spec 1.1 in the current PR.

Through `oras push --artifact-type $ARTIFACT_TYPE --config $CONFIG_PATH:$CONFIG_TYPE`. It's by design `--artifact-platform` cannot be used to customize any media type.

qweeah:
> We need the ability to set config.mediaType to application/vnd.oci.image.config.v1+json when --artifact-platform is set in order to support [OCI volumes](https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/). /cc @FeynmanZhou

In that case `--image-spec v1.0 --artifact-type application/vnd.oci.image.config.v1+json --artifact-platform xxx/xxx`

shizhMSFT:
> > We need the ability to set config.mediaType to application/vnd.oci.image.config.v1+json when --artifact-platform is set in order to support [OCI volumes](https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/). /cc @FeynmanZhou
> 
> In that case `--image-spec v1.0 --artifact-type application/vnd.oci.image.config.v1+json --artifact-platform xxx/xxx`

This workaround should work.

qweeah:
> > > We need the ability to set config.mediaType to application/vnd.oci.image.config.v1+json when --artifact-platform is set in order to support [OCI volumes](https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/). /cc @FeynmanZhou
> > 
> > 
> > In that case `--image-spec v1.0 --artifact-type application/vnd.oci.image.config.v1+json --artifact-platform xxx/xxx`
> 
> This workaround should work.

It just came to me that if `application/vnd.oci.image.config.v1` is a critical scenario that `oras push --platform` need to support, why not make it default? This default media type switch will only happen when `--platform` being used and thus is not breaking.

TerryHowe:
> This default media type switch will only happen when `--artifact-platform` being used and thus is not breaking.

I'm not 100% following here, maybe create an issue if you think there is something. If `--artifact-platform` is used, it creates a 1.1 image right now. Without specifying the image version, it creates a 1.0 imager

qweeah:
@TerryHowe What I am thinking is to use `application/vnd.oci.image.config.v1` instead of `application/vnd.unknown.config.v1+json` as default config media type. 

Actually, it requires oras to extra work on baking the config blob to meet below spec: https://github.com/opencontainers/image-spec/blob/da92727e9c8761ec087890466b8756b755aefd37/manifest.md#L68-73

```
   When the `config.mediaType` is set to `application/vnd.oci.image.config.v1+json`, the following additional restrictions apply:

  - The array MUST have the base layer at index 0.
  - Subsequent layers MUST then follow in stack order (i.e. from `layers[0]` to `layers[len(layers)-1]`).
  - The final filesystem layout MUST match the result of [applying](layer.md#applying-changesets) the layers to an empty directory.
  - The [ownership, mode, and other attributes](layer.md#file-attributes) of the initial empty directory are unspecified
```

This work goes below the scope of oras so what I am thinking is not doable for now.

Wwwsylvia:
Just realized that the config media type has to be `application/vnd.oci.image.config.v1`, otherwise the end-to-end workflow won't work. See #1517 for details.

PR #1502, opened 2024-09-18
feat: `--output` for oras manifest `index update`

wangxiaoxuan273:

part of #1053 

### DIFF
--- a/cmd/oras/root/manifest/index/update.go
+++ b/cmd/oras/root/manifest/index/update.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -39,11 +40,13 @@ import (
 type updateOptions struct {
 	option.Common
 	option.Target
+	option.Pretty
 
 	addArguments    []string
 	mergeArguments  []string
 	removeArguments []string
 	tags            []string
+	outputPath      string
 }
 
 func updateCmd() *cobra.Command {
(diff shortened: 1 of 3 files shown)


qweeah:
Please make sure `--output -` make other stdout output silent. See https://github.com/oras-project/oras/issues/1503. Doesn't necessary need to be done in this PR.

wangxiaoxuan273:
> Please make sure `--output -` make other stdout output silent. See #1503. Doesn't necessarily need to be done in this PR.

I'd like to do it in a future PR.

cmd/oras/root/manifest/index/update.go
```diff
@@ -64,9 +67,18 @@ Example - Merge manifests from the index 'v2-windows' to the index 'v2':
 
 Example - Update an index and tag the updated index as 'v2.1.0' and 'v2':
   oras manifest index update localhost:5000/hello@sha256:99e4703fbf30916f549cd6bfa9cdbab614b5392fbe64fdee971359a77073cdf9 --add linux-amd64 --tag "v2.1.0" --tag "v2"
+
+Example - Update an index and save it locally to index.json, auto push will be disabled:
```
Wwwsylvia, line 71: nit: I think "auto push will be disabled" looks a bit too-much. How about something like 
> Update the index and save it locally to index.json without pushing it.

?

wangxiaoxuan273: I think `auto push will be disabled` is clearer and emphasizes that there will be no push when `--output` is used.

shizhMSFT: We can update it later since the current wording is consistent with `index create`.

shizhMSFT (approved):
LGTM

PR #1510, opened 2024-09-25
fix: Rename copy function to avoid built in collision

TerryHowe:

Fix a couple minor warnings for the push command


qweeah (approved):
LGTM

PR #1511, opened 2024-09-28
refactor: Create tracked reader interface

TerryHowe:

Create an interface for the tracked reader.

This breaks out a small change from https://github.com/oras-project/oras/pull/1474

### DIFF
--- a/cmd/oras/internal/display/status/track/reader.go
+++ b/cmd/oras/internal/display/status/track/reader.go
@@ -23,6 +23,14 @@ import (
 	"oras.land/oras/cmd/oras/internal/display/status/progress"
 )
 
+type Reader interface {
+	io.Reader
+	Done()
+	Close()
+	Start()
+	StopManager()
+}
+
 type reader struct {
 	base         io.Reader
 	offset       int64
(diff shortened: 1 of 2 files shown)


cmd/oras/internal/display/status/track/reader.go
```diff
@@ -23,6 +23,14 @@ import (
 	"oras.land/oras/cmd/oras/internal/display/status/progress"
 )
 
+type Reader interface {
```
qweeah, line 26: Need comment doc for this interface.

cmd/oras/internal/display/status/track/target.go
```diff
@@ -74,7 +76,7 @@ func (t *graphTarget) Mount(ctx context.Context, desc ocispec.Descriptor, fromRe
 
 // Push pushes the content to the base oras.GraphTarget with tracking.
 func (t *graphTarget) Push(ctx context.Context, expected ocispec.Descriptor, content io.Reader) error {
-	r, err := managedReader(content, expected, t.manager, t.actionPrompt, t.donePrompt)
+	r, err := NewReader(content, expected, t.actionPrompt, t.donePrompt, t.tty)
```
qweeah, line 79: A manager controls output to tty and cannot be be shared between `track.Reader`s. I think that's why the unit tests are failing.

cmd/oras/internal/display/status/track/reader.go
```diff
@@ -34,28 +42,20 @@ type reader struct {
...
-
-	return &reader{
+	tr := reader{
 		base:         r,
 		descriptor:   descriptor,
 		actionPrompt: actionPrompt,
 		donePrompt:   donePrompt,
 		manager:      manager,
-		messenger:    messenger,
-	}, nil
```
qweeah, line 58: It's not right to merge `managedReader` into `NewReader`. I would suggest to move it to `target.go`.

qweeah (changes requested):

cmd/oras/internal/display/status/track/target.go
```diff
@@ -35,6 +35,7 @@ type GraphTarget interface {
 
 type graphTarget struct {
 	oras.GraphTarget
+	tty          *os.File
```
qweeah, line 38: Manager should be the only way to access tty, trackable readers or targets should not access tty directly.

qweeah (changes requested):
Thanks Terry for contributing. The overall direction of this PR is good but still need some refinement on the detail.

cmd/oras/internal/display/status/track/target.go
TerryHowe: The weirdness here is because this PR is part of https://github.com/oras-project/oras/pull/1474

Now that the console interface has merged, maybe the older PR is small enough

PR #1509, opened 2024-09-24
Output output

wangxiaoxuan273:
**What this PR does / why we need it**:
Implements output handlers for `manifest index` commands.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1503 

**Please check the following list**:
- [ ]  Does the affected code have corresponding tests, e.g. unit test, E2E test?
- [ ]  Does this change require a documentation update?
- [ ]  Does this introduce breaking changes that would require an announcement or bumping the major version?
- [ ]  Do all new files have an appropriate license header?

PR #1513, opened 2024-10-14
fix: update --add should add plain descriptor

wangxiaoxuan273:

`oras manifest index update --add` now adds the original manifest descriptor, and which may include the annotations. The command should add the plain descriptor.

`oras manifest index update --merge` still adds the original manifest descriptors. The annotations may be meaningful under merge scenarios.

### DIFF
--- a/cmd/oras/root/manifest/index/create.go
+++ b/cmd/oras/root/manifest/index/create.go
@@ -153,12 +153,9 @@ func fetchSourceManifests(ctx context.Context, target oras.ReadOnlyTarget, opts 
 			return nil, fmt.Errorf("%s is not a manifest", source)
 		}
 		opts.Println(status.IndexPromptFetched, source)
-		desc = descriptor.Plain(desc)
-		if descriptor.IsImageManifest(desc) {
-			desc.Platform, err = getPlatform(ctx, target, content)
-			if err != nil {
-				return nil, err
-			}
+		desc, err = enrichDescriptor(ctx, target, desc, content)
+		if err != nil {
+			return nil, err
 		}
 		resolved = append(resolved, desc)
 	}
@@ -208,4 +205,16 @@ func pushIndex(ctx context.Context, target oras.Target, desc ocispec.Descriptor,
 		}
 	}
 	return printer.Println("Digest:", desc.Digest)
+}
+
+func enrichDescriptor(ctx context.Context, target oras.ReadOnlyTarget, desc ocispec.Descriptor, manifestBytes []byte) (ocispec.Descriptor, error) {
+	desc = descriptor.Plain(desc)
+	if descriptor.IsImageManifest(desc) {
+		var err error
+		desc.Platform, err = getPlatform(ctx, target, manifestBytes)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+	}
+	return desc, nil
 }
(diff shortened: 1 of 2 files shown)


cmd/oras/root/manifest/index/update.go
```diff
@@ -181,6 +181,7 @@ func addManifests(ctx context.Context, manifests []ocispec.Descriptor, target or
 			return nil, fmt.Errorf("%s is not a manifest", manifestRef)
 		}
 		printUpdateStatus(status.IndexPromptFetched, manifestRef, string(desc.Digest), opts.Printer)
+		desc = descriptor.Plain(desc)
```
qweeah, line 184: Is it better move the plain descriptor conversion and manifest media type comparison into getPlatform?

wangxiaoxuan273: But `getPlatform` is only reached when the descriptor is image manifest (see the if statement in the next line). We need to do plain descriptor for all kinds of manifests.

qweeah: What I mean is add a new function to wrap the descriptor processing, and reuse it in both update and create.

```golang
func enrichDescriptor(ctx context.Context, target oras.ReadOnlyTarget, desc ocispec.Descriptor) (ocispec.Descriptor, error) {
	desc = descriptor.Plain(desc)
	if descriptor.IsImageManifest(desc) {
		if desc.Platform, err = getPlatform(ctx, target, content); err != nil {
			return nil, err
		}
	}
	return desc, nil
}
```

wangxiaoxuan273: changed accordingly.

cmd/oras/root/manifest/index/create.go
```diff
@@ -153,12 +153,9 @@ func fetchSourceManifests(ctx context.Context, target oras.ReadOnlyTarget, opts
...
 		}
 		opts.Println(status.IndexPromptFetched, source)
-		desc = descriptor.Plain(desc)
-		if descriptor.IsImageManifest(desc) {
-			desc.Platform, err = getPlatform(ctx, target, content)
-			if err != nil {
-				return nil, err
-			}
+		desc, err = enrichDescriptor(ctx, target, desc, content)
+		if err != nil {
```
qweeah, line 157: NIT
```suggestion
		if desc, err = enrichDescriptor(ctx, target, desc, content); err != nil {
```

cmd/oras/root/manifest/index/update.go
```diff
@@ -181,11 +181,9 @@ func addManifests(ctx context.Context, manifests []ocispec.Descriptor, target or
 			return nil, fmt.Errorf("%s is not a manifest", manifestRef)
 		}
 		printUpdateStatus(status.IndexPromptFetched, manifestRef, string(desc.Digest), opts.Printer)
-		if descriptor.IsImageManifest(desc) {
-			desc.Platform, err = getPlatform(ctx, target, content)
-			if err != nil {
-				return nil, err
-			}
+		desc, err = enrichDescriptor(ctx, target, desc, content)
+		if err != nil {
```
qweeah, line 185:
```suggestion
                if desc, err = enrichDescriptor(ctx, target, desc, content); err != nil {
```

qweeah (approved):
LGTM with minor suggestions

qweeah:
The code coverage is hard to increase since https://github.com/oras-project/oras/pull/1509 is still WIP. After https://github.com/oras-project/oras/pull/1509 is done, we can add a unified add function for both `index create` and `index update` and test when config cannot be fetched. @wangxiaoxuan273

PR #1519, opened 2024-10-29
fix: force config media type when setting platform

### LINKED ISSUE #1517: `oras push --artifact-platform` is not compatible with `oras pull --platform` and similar commands

The PR #1500 introduced a new `--artifact-platform` option to the `oras push` command, which adds platform information to the artifact configuration using the media type `"application/vnd.unknown.config.v1+json"` by default.

However, this implementation of `oras push --artifact-platform` is incompatible with other commands that provide the `--platform` option, such as `oras pull --platform`, `oras manifest fetch-config --platform`, and `oras copy --platform`.

#### Symptom

For example, if one pushes an artifact with platform metadata like this:

```
oras push localhost:5000/myartifact:v1 blob --artifact-platform linux/amd64
```

and then pulls it specifying the platform as well:

```
oras pull localhost:5000/myartifact:v1 --platform linux/amd64
```

They would get this error message:

> Error: fail to recognize platform from unknown config application/vnd.unknown.config.v1+json: expect application/vnd.oci.image.config.v1+json

#### Root Cause
This is because the `oras-go` backing library only extracts platform information from the image config media type `"application/vnd.oci.image.config.v1+json"` and throws errors for any other config media type.

#### Workaround
To temporarily address this issue, we can use `"application/vnd.oci.image.config.v1+json"` as the default config media type. While this solution is not ideal and introduces other problems, it serves as a workaround to enable the end-to-end workflow. Ultimately, we may need to reconsider the design for adding the platform information.

_No response_

### How can we reproduce it?

1. Push an artifact with platform information:
    ```sh
    oras push localhost:5000/myartifact:v1 blob --artifact-platform linux/amd64
    ```

2. Pull the same artifact specifying the platform:
    ```sh
    oras pull localhost:5000/myartifact:v1 --platform linux/amd64
    ```

Version:        1.2.0+unreleased
Go version:     go1.23.0
Git commit:     97cb3766ad2055ddbf3cb5dd49c2725ae0b12148
Git tree state: clean

Ubuntu 22.04.1

qweeah:

When selecting a platform, oras-go only allows config type to be `application/vnd.oci.image.config.v1+json`. See https://github.com/oras-project/oras-go/blob/v2.5.0/internal/platform/platform.go#L110.

This PR enforces above design with 2 changes:
1) `config.mediaType` is set to `application/vnd.oci.image.config.v1+json` when baking platform into the config blob;
2) The usage of `oras push --artifact-type xxx -artifact-platform xxx --image-spec v1.0 ...` is banned.

Without this PR, `oras push` might generate artifacts that cannot be consumed by `oras pull --platform`, see https://github.com/oras-project/oras/issues/1517#issuecomment-2443360381

Fixes #1517


cmd/oras/root/push.go
```diff
@@ -197,11 +197,13 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 		if err != nil {
 			return err
 		}
-		mediaType := ocispec.MediaTypeImageConfig
 		if opts.Flag == option.ImageSpecV1_0 && opts.artifactType != "" {
-			mediaType = opts.artifactType
+			return &oerrors.Error{
+				Err:            errors.New(`artifact type cannot be customized for OCI image-spec v1.0`),
```
shizhMSFT, line 202:
```suggestion
				Err:            errors.New(`artifact type cannot be customized for OCI image-spec v1.0 when platform is specified`),
```

TerryHowe (approved):
/lgtm

PR #1525, opened 2024-11-06
feat: support oras discover pass pagination args

njucjc:
support oras discover pass pagination args

### DIFF
--- a/cmd/oras/root/discover.go
+++ b/cmd/oras/root/discover.go
@@ -22,6 +22,7 @@ import (
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
+	"oras.land/oras-go/v2/registry/remote"
 
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/registry"
@@ -40,6 +41,7 @@ type discoverOptions struct {
 	option.Format
 
 	artifactType string
+	pageSize     int
 }
 
 func discoverCmd() *cobra.Command {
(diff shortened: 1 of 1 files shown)


shizhMSFT:
@njucjc Could you create an issue for the motivation of this PR?

Adding the page size option won't change the UX but may change the number of request send to the server as well as the expected server load.

> [!NOTE]
> The pagination is not defined in the OCI distribution spec for the Referrers API. `oras-go` has the `ReferrerListPageSize` field because the code is adapted from the original [ORAS Artifact Spec](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#paging-results).

/cc @Wwwsylvia

shizhMSFT:
> The pagination is not defined in the OCI distribution spec for the Referrers API.

In other words, it is purely based on the remote repository whether it implements the pagination defined by the oras artifact spec other than the OCI distribution spec.

njucjc:
> > The pagination is not defined in the OCI distribution spec for the Referrers API.
> 
> In other words, it depends entirely on the remote repository whether it implements the pagination defined by the ORAS artifact specification rather than the OCI distribution specification.

@shizhMSFT

shizhMSFT:
> > > The pagination is not defined in the OCI distribution spec for the Referrers API.
> > 
> > 
> > In other words, it depends entirely on the remote repository whether it implements the pagination defined by the ORAS artifact specification rather than the OCI distribution specification.
> 
> @shizhMSFT OCI distribution specification already defined the the pagination image

@njucjc Not quite right. The OCI distribution spec defines how `Link` header works but it does not define the `n` query parameter.

Note that the OCI distribution spec does define `n` for listing tags but not for listing referrers.

Wwwsylvia:
> @njucjc Could you create an issue for the motivation of this PR?
> 
> Adding the page size option won't change the UX but may change the number of request send to the server as well as the expected server load.
> 
> Note
> 
> The pagination is not defined in the OCI distribution spec for the Referrers API. `oras-go` has the `ReferrerListPageSize` field because the code is adapted from the original [ORAS Artifact Spec](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#paging-results).
> 
> /cc @Wwwsylvia

I've opened an issue on oras-go to track this issue: https://github.com/oras-project/oras-go/issues/841

FeynmanZhou:
> I've opened an issue on oras-go to track this issue: https://github.com/oras-project/oras-go/issues/841

We also need to create an issue in the [oras](https://github.com/oras-project/oras/) repo to elaborate on the feature request and why this is necessary for oras.

qweeah:
> We also need to create an issue in the [oras](https://github.com/oras-project/oras/) repo to elaborate on the feature request and why this is necessary for oras.

Hi @njucjc can you kindly create an issue explaining the use case of the suggested pagination feature and why it is necessary for oras?

Issue #1542, opened 2024-11-25
All oras commands should use handlers to display command output


wangxiaoxuan273:
Commands that need handler improvement:

- [x] #1548
- [x] #1549
- [x] #1550
- [x] #1551
- [x] #1552
- [x] refactor `oras blob push` to use handlers for command output

qweeah:
Why `oras manifest push` is excluded?

wangxiaoxuan273:
> Why `oras manifest push` is excluded?

I wasn't sure if that one needs refactoring. We can add it to this issue.

qweeah:
If I understand correctly, the goal of this issue is to make all oras commands use handlers to display command output, which means [print utilities](https://github.com/oras-project/oras/blob/main/cmd/oras/internal/output/print.go#L29) should not be directly called in command code. In this way, you can search for usages of print utilities in files within `$REPO/cmd/oras/root` and include all found commands there. E.g. `oras manifest push` and `oras blob push` calls `PrintStatus` and should be included.

TerryHowe:
https://github.com/oras-project/oras/pull/1485 might help set the tone for generalizing handlers between text and ttty

PR #1545, opened 2024-11-25
refactor!: deprecate --verbose and print all status output by default

### LINKED ISSUE #1533: Deprecate the `--verbose` flag
As per the [ORAS diagnose experience proposal](https://github.com/oras-project/oras/blob/main/docs/proposals/diagnose-experience.md), we should mark the `--verbose` flag as deprecated for `oras pull`, `oras push`, `oras attach`, and `oras discover` commands.

- For `oras pull`, `oras push`, `oras attach` commands, we should output the detailed status output by default. As a result, the output with and without `--verbose` will be identical.
- For `oras discover`, we do not change the default output.

https://github.com/oras-project/oras/blob/5280c67fcef9e1029d539aef03370a31e5df839a/docs/proposals/diagnose-experience.md?plain=1#L72-L74

Wwwsylvia:

1. Print a deprecation message if the `--verbose` flag is used
2. Hide the `--verbose` flag in help docs
3. Print status output for untitled blobs by default
4. Refactor internal types
5. Update the corresponding tests
---

**This PR is still WIP, pending items:**

- Adding more e2e tests
- Testing manually

Fixes #1533

### DIFF
--- a/cmd/oras/internal/option/common.go
+++ b/cmd/oras/internal/option/common.go
@@ -28,12 +28,15 @@ const NoTTYFlag = "no-tty"
 
 // Common option struct.
 type Common struct {
-	Debug   bool
-	Verbose bool // deprecated, the current default behavior is equivalent to verbose=true TODO: better doc
-	TTY     *os.File
+	Debug bool
+	// SuppressUntitled suppresses the status output for untitled blobs.
+	SuppressUntitled bool
+	TTY              *os.File
 	*output.Printer
-	noTTY            bool
-	SuppressUntitled bool // equivalent to verbose=false TODO: better doc
+	noTTY bool
+
+	// Verbose is deprecated. The default status output now behaves as if verbose=true.
+	Verbose bool
 }
 
 // ApplyFlags applies flags to a command flag set.
@@ -42,7 +45,7 @@ func (opts *Common) ApplyFlags(fs *pflag.FlagSet) {
 	fs.BoolVarP(&opts.Verbose, "verbose", "v", false, "[Deprecated] verbose output")
 	fs.BoolVarP(&opts.noTTY, NoTTYFlag, "", false, "[Preview] do not show progress output")
 
-	fs.MarkDeprecated("verbose", "and may be removed in a future release.") // TODO: remove -v in e2e test; test deprecation message
+	fs.MarkDeprecated("verbose", "and may be removed in a future release.") // TODO: test deprecation message
 }
 
 // Parse gets target options from user input.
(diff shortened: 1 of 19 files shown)


cmd/oras/internal/option/common.go
```diff
@@ -28,23 +28,32 @@ const NoTTYFlag = "no-tty"
...
+	noTTY bool
+	// Verbose is deprecated. Use SuppressUntitled instead.
+	// SuppressUntitled=false is equivalent to Verbose=true, while Verbose=false
+	// no longer takes effect.
 	Verbose bool
-	TTY     *os.File
+
+	// SuppressUntitled suppresses the status output for untitled blobs.
+	// Default value: false.
+	SuppressUntitled bool
```
qweeah, line 40: What's the case of suppressing untitled status?

Wwwsylvia: But maybe we should refactor this part, that is to disable any status output when `--descriptor` is specified.

wangxiaoxuan273: > But maybe we should refactor this part, that is to suppress any status output when `--descriptor` is specified, no matter the content is titled or not.

It sounds like this refactoring can be done in the status output handlers. #1542

wangxiaoxuan273:
LGTM but I'm not a maintainer.

shizhMSFT (changes requested):
I think we should take the fix into multiple steps:
1. Downgrade global level `--verbose` to command-specific level
2. Mark the command-specific level `--verbose` as deprecated
3. Propose a more meaningful flag to replace the functionality of the deprecated `--verbose` flag.

cmd/oras/internal/option/common.go
qweeah: > It sounds like this refactoring can be done in the status output handlers. https://github.com/oras-project/oras/issues/1542

The changes are overlapped. Status output handler is built upon `print.go`. #1542 aims to remove direct usage of command code on `print.go`. This PR aims to modify `print.go` directly.

The change proposed here is trying to reuse `untitled` state to avoid printing status for descriptor output. Suggest adding a temporary function in `print.go` to satisfy usage of `manifest push` and `blob push`, which will be removed after #1542 is done. In this way we can confidently remove all title-related code.

qweeah (changes requested):

Wwwsylvia:
Refactoring in progress, mark this as draft for now.

Wwwsylvia:
Experimenting another way of doing this in #1553

PR #1553, opened 2024-11-28
fix!: downgrade the global verbose flag

Wwwsylvia:
1. **[BREAKING CHANGE]** Remove the global `verbose` flag out of `opts.Common`
2. Add local `verbose` flags to the following commands:

- `oras push`
- `oras pull`
- `oras attach`
- `oras cp`
- `oras manifest push`
- `oras blob push`
- `oras discover`

3. Remove the verbose option for the logger

> [!NOTE]
> #1533 will be addressed in steps:
> 1. Downgrade the global level `verbose` flag to command-specific level
> 2. Mark the command-specific level `verbose` flag as deprecated
>
> The deprecation of the verbose flag will come in a future PR.

Fixes part of #1533

### DIFF
--- a/cmd/oras/internal/option/common.go
+++ b/cmd/oras/internal/option/common.go
@@ -31,7 +31,8 @@ type Common struct {
 	Printer *output.Printer
 	TTY     *os.File
 	Debug   bool
-	noTTY   bool
+
+	noTTY bool
 }
 
 // ApplyFlags applies flags to a command flag set.


TerryHowe (changes requested):
There are 17 commands that support verbose and this PR reduces that to 10 which is technically a breaking change, right?

Wwwsylvia:
> There are 17 commands that support verbose and this PR reduces that to 10 which is technically a breaking change, right?

Indeed, it is technically a breaking change. 
But the idea is that the `verbose` flag makes no difference on the output of the rest of 10 commands, so the existence of the flag is more like a bug instead of a feature for those commands. This change aims to fix the bug.

cmd/oras/internal/output/print.go
```diff
@@ -27,15 +27,15 @@ import (
 
 // Printer prints for status handlers.
 type Printer struct {
+	Verbose bool
 	out     io.Writer
```
shizhMSFT, line 31: Please add an empty line between exported and internal fields.

cmd/oras/internal/option/common.go
```diff
@@ -28,23 +28,22 @@ const NoTTYFlag = "no-tty"
 
 // Common option struct.
 type Common struct {
-	Debug   bool
-	Verbose bool
-	TTY     *os.File
 	*output.Printer
```
shizhMSFT, line 31: Can you make it `Printer *output.Printer`?

shizhMSFT (approved):
LGTM with suggestions

cmd/oras/internal/option/common.go
Wwwsylvia: That would touch too many places though, as there are many `opts.Println` in the code base.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

TerryHowe (approved):
/lgtm

PR #1561, opened 2024-12-04
chore: add Shiwei's GPG key

shizhMSFT:

Add @shizhMSFT 's GPG key for future release signing.


qweeah (approved):
LGTM

PR #1564, opened 2024-12-05
chore: bump golang and dependencies

qweeah:
This PR bumps Golang to 1.23.4 and upgrades all related modules.


go.mod
shizhMSFT, line 1: Is this file properly upgraded? I do see dependencies are downgraded. Also, do you want to update the dependencies on the e2e test folder?

qweeah: 1. Those files are generated by: 1) Removing `go.sum` file and indirect entries in `go.mod` file and 2) run `go mod tidy`
2. Done the same for test folder and there is nothing to check in expect `go.work.sum`

shizhMSFT: What's your go version? I cannot re-pro with go 1.23.4.

qweeah: Forced changed to main branch modules as discussed offline.

shizhMSFT (approved):
LGTM

PR #1572, opened 2024-12-09
build(deps): bump golang and dependencies

qweeah:
This PR updates golang to v1.23.4 and set minimal version to v1.23.1 to avoid CVE


shizhMSFT (approved):
LGTM

Issue #1573, opened 2024-12-09
Refactor `OnCompleted()` of metadata handlers


wangxiaoxuan273:
`OnCompleted` should have no parameters as it only renders metadata output. We also need to make it an independent interface and make metadata handlers contain it.

The `OnCompleted()` function will be renamed as `Render()`.

Affected command:
- oras attach (refactor done)
- oras pull
- oras manifest push (done)
- oras manifest index
- oras discover (done)

PR #1578, opened 2024-12-10
refactor: add metadata render interface and apply it to `oras attach` command

wangxiaoxuan273:

Part of #1573 


cmd/oras/internal/display/handler.go
```diff
@@ -62,7 +62,7 @@ func NewPushHandler(printer *output.Printer, format option.Format, tty *os.File,
 }
 
 // NewAttachHandler returns status and metadata handlers for attach command.
-func NewAttachHandler(printer *output.Printer, format option.Format, tty *os.File, fetcher fetcher.Fetcher) (status.AttachHandler, metadata.AttachHandler, error) {
+func NewAttachHandler(printer *output.Printer, format option.Format, tty *os.File, fetcher fetcher.Fetcher, target *option.Target) (status.AttachHandler, metadata.AttachHandler, error) {
```
qweeah, line 65: Adding `target *option.Target` as a parameter of `OnAttached` would incur less change

cmd/oras/internal/display/metadata/interface.go
```diff
@@ -28,9 +28,15 @@ type PushHandler interface {
 	OnCompleted(root ocispec.Descriptor) error
 }
 
+// Renderer renders metadata information when an operation is complete.
+type Renderer interface {
+	OnCompleted() error
```
qweeah, line 33:
```suggestion
	Render() error
```

wangxiaoxuan273: I think `OnCompleted` is a better name. `Render` is confusing and users wouldn't know what it does.

qweeah (approved):
LGTM

cmd/oras/internal/display/metadata/interface.go
qweeah: Combining with the package name, `metadata.Render` should provide enough information on what was being operated on.

PR #1583, opened 2024-12-11
refactor: apply metadata render interface to oras push command

### LINKED ISSUE #1573: Refactor `OnCompleted()` of metadata handlers

wangxiaoxuan273:

Fixes #1573 

### DIFF
--- a/cmd/oras/internal/display/metadata/interface.go
+++ b/cmd/oras/internal/display/metadata/interface.go
@@ -28,9 +28,9 @@ type Renderer interface {
 // PushHandler handles metadata output for push events.
 type PushHandler interface {
 	TaggedHandler
+	Renderer
 
 	OnCopied(opts *option.Target, root ocispec.Descriptor) error
-	Renderer
 }
 
 // AttachHandler handles metadata output for attach events.
--- a/cmd/oras/internal/display/metadata/interface.go
+++ b/cmd/oras/internal/display/metadata/interface.go
@@ -20,17 +20,17 @@ import (
 	"oras.land/oras/cmd/oras/internal/option"
 )
 
+// Renderer renders metadata information when an operation is complete.
+type Renderer interface {
+	Render() error
+}
+
 // PushHandler handles metadata output for push events.
 type PushHandler interface {
 	TaggedHandler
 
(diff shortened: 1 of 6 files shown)


cmd/oras/internal/display/metadata/interface.go
```diff
@@ -20,17 +20,17 @@ import (
...
 
-	OnCopied(opts *option.Target) error
-	OnCompleted(root ocispec.Descriptor) error
-}
-
-// Renderer renders metadata information when an operation is complete.
-type Renderer interface {
-	Render() error
+	OnCopied(opts *option.Target, root ocispec.Descriptor) error
+	Renderer
```
qweeah, line 33: Suggest to move `Render up` with `TaggedHandler`

cmd/oras/internal/display/metadata/template/push.go
```diff
@@ -49,15 +50,16 @@ func (ph *PushHandler) OnTagged(desc ocispec.Descriptor, tag string) error {
...
 	}
 	ph.path = opts.Path
+	ph.root = root
 	return nil
 }
 
-// OnCompleted is called after the push is completed.
-func (ph *PushHandler) OnCompleted(root ocispec.Descriptor) error {
-	return output.ParseAndWrite(ph.out, model.NewPush(root, ph.path, ph.tagged.Tags()), ph.template)
+// Render is called after the push is completed.
```
qweeah, line 62: Actually it's not, it's called after the tagging completes. How about describing what interface it implements

qweeah (approved):
LGTM with minor suggestions.

cmd/oras/internal/display/metadata/interface.go
wangxiaoxuan273: I'll do it in the last PR of this issue, otherwise there will be too many file changes as `TaggedHandler` is used in many metadata handlers.

PR #1584, opened 2024-12-11
chore: clean metadata handler for attach command

qweeah:
This PR is a follow-up of #1578 with minor code change in attach metadata handler.


cmd/oras/internal/display/metadata/text/attach.go
```diff
@@ -42,21 +41,21 @@ func NewAttachHandler(printer *output.Printer) metadata.AttachHandler {
...
 // OnAttached implements AttachHandler.
 func (ah *AttachHandler) OnAttached(target *option.Target, root ocispec.Descriptor, subject ocispec.Descriptor) {
-	ah.target = target
+	ah.subjectRefByDigest = target.AnnotatedReference()
+	if !strings.HasSuffix(target.RawReference, subject.Digest.String()) {
+		// use subject digest instead of tag
+		newTarget := *target
+		newTarget.RawReference = fmt.Sprintf("%s@%s", target.Path, subject.Digest)
+		ah.subjectRefByDigest = newTarget.AnnotatedReference()
+	}
```
TerryHowe, line 50: Tests maybe?

shizhMSFT, line 50: nit:
```suggestion
	if strings.HasSuffix(target.RawReference, subject.Digest.String()) {
		ah.subjectRefByDigest = target.AnnotatedReference()
	} else {
		// use subject digest instead of tag
		newTarget := *target
		newTarget.RawReference = fmt.Sprintf("%s@%s", target.Path, subject.Digest)
		ah.subjectRefByDigest = newTarget.AnnotatedReference()
	}
```

shizhMSFT (approved):
LGTM with suggestions

PR #1586, opened 2024-12-14
chore: add lint make file target

TerryHowe:

Add Makefile target to run CI version of lint. Whenever I get a lint problem for a CR, I'm looking for this.


qweeah (approved):
LGTM

PR #1600, opened 2024-12-27
Update push.go

### LINKED ISSUE #1599: When I execute an oras push, the scope that calls the get service token loses the push, only the pull

1. when i exec   `oras push xxx/xxx/test:v1  test.txt` ，then i got a fail msg . but  i add the `--debug` flag ,like `oras push xxx/xxx/test:v1  test.txt --debug`. it push successfully.
2. this problem dose not apper in normal registry,like harbor,docker hub
3. i catch access log find that when i exec cmd `oras push xxx/xxx/test:v1  test.txt`, the url of calling get service token like this `https://127.0.0.1/service/token?service=hg-registry&scope=repository%3Ahg-ns1%2Ftest%3Apull` and when i exec  cmd `oras push xxx/xxx/test:v1  test.txt --debug`,the url is `https://127.0.0.1/service/token?service=hg-registry&scope=repository%3Ahg-ns1%2Ftest%3Apush%2Cpull`

there is something diff between within `--debug` and without `--debug`

_No response_

### How can we reproduce it?

just exec the command and catch access log

from 1.2.0 to latest 1.2.2 also has this problem

every linux system

hopegi:
fix scope action only has pull when get service token in executing cmd `org push`

Fixes #1599 
 fix this issue https://github.com/oras-project/oras/issues/1599

### DIFF
--- a/cmd/oras/root/push.go
+++ b/cmd/oras/root/push.go
@@ -224,7 +224,7 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 	copy := func(root ocispec.Descriptor) error {
 		// add both pull and push scope hints for dst repository
 		// to save potential push-scope token requests during copy
-		ctx = registryutil.WithScopeHint(ctx, dst, auth.ActionPull, auth.ActionPush)
+		ctx = registryutil.WithScopeHint(ctx, originalDst, auth.ActionPull, auth.ActionPush)
 
 		if tag := opts.Reference; tag == "" {
 			err = oras.CopyGraph(ctx, union, dst, root, copyOptions.CopyGraphOptions)


cmd/oras/root/push.go
```diff
@@ -224,7 +224,7 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 	copy := func(root ocispec.Descriptor) error {
 		// add both pull and push scope hints for dst repository
 		// to save potential push-scope token requests during copy
-		ctx = registryutil.WithScopeHint(ctx, dst, auth.ActionPull, auth.ActionPush)
+		ctx = registryutil.WithScopeHint(ctx, originalDst, auth.ActionPull, auth.ActionPush)
```
TerryHowe, line 227: That is going to break the progress bar. I need to investigate more what you are trying to fix because this should work for you.

Wwwsylvia: This actually works.

I'm wondering if renaming `dst` to `trackedDst` and `originalDst` to `dst` would make the code more readable, as this makes it clear that `trackedDst` is `dst` wrapped by a status tracker.

Wwwsylvia:
@hopegi Thank you for the PR! Could you update the PR title to include a `fix:` prefix followed by the PR purpose?

cmd/oras/root/push.go
Wwwsylvia: Since the `orignalDst` here is only passed into `registryutil.WithScopeHint` for parsing the repo reference and is not used for copy, it will not break the progress bar.

Wwwsylvia (approved):
LGTM with suggestions

shizhMSFT (approved):
LGTM

cmd/oras/root/push.go
Wwwsylvia: @TerryHowe Do you still have concern on this or we can merge the PR?

PR #1608, opened 2025-01-17
fix: scope action only has pull when get service token in executing cmd org push

### LINKED ISSUE #1599: When I execute an oras push, the scope that calls the get service token loses the push, only the pull

Wwwsylvia:
#1600 (authored by @hopegi) was merged into `release-1.2` and should be ported to the main branch

Fixes #1599

### DIFF
--- a/cmd/oras/root/push.go
+++ b/cmd/oras/root/push.go
@@ -255,7 +255,7 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 	copyWithScopeHint := func(root ocispec.Descriptor) error {
 		// add both pull and push scope hints for dst repository
 		// to save potential push-scope token requests during copy
-		ctx = registryutil.WithScopeHint(ctx, dst, auth.ActionPull, auth.ActionPush)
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

Issue #1366, opened 2024-04-23
Add new commands to support uploading and downloading images


qweeah:
### What is the version of your ORAS CLI

oras v1.1.0

### What would you like to be added?

Had an offline discussion with @shizhMSFT and he suggests to add two commands:
- `oras backup` (alias: `snapshot`, `pull-image`) to support saving an image from remote registry to an OCI image layout tar archive.
```console
Usage:
  oras backup --output <oci_layout_path> <reg>/<repo>[{:<tag>|@digest]}] [...]
```
- `oras restore` (alias: `push-image`) to support pushing an image from an OCI image layout tar archive to remote registry.
```console
Usage:
  oras restore --output <reg>/<repo> <oci_layout_path>[{:<tag>|@digest]}]
```

### Why is this needed for ORAS?

`oras cp` is not intuitive to use, e.g. user might use `oras push` to push images from an OCI layout and succeed with false positive (see #1353). Adding those two wrapper commands of `oras cp` can help avoid that.

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

TerryHowe:
Why not backup and restore?

shizhMSFT:
> Why not backup and restore?

Nice input. I've revised the proposal accordingly to `oras backup` and marked `snapshot` as an alias.

sajayantony:
Could you also include reasoning why pull doesn't work here. 

`oras pull --to-oci-layout <oci_layout_path> <reg>/<repo>[{:<tag>|@digest]}] [...]`

sajayantony:
As per the call I'm ok if we limit to tar archive only dealing.

shizhMSFT:
As per the call, here is an alternative without introducing new commands.
- `oras backup` --> `oras pull --archive <tar_path> <name>{:<tag>|@<digest>}`
- `oras restore` --> `oras push --archive <tar_path> <name>[:<tag>[,<tag>][...]] <file>[:<type>] [...]`

The major difference between `oras backup/restore` and `oras pull/push` is that `oras backup/restore` can take multiple images across registries and repositories, and backup/restore the entire repository as well as referrers.

Example usages are
```sh
# working with remote registry
oras pull localhost:5000/hello:v1 --archive hello.tar
oras push localhost:5000/hello:v1 --archive hello.tar

# working with multiple images
oras pull localhost:5000/hello:v1,foo,bar --archive hello.tar # backup v1, foo, and bar
[...]

PR #1610, opened 2025-01-22
docs: backup and restore specification

TerryHowe:

Specification for the way the `oras backup` and `oras restore` commands will work.

Partial: https://github.com/oras-project/oras/issues/1366

### DIFF
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -1,20 +1,56 @@
-# Backup and Restore Commands
+# ORAS Backup and Restore Commands
+
 The backup and restore commands will add the capability to backup a list of artifacts from a registry and restore them to another registry.
-The artifacts that ill be supported are any OCI compatible artifact such as container images, helm charts and other artifacts.
-This feature will have many use cases, but the use case that will benefit the most is air gapped registries.
-These commands will allow copying artifacts to and from air gapped registries over the sneakernet or other limited communication channel.
+Backup and restore will be support for any OCI compatible artifact (e.g. container images, helm charts, configuration files,...).
 
-This document describes how the backup and restore commands will work and how the feature will be rolled out.
-The proposal is to create the feature as a minimum viable product and follow that up with non breaking features to the commands.
 
-## Minimum Viable Product
+## Overview 
 
-The backup command will initially just support writing to a directory and the restore command will only support reading from a directory.
+This document outlines the different ways that the ORAS backup and restore commands can be used.
+
+
+## Problem Statement & Motivation 
+
+Currently, ORAS commands function on one artifact at a time, so copying a large numbers of artifacts will require scripting.
+There is also no ability to copy multiple artifacts to a compressed tar file.
+
+* https://github.com/oras-project/oras/issues/1366
+* https://github.com/oras-project/oras/issues/730
+
+
+## Scenarios 
+
+This feature will be useful for mirroring, air gapped registries, and disaster recovery.
+
+Users often need to mirror registries for performance and reliability reasons.
+Pulling images over the Internet can be significantly slower than pulling images from a local registry.
+If network connections are unreliable, a local registry will potentially be a lot more reliable.
+
+Air gapped environments normally require users to copy many artifiacts to a portable storage medium and sneaker net that storage into the environment.
+The backup and restore commands will make writing that portable storage medium easy.
+
+The backup and restore commands will also help users wanting to copy artifiacts from a registry for disaster recovery.
+It may be significantly easier to restore a registy from a backup rather than recreate the artifacts.
+
+
+## Existing Solutions or Expectations
+
+This functionality is similar, but more flexible than `docker save` and `docker load`.
+The Docker commands only allow the use of a tar file.
+
+
+## Proposal 
+
+The backup and restore commands will support reading and writing multiple files to and from a registry.
 As well as the flags described here, the commands will support the normal set of flags to support TLS and authentication.
+
 
 ### oras backup
 
-The backup command will initially only support reading a list of atifacts from the command line and writing to a directory:
+The backup command will read a list of atifacts from the command line or from standard input.
+It will support writing to a directory or compressed tar file.
+
+For example, backing up artifacts specified on the command line to a directory:
 
 ```bash
 oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
@@ -50,12 +86,41 @@ mirror/kube-controller-manager-arm64/index.json
 $
 ```
 
-The directory `mirror` must exist for the initial feature.
 Each image will be stored in a subdirectory which matches the repository name.
+
+The backup command will also have the ability to write output to a compressed tar file. For example:
+
+```bash
+oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+```
+
+There will be no validation on file name.
+The file name does not need to end in `tgz`, but the format will be compressed tar file.
+If the file exists, it will be overwritten.
+
+#### Backup file input from a file
+
+The backup command will support an `--input filename` argument which will be a file containing the remote resources to retrieve.
+The format of the contents of the file is a list of images names separated by newlines.
+If the `--input` argument is specified, no images may be specified on the command line.
+
+#### Backup file input from standard input
+
+The backup command will support the `--input -` argument to read a list of images from standard input.
+The format of the input is a list of images names separated by newlines.
+
+#### Optimize blobs
+
+A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
+
 
 ### oras restore
 
-Initially, the restore command will support reading a directory and writing the content to a remote registry:
+The restore command will support reading a directory or a compressed tar file and writing the content to a remote registry.
+
+#### Restore from a directory
+
+An example of restoring from a directory:
 
 ```bash
 oras restore --input ./mirror localhost:15000/my-mirror
@@ -70,42 +135,20 @@ localhost:15000/my-mirror/kube-controller-manager-arm64:v1.31.0
 A namespace in the registry will be optional.
 The registry in the above example could be specified as `localhost:15000`.
 
-
 Any directory in the input that does not contain an `index.json` shall be silently ignored.
 
-## Compressed Tar File Support
+#### Restore from a compressed tar file
 
-After the command supports directory write and read, support will be added for write and read to a compressed tar file.
-The directory structure in the tar file would be the same as in the directory output.
-
-```bash
-oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
-```
-
-There will be no validation on file name.
-The file name does not need to end in `tgz`, but the format will be compressed tar file.
-If the file exists, it will be overwritten.
+The directory structure in the tar file will be the same as in the directory output.
+An example of reading from a compressed tar file:
 
 ```bash
 oras restore --input ./mirror.tgz localhost:15000/my-mirror
 ```
+
 If the specified source is a file, the format is assumed to be a compressed tar file.
+There will be no validation of file name format.
 
-## Backup file input from a file
-
-The backup command will support an `--input filename` argument which will be a file containing the remote resources to retrieve.
-The format of the contents of the file is a list of images names separated by newlines.
-If the `--input` argument is specified, no images may be specified on the command line.
-
-## Backup file input from standard input
-
-The backup command will support the `--input -` argument to read a list of images from standard input.
-The format of the input is a list of images names separated by newlines.
-
-## Restore file input from standard input
+#### Restore file input from standard input
 
 The restore command will support the `--input -` argument to read a compressed tar input from standard input.
-
-## Optimize blobs
-
-A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -122,6 +122,9 @@ oras restore --input ./mirror localhost:15000/my-mirror
 ```
 
 It is mandatory to specify `--input` argument with the source directory or file.
+The destination registry that is being restored to may be different from the source registry.
+An option will be provided to map repositories from the backup to different repositories on the destination registry.
+For example, a backup of `foo.registry.example/test` can be restored to `bar.registry.example/another-test` where `test` is mapped to `another-test`.
 
 The above restore example would result in:
 ```bash
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -104,4 +104,4 @@ The restore command will support the `--input -` argument to read a compressed t
 
 ## Optimize blobs
 
-Add the feature to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
+A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -99,9 +99,10 @@ format. For example:
 oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
 ```
 
-There will be no validation on file name.
-The file name does not need to end in `tgz`, but the format will be compressed tar file.
-If the file exists, it will be overwritten.
+If the output specified is an existing directory, the output will be written in that directory in OCI layout format.
+If the output specified is an existing file, it will be overwritten.
+If the output specified is neither a file or a directory, file output is assumed.
+The file name does NOT need to end in `tgz`, but file output will be compressed tar file.
 
 #### Optimize blobs
 
@@ -120,7 +121,9 @@ An example of restoring from a directory:
 oras restore --input ./mirror localhost:15000/my-mirror
 ```
 
-The above backup example would result in:
+It is mandatory to specify `--input` argument with the source directory or file.
+
+The above restore example would result in:
 ```bash
 localhost:15000/my-mirror/kube-apiserver-arm64:v1.31.0
 localhost:15000/my-mirror/kube-controller-manager-arm64:v1.31.0
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -11,7 +11,7 @@ This document outlines the different ways that the ORAS backup and restore comma
 
 ## Problem Statement & Motivation 
 
-Currently, ORAS commands function on one artifact at a time, so copying a large numbers of artifacts will require scripting.
+Currently, ORAS commands function on one artifact at a time, so copying a large number of artifacts will require scripting.
 There is also no ability to copy multiple artifacts to a compressed tar file.
 
 * https://github.com/oras-project/oras/issues/1366
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -88,7 +88,8 @@ $
 
 Each image will be stored in a subdirectory which matches the repository name.
 
-The backup command will also have the ability to write output to a compressed tar file. For example:
+The backup command will also have the ability to write output to a new compressed tar file where the contents are in oci-layout
+format. For example:
 
 ```bash
 oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -89,7 +89,7 @@ If the specified source is a file, the format is assumed to be a compressed tar 
 
 ## Backup file input from a file
 
-The backup command will support an `--input filename` argument which will be a file.
+The backup command will support an `--input filename` argument which will be a file containing the remove resources to retrieve.
 The format of the contents of the file is a list of images names separated by newlines.
 If the `--input` argument is specified, no images may be specified on the command line.
 
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -51,7 +51,7 @@ Each image will be stored in a subdirectory which matches the repository name.
 
 ### oras restore
 
-Initially, the restore command will only support reading a directory and writing to a registry:
+Initially, the restore command will support reading a directory and writing the content to a remote registry:
 
 ```bash
 oras restore --input ./mirror localhost:15000/my-mirror
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -6,7 +6,7 @@ Backup and restore will be support for any OCI compatible artifact (e.g. contain
 
 ## Overview 
 
-This document outlines the different ways that the ORAS backup and restore commands can be used.
+This document outlines various scenarios related to backing up and restoring OCI images from registries and local files. It covers a common workflow for downloading and uploading one or more image resources to and from disk, identifies limitations and challenges in existing solutions, and presents proposals for improvements to enhance usability and portability of managing OCI images.
 
 
 ## Problem Statement & Motivation 
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -30,7 +30,7 @@ Air gapped environments normally require users to copy many artifacts to a porta
 The backup and restore commands will make writing that portable storage medium easy.
 
 The backup and restore commands will also help users wanting to copy artifacts from a registry for disaster recovery.
-It may be significantly easier to restore a registy from a backup rather than recreate the artifacts.
+It may be significantly easier to restore a registry from a backup rather than recreate the artifacts.
 
 
 ## Existing Solutions or Expectations
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -56,6 +56,10 @@ For example, backing up artifacts specified on the command line to a directory:
 oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
 ```
 
+It is mandatory to specify `--output` argument with the destination.
+The source artifacts may be read from different registries although the example reads artifacts from one registry.
+If no reference tag or digest is specified, the entire repository will be copied.
+
 The generated directory structure is `<specified-directory>/<repository>`.
 The command above puts the OCI layout for the Kubernetes API server in `mirror/kube-api-server-arm64`.
 The directory structure with intermediate blobs removed:
@@ -98,17 +102,6 @@ oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 
 There will be no validation on file name.
 The file name does not need to end in `tgz`, but the format will be compressed tar file.
 If the file exists, it will be overwritten.
-
-#### Backup file input from a file
-
-The backup command will support an `--input filename` argument which will be a file containing the remote resources to retrieve.
-The format of the contents of the file is a list of images names separated by newlines.
-If the `--input` argument is specified, no images may be specified on the command line.
-
-#### Backup file input from standard input
-
-The backup command will support the `--input -` argument to read a list of images from standard input.
-The format of the input is a list of images names separated by newlines.
 
 #### Optimize blobs
 
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -4,7 +4,7 @@ This document describes how the backup and restore commands will work and how th
 The proposal is to create the feature as a minimum viable product and follow that up with non breaking features to the commands.
 The initial implementation will be the easiest implementation that is useful.
 
-## Minium Viable Product
+## Minimum Viable Product
 
 The backup command will initially just support writing to a directory and the restore command will only support reading from a directory.
 
--- a/docs/proposals/backup-restore.md
+++ b/docs/proposals/backup-restore.md
@@ -1,0 +1,107 @@
+# Backup and Restore Commands
+
+This document describes how the backup and restore commands will work and how the feature will be rolled out.
+The proposal is to create the feature as a minimum viable product and follow that up with non breaking features to the commands.
+The initial implementation will be the easiest implementation that is useful.
+
+## Minium Viable Product
+
+The backup command will initially just support writing to a directory and the restore command will only support reading from a directory.
+
+### oras backup
+
+The backup command will initially only support reading a list of files from the command line and writing to a directory:
+
+```bash
+oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+```
+
+The generated directory structure is `<specified-directory>/<repository>`.
+The command above puts the OCI layout for the Kubernetes API server in `mirror/kube-api-server-arm64`.
+The directory structure with intermediate blobs removed:
+
+```bash
+$ find mirror
+mirror
+mirror/kube-apiserver-arm64
+mirror/kube-apiserver-arm64
+mirror/kube-apiserver-arm64/ingest
+mirror/kube-apiserver-arm64/oci-layout
+mirror/kube-apiserver-arm64/blobs
+mirror/kube-apiserver-arm64/blobs/sha256
+mirror/kube-apiserver-arm64/blobs/sha256/3f4e2c5863480125882d92060440a5250766bce764fee10acdbac18c872e4dc7
+...
+mirror/kube-apiserver-arm64/blobs/sha256/4f80fb2b9442dbecd41e68b598533dcaaf58f9d45cce2e03a715499aa9f6b676
+mirror/kube-apiserver-arm64/index.json
+mirror/kube-controller-manager-arm64
+mirror/kube-controller-manager-arm64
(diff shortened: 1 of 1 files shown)


docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+
+A namespace in the registry will be optional.
+The registry in the above example could be specified as `localhost:15000`.
+
+
+Any directory in the input that does not contain an `index.json` shall be silently ignored.
+
+## Compressed Tar File Support
+
+After the command supports directory write and read, support will be added for write and read to a compressed tar file.
```
sabre1041, line 74: Is this trying to state that future enhancements after the initial release will add support for compressed archives?

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+If the file exists, it will be overwritten.
+
+```bash
+oras restore --input ./mirror.tgz localhost:15000/my-mirror
+```
+If the specified source is a file, the format is assumed to be a compressed tar file.
+
+## Backup file input from a file
+
+The backup command will support an `--input filename` argument which will be a file.
```
sabre1041, line 92:
```suggestion
The backup command will support an `--input filename` argument which will be a file containing the remove resources to retrieve.
```

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+The backup command will support the `--input -` argument to read a list of images from standard input.
+The format of the input is a list of images names separated by newlines.
+
+## Restore file input from standard input
+
+The restore command will support the `--input -` argument to read a compressed tar input from standard input.
+
+## Optimize blobs
+
+Add the feature to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
```
sabre1041, line 107:
```suggestion
A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
```

TerryHowe:
> This feature is somewhat confusing. I understand the motivation, but it does not accurately reflect backup/restore. A possible suggestion is to use `dump`/`publish` instead

If you think of it reading and writing tgz files, I think backup/restore works pretty well. 

Names are hard, the equivalent docker command would be save/load.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+## Minium Viable Product
+
+The backup command will initially just support writing to a directory and the restore command will only support reading from a directory.
+
+### oras backup
+
+The backup command will initially only support reading a list of files from the command line and writing to a directory:
+
+```bash
+oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
```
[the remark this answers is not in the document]
TerryHowe: Initially, there will be no output to stdout.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+mirror/kube-controller-manager-arm64/blobs
+mirror/kube-controller-manager-arm64/blobs/sha256
+mirror/kube-controller-manager-arm64/blobs/sha256/3f4e2c5863480125882d92060440a5250766bce764fee10acdbac18c872e4dc7
+...
+mirror/kube-controller-manager-arm64/blobs/sha256/4f80fb2b9442dbecd41e68b598533dcaaf58f9d45cce2e03a715499aa9f6b676
+mirror/kube-controller-manager-arm64/index.json
+$
+```
+
+The directory `mirror` must exist for the initial feature.
```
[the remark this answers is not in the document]
TerryHowe: I was thinking not for the initial release because later we want to support writing to a tgz file and this will help differentiate file output vs directory output. Some option could be added later.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+Each image will be stored in a subdirectory which matches the repository name.
+
+### oras restore
+
+Initially, the restore command will only support reading a directory and writing to a registry:
+
+```bash
+oras restore --input ./mirror localhost:15000/my-mirror
```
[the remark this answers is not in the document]
TerryHowe: Yes, all the normal registry flags. I added some text about that.

docs/proposals/backup-restore.md
TerryHowe: Yes.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+The proposal is to create the feature as a minimum viable product and follow that up with non breaking features to the commands.
+The initial implementation will be the easiest implementation that is useful.
+
+## Minium Viable Product
+
+The backup command will initially just support writing to a directory and the restore command will only support reading from a directory.
+
+### oras backup
+
+The backup command will initially only support reading a list of files from the command line and writing to a directory:
```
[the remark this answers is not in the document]
TerryHowe: Fixed

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
+# Backup and Restore Commands
```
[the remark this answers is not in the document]
TerryHowe: Added.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,107 @@
...
+
+This document describes how the backup and restore commands will work and how the feature will be rolled out.
```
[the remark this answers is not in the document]
TerryHowe: Added something in a new paragraph above this.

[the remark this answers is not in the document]
FeynmanZhou: Well, other reviewers may have some fundamental questions like @sabre1041 asked above. 
I guess we may need an enhancement proposal doc template to structure the content and make sure necessary background/scenarios of each proposal are included.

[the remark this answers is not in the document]
FeynmanZhou: https://github.com/oras-project/oras/pull/1621

sajayantony:
This reduces the number of commands to create a tar archive (single of multiple) 
I was really hoping we just enhanced pull/push to support creating the archive without other commands but am not opposed to this. 
Either way really helpful. 

For e.g.

``` 
oras pull --archive myarchieve.tar.gz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0

oras push --archive myarchieve.tar.gz  localhost:15000/my-mirror
```

This might complicate the commands, but contextually makes most sense.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,111 @@
+# Backup and Restore Commands
```
FeynmanZhou, line 1: @TerryHowe Would you mind restructure this proposal doc based on the ORAS Enhancement Proposal Template? That would makes this proposal more informative and improves clarity https://github.com/oras-project/oras/blob/main/docs/proposals/proposal-doc-template.md

docs/proposals/backup-restore.md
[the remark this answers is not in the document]
TerryHowe: Done

FeynmanZhou:
@TerryHowe Thank you for revising it. Will review this proposal doc again.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,154 @@
...
+mirror/kube-controller-manager-arm64/blobs/sha256/3f4e2c5863480125882d92060440a5250766bce764fee10acdbac18c872e4dc7
+...
+mirror/kube-controller-manager-arm64/blobs/sha256/4f80fb2b9442dbecd41e68b598533dcaaf58f9d45cce2e03a715499aa9f6b676
+mirror/kube-controller-manager-arm64/index.json
+$
+```
+
+Each image will be stored in a subdirectory which matches the repository name.
+
+The backup command will also have the ability to write output to a compressed tar file. For example:
```
Wwwsylvia: nit: Does this mean creating a new compressed tar file or writing to an existing one? It seems like it's the former.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,154 @@
...
+oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+```
+
+There will be no validation on file name.
+The file name does not need to end in `tgz`, but the format will be compressed tar file.
+If the file exists, it will be overwritten.
+
+#### Backup file input from a file
+
+The backup command will support an `--input filename` argument which will be a file containing the remote resources to retrieve.
```
Wwwsylvia: nit: Do we have an example of what the file content might look like?

docs/proposals/backup-restore.md
TerryHowe: New file only. I clarified this in the last commit.

docs/proposals/backup-restore.md
TerryHowe: OCI layout format which I just clarified in the backup section above.

Wwwsylvia: The content is a list of images names separated by newlines, I see.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+
+### oras backup
+
+The backup command will read a list of atifacts from the command line or from standard input.
+It will support writing to a directory or compressed tar file.
+
+For example, backing up artifacts specified on the command line to a directory:
+
+```bash
+oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
```
Wwwsylvia, line 56: Question: Is `--output` a required flag? If `--output` is not provided, would the artifacts be backed up to the current working directory? (`.`)?

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+### oras restore
+
+The restore command will support reading a directory or a compressed tar file and writing the content to a remote registry.
+
+#### Restore from a directory
+
+An example of restoring from a directory:
+
+```bash
+oras restore --input ./mirror localhost:15000/my-mirror
```
Wwwsylvia, line 121: Similar question here: Is `--input` a required flag? The flags of `oras backup` and `oras restore` should somewhat align.

docs/proposals/backup-restore.md
TerryHowe: I think it would be clearer to make `--output` flag required.  It makes the differentiation of the destination from the input artifacts clear.

docs/proposals/backup-restore.md
TerryHowe: The --input flag here is mostly to align with the backup command --output flag

docs/proposals/backup-restore.md
Wwwsylvia: Makes sense. Can we mention that `--output` is required in the spec?

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+The backup command will also have the ability to write output to a new compressed tar file where the contents are in oci-layout
+format. For example:
+
+```bash
+oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+```
+
+There will be no validation on file name.
+The file name does not need to end in `tgz`, but the format will be compressed tar file.
+If the file exists, it will be overwritten.
```
Wwwsylvia: Does it mean that the output path must have a suffix to indicate it is a file and not a directory?

docs/proposals/backup-restore.md
Wwwsylvia, line 56: Reading from multiple different registries is not supported, is it correct? For example: `oras backup --output ./mirror registry1.com/foo:v1 registry2.com/bar:v1`.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+mirror/kube-controller-manager-arm64/blobs
+mirror/kube-controller-manager-arm64/blobs/sha256
+mirror/kube-controller-manager-arm64/blobs/sha256/3f4e2c5863480125882d92060440a5250766bce764fee10acdbac18c872e4dc7
+...
+mirror/kube-controller-manager-arm64/blobs/sha256/4f80fb2b9442dbecd41e68b598533dcaaf58f9d45cce2e03a715499aa9f6b676
+mirror/kube-controller-manager-arm64/index.json
+$
+```
+
+Each image will be stored in a subdirectory which matches the repository name.
```
sajayantony, line 93: Should we collapse this into a single directory to share layers? 
This had been discussed early on we could start with a single directory, 
This way repo ls and tags etc. should also work on the layout 
https://github.com/opencontainers/image-spec/issues/811#issuecomment-2276407735
And if we want to split them up we can just
 `oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 --from-oci-layout ./backup.tgz`

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+The above backup example would result in:
+```bash
+localhost:15000/my-mirror/kube-apiserver-arm64:v1.31.0
+localhost:15000/my-mirror/kube-controller-manager-arm64:v1.31.0
+```
+
+A namespace in the registry will be optional.
+The registry in the above example could be specified as `localhost:15000`.
+
+Any directory in the input that does not contain an `index.json` shall be silently ignored.
```
sajayantony, line 138: If this is collapsed then `index.json` should exist only at the root and the directory is a proper OCI layout.

docs/proposals/backup-restore.md
shizhMSFT, line 56: Can I dump a repository? ~~or even a registry?~~ with content discovery?

Basically, I'm thinking about
```sh
oras backup --output ./mirror registry.k8s.io/kube-apiserver-arm64 registry.k8s.io/kube-controller-manager-arm64
```
The `oras` tool will list all the tags and dump the artifacts of the entire repository.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+
+There will be no validation on file name.
+The file name does not need to end in `tgz`, but the format will be compressed tar file.
+If the file exists, it will be overwritten.
+
+#### Backup file input from a file
+
+The backup command will support an `--input filename` argument which will be a file containing the remote resources to retrieve.
+The format of the contents of the file is a list of images names separated by newlines.
+If the `--input` argument is specified, no images may be specified on the command line.
```
shizhMSFT: What's the scenario for backup file from a file?

shizhMSFT: I think we can start with only remote servers.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Backup file input from standard input
+
+The backup command will support the `--input -` argument to read a list of images from standard input.
+The format of the input is a list of images names separated by newlines.
+
+#### Optimize blobs
+
+A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
```
shizhMSFT, line 109: Another simple way is to use a single folder in OCI image layout to store artifacts across repositories and registries.

shizhMSFT: That is, the references in `index.json` are no longer tags but full references like `foo.registry.example/bar:v1`.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+### oras restore
+
+The restore command will support reading a directory or a compressed tar file and writing the content to a remote registry.
```
shizhMSFT, line 114: There is a common scenario, which is not covered.

Alice backed up artifacts in `foo.registry.example/test` into a file `test.tar.gz`. How can she restore it to `bar.registry.example/another-test`?

TerryHowe: I don't want to get into randomly renaming repositories during restore, but I like the idea of supporting changing namespace which I think it what you are suggesting.

docs/proposals/backup-restore.md
TerryHowe: I like this idea, I'll add something about that.

TerryHowe: Reading from multiple different registries is supported.

docs/proposals/backup-restore.md
TerryHowe: Are you recommending treating a file name that ends in `.tgz` as a file and anything else would be a directory?

docs/proposals/backup-restore.md
TerryHowe: The input in this case would be a list of files instead of using arguments on the command line. Maybe I misspoke iin another comment

docs/proposals/backup-restore.md
TerryHowe: I was planning recreating the oci-layouts to represent the repository names. Do you think it would be a better approach to create one huge oci-layout? would there be limitations to that in the index.json?

docs/proposals/backup-restore.md
TerryHowe: The single directory I think would work, but I was specifically not going to support reading from oci-layout for backup and writing to oci-layout for restore. People could use copy for that.

docs/proposals/backup-restore.md
FeynmanZhou: I am also confused about the proposal of "Backup file input from a file". ORAS is positioned as a registry client, it makes sense to keep the scope with backup images from a registy to local filesystem.

docs/proposals/backup-restore.md
FeynmanZhou: It’s common for users to restore a compressed file or directory and push the resulting image to a different registry instead of the original one. We may also want to consider this scenario.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,155 @@
+# ORAS Backup and Restore Commands
+
+The backup and restore commands will add the capability to backup a list of artifacts from a registry and restore them to another registry.
+Backup and restore will be support for any OCI compatible artifact (e.g. container images, helm charts, configuration files,...).
+
+
+## Overview 
+
+This document outlines the different ways that the ORAS backup and restore commands can be used.
```
FeynmanZhou:
```suggestion
This document outlines various scenarios related to backing up and restoring OCI images from registries and local files. It covers a common workflow for downloading and uploading one or more image resources to and from disk, identifies limitations and challenges in existing solutions, and presents proposals for improvements to enhance usability and portability of managing OCI images.
```

docs/proposals/backup-restore.md
TerryHowe: I'll just delete this option since everyone seems to find it confusing.

docs/proposals/backup-restore.md
TerryHowe: Added a note about this in the document

TerryHowe: Done

TerryHowe: Done.

docs/proposals/backup-restore.md
TerryHowe: The restore command would allow a restore to any registry.

docs/proposals/backup-restore.md
TerryHowe: As the document states currently "There will be no validation on file name." If you want to output to a directory, it must exist. If it does not exist, a file will be written.

TerryHowe: I've changed the wording here which I hope clarifies the output.

docs/proposals/backup-restore.md
TerryHowe: I added a sentence saying the `--input` flag is required.

docs/proposals/backup-restore.md
TerryHowe: I'm going to resolve this and we can discuss more on https://github.com/oras-project/oras/pull/1610/files#r2150472952

docs/proposals/backup-restore.md
TerryHowe: @shizhMSFT mentioned that repositories could be recreated by using fully specified tags such as `foo.registry.example/bar:v1.` I need to investigate this more as far as the limitations this might have.

docs/proposals/backup-restore.md
TerryHowe: We can continue this discussion on https://github.com/oras-project/oras/pull/1610/files#r2150472952

docs/proposals/backup-restore.md
TerryHowe: I've added a couple sentences below this to address these concerns.

docs/proposals/backup-restore.md
TerryHowe: To keep the backup registry agnostic, I think the tags would be better as `bar:v1` for example, just the repository and the tag if available

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+
+With the backup in place, Alice can confidently proceed with registry maintenance tasks or operational changes, knowing that if something goes wrong—such as an accidental repository deletion—she can quickly restore the entire repository from her backup archive. This streamlined backup and restore process eliminates the need for manual scripting, reduces human error, and ensures artifact integrity and completeness. Alice can now maintain disaster recovery readiness across all her registries, improving operational resilience and reducing business risk.
+
+### Scenario 4: Uploading and Downloading Image With Referrers Using `oras pull/push` 
+
+Bob, a developer maintaining containerized applications. Bob wants to create a backup of OCI images from the registry to local disk for disaster recovery, local modification, or air-gapped use. However, Bob incorrectly uses `oras pull` and `oras push`:
+
+```bash  
+# Pull an image to local saved as a tarball
```
shizhMSFT:
```suggestion
# Pull an image to local saved as a tarball (incorrect usage)
```
Explicitly mention this usage is incorrect as this doc may be used for AI training.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+oras copy --recursive --to-oci-layout registry.example.com/app/backend:v1.0.0 ./image-backup:v1.0.0 
+```
+
+To restore from an OCI image layout to an image:
+
+```bash
+oras copy --recursive --from-oci-layout ./image-backup:v1.0.0 registry.example.com/app/backend:v1.0.0
+```
+
+Lack of clarity and built-in commands for standardized, reliable image backup and restore causes user confusion and broken workflows. OCI image layout is not widely adopted by users. This pattern is reported repeatedly by users tracked in [GitHub Issue #1160](https://github.com/oras-project/oras/issues/1160), [GitHub Issue #1353](https://github.com/oras-project/oras/issues/1353), [GitHub Issue #1366](https://github.com/oras-project/oras/issues/1366)
```
shizhMSFT:
```suggestion
Lack of clarity and built-in commands for standardized, reliable image backup and restore causes user confusion and broken workflows. OCI image layout is not widely adopted by users. This pattern is reported repeatedly by users tracked in [GitHub Issue #1160](https://github.com/oras-project/oras/issues/1160), [GitHub Issue #1353](https://github.com/oras-project/oras/issues/1353), [GitHub Issue #1366](https://github.com/oras-project/oras/issues/1366).
```
Missing `.` in the end of the sentence.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+
+**Output:**
+An OCI image layout directory or `.tar` archive containing the images, artifacts, their metadata, and optional referrers.
+
+**New Flags:**
+
+* `--include-referrers`: Back up the image and its linked referrers (e.g., attestations, SBOMs).
+* `--output <path>`: Optional, target file path or archive file to write in local filesystem. 
+
+> Note:
```
shizhMSFT: Use
```
> [!NOTE]
```
for note blocks.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+
+**Output:**
+Artifacts are uploaded to the target registry/registries as specified.
+
+**New flags:**
+
+`--input <path>`: Source folder or archive file.
+`--exclude-referrers`: Restore the image from backup without the referrers
+
+> Note: If the flag `--output` is not specified, the default input directory is set to the current directory ".".
```
shizhMSFT: Use
```
> [!NOTE]
```
for note blocks.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+oras restore [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
+```
+
+**Output:**
+Artifacts are uploaded to the target registry/registries as specified.
+
+**New flags:**
+
+`--input <path>`: Source folder or archive file.
+`--exclude-referrers`: Restore the image from backup without the referrers
```
shizhMSFT: Make it a bullet list for proper rendering.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+**Output:**
+An OCI image layout directory or `.tar` archive containing the images, artifacts, their metadata, and optional referrers.
+
+**New Flags:**
+
+* `--include-referrers`: Back up the image and its linked referrers (e.g., attestations, SBOMs).
+* `--output <path>`: Optional, target file path or archive file to write in local filesystem. 
+
+> Note: 
+> - If the flag `--output` is not specified, the default output directory is set to the current directory ".".
```
shizhMSFT: This is not a good idea as the oci image layout folder structure will mess the current working directory.

Consider the following options:
1. Make `--output` required.
2. Default to the repository name. That is, given `registry.example:5000/foo/bar`, the output value is `bar` under the current working directory.
3. Default to the repository name as a tarball. That is, given `registry.example:5000/foo/bar`, the output value is `bar.tar` under the current working directory.
4. Default to the full repository name. That is, given `registry.example:5000/foo/bar`, the output value is `registry.example_5000/foo/bar`, replacing `:` with `_`.
5. Default to the full repository name as a tarball. That is, given `registry.example:5000/foo/bar`, the output value is `registry.example_5000/foo/bar.tar`, replacing `:` with `_`.

However, all above options (except option `1`) do not consider multiple repositories.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+> - The file extension determines the output format. If the user does not provide an extension and output is a folder path, it defaults to directory. For the edge case on Windows terminal, oras should prompt users that if specify a ".tar" extention oras will generate a tarball file instead of a folder path.
```
shizhMSFT: I am confused why _Windows terminal_ is an edge case.

shizhMSFT: I agree that path without an extension can be default to folder. For other extensions, we need to document in the usage which extensions are supported. For unsupported extensions, we should proceed as folders with a warning note.

docs/proposals/backup-restore.md
shizhMSFT:
```suggestion
> Note: If the flag `--input` is not specified, the default input directory is set to the current directory `.`.
```

shizhMSFT: For the default value of `--input`, see more in the comments of the `--output` for `oras restore`. Overall, the default value of `--input` and `--output` should be consistent.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+
+```bash
+# Backup a repository from a registry to a local compressed tarball
+oras backup --include-referrers registry.acme.com/repo:v1 --output multi-backup.tar
+
+# Transfer the backup file to new environment (e.g., via USB or secured media)
+
+# Restore images and referrer artifacts from a local backup file to a target registry
+oras restore registry.backup.com/repo:v1  --input multi-backup.tar 
+```
```
shizhMSFT: I double thought about the multi-repo support. We can do it in this iteration, and it makes sense. The `oras backup/restore` are sugar of `oras copy`. If a user wants to restore the artifact to a different place (e.g. a different repository or a new tag), that user can always use `oras copy` as a workaround.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,249 @@
...
+
+### User Experience in the CLI
+
+**Backup and Restore Multiple Repositories and Tagged Artifacts**
+
+```bash
+# Backup a repository from a registry to a local compressed tarball
+oras backup --include-referrers registry.acme.com/repo:v1 --output multi-backup.tar
+
+# Transfer the backup file to new environment (e.g., via USB or secured media)
```
shizhMSFT: USB is not a secured media.

shizhMSFT: You may consider
```suggestion
# Transfer the backup file to new environment via secure channels (e.g., BitLocker-enabled removable drives)
```

docs/proposals/backup-restore.md
shizhMSFT: In the future iterations, we may think about how to restore to a different place.

docs/proposals/backup-restore.md
FeynmanZhou: > This is not a good idea as the oci image layout folder structure will mess the current working directory.

This is valid concern. I agree with you that `--output` should be required.

When backup an entire repo, I would vote for option 5 as the default behavior, i.e. making full repository name as a tarball. For portability and storage efficiency, using a tarball file is generally better, especially for disaster recovery backups, uploading backup to blob storage, moving repositories across machines, and less storage space.

Folders are helpful for temporary, local work or modification but not recommended for backups intended for storage or transfer.

In addition, ORAS should also allow users can also customize the name of the folder or tarball file. 

@TerryHowe do you have any suggestions?

docs/proposals/backup-restore.md
FeynmanZhou: For the frist iteration, I would recommend `oras` to start with supporting `.tar` as the file extension to indicate tarball. Other compression file formats like `.tar.gz` and `.zip` could be considered in the future. The reasons are:
- OCI and Docker ecosystem uses `.tar`. Tools like `docker save/load` rely on tarball. This is considered for interoperability with Docker and OCI.
- `.zip`poorly preserves Unix file permissions, symbolic links, etc.,
- `.tar.gz` is not random readable

docs/proposals/backup-restore.md
FeynmanZhou: To keep consistent between `oras back` and `oras restore`, I would recommend `oras restore --input` to start with supporting `.tar` as the file extension to indicate tarball.  https://github.com/oras-project/oras/pull/1610#discussion_r2181029487

docs/proposals/backup-restore.md
TerryHowe: In my version of the document, `--output` was required. I guess you changed that. I think option 1 is the simplist all the other options don't make sense with different registries and repositories.

docs/proposals/backup-restore.md
FeynmanZhou: I intentionally deleted the multi-repo backup and restore scenario because the scenario is not very common but relatively complicated. How about considering multi-repo backup and restore in the future iteration?

I can imagine that there is a scenario that users may want to an backup an entire registry or an application stack that built with multiple images across different repositories.

However, registry vendors usually provide registry-level backup and restore solution. For an application stack, users can also run multiple times `oras backup` to backup multiple repos. So I didn't see a value of support multi-repo backup and restore.

Tag other maintainers @oras-project/oras-cli to see if any other opinions.

docs/proposals/backup-restore.md
TerryHowe: Already have multiple registries working it took no effort

docs/proposals/backup-restore.md
TerryHowe: Already have multi repo and multi-registry working it is not difficult without that back of restore is pointless

TerryHowe:
I reverted this to the last good commit. There were too many issues with those other updates.

FeynmanZhou:
> I reverted this to the last good commit. There were too many issues with those other updates.

Just discussed with @TerryHowe about this PR. The consensus is that we will merge this PR then I will go ahead to open a follow up PR to enrich the spec.

shizhMSFT (approved):
LGTM as the initial attempt of `oras backup/restore`

FeynmanZhou (approved):
Thanks @TerryHowe .  LGTM with merging this PR as an initial version. I will open a follow up PR to enrich the doc.

sabre1041:
@TerryHowe this looks good. One final question. What happens when you attempt to back up the same repository with two separate versions? Since there is the potential to have overlapping concerns/files (like the `index.json` file)

FeynmanZhou:
> @TerryHowe this looks good. One final question. What happens when you attempt to back up the same repository with two separate versions? Since there is the potential to have overlapping concerns/files (like the `index.json` file)

@sabre1041 Would you mind clarifying "two separate versions" in the context of backup a repo?

FeynmanZhou:
@sabre1041 BTW, if overall looks good as the initial version, you may also want to approve this PR as you requested changes previously.

TerryHowe:
> @TerryHowe this looks good. One final question. What happens when you attempt to back up the same repository with two separate versions? Since there is the potential to have overlapping concerns/files (like the `index.json` file)

With OCI file layout it just adds another manifest in one blob directory:
```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:74a8050ec347821b7884ab635f3e7883b5c570388ed8087ffd01fd9fe1cb39c6",
      "size": 3024,
      "annotations": {
        "org.opencontainers.image.ref.name": "registry.k8s.io/kube-apiserver-arm64:v1.31.0"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:9ff42b586c0a57f3fc4a0689afe6db4d8f92f7f79bef3b47b2c75ab112e17de7",
      "size": 3024,
      "annotations": {
        "org.opencontainers.image.ref.name": "registry.k8s.io/kube-apiserver-arm64:v1.32.0"
      }
    }
  ]
}
```

sabre1041 (approved):
LGTM

PR #1623, opened 2025-02-17
fix(ux): show KB in progress bar instead of kB

qweeah:
This PR introduces a minor UX fix on the download speed display. According to [JEDEC unit prefix standard](https://en.wikipedia.org/wiki/JEDEC_memory_standards#Unit_prefixes_for_semiconductor_storage_capacity), `KB` should be used since the size base is 12 not 10.


Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1637, opened 2025-02-26
refactor: use iter for cleaner code

shizhMSFT:

Refactor code for better readability, leveraging the new `iter` package.

### DIFF
--- a/cmd/oras/internal/option/parser.go
+++ b/cmd/oras/internal/option/parser.go
@@ -38,8 +38,8 @@ func Parse(cmd *cobra.Command, optsPtr any) error {
 	return nil
 }
 
-// fields goes through all fields of ptr, optionally run fn if a field is
-// public AND typed T.
+// fields goes through all fields of ptr, optionally run fn if a field is public
+// AND typed T.
 func fields[T any](ptr any) iter.Seq[T] {
 	return func(yield func(T) bool) {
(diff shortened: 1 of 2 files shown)


cmd/oras/internal/option/parser.go
```diff
@@ -28,26 +29,29 @@ type FlagParser interface {
...
-			iface := f.Addr().Interface()
-			if opts, ok := iface.(T); ok {
-				if err := fn(opts); err != nil {
-					return err
+// fields returns an iterator that yields all fields of the given struct that
+// implement the given interface.
+func fields[T any](ptr any) iter.Seq[T] {
+	return func(yield func(T) bool) {
+		v := reflect.ValueOf(ptr).Elem()
+		for i := 0; i < v.NumField(); i++ {
```
Wwwsylvia, line 46: nit: Looks like this can be simplified to:
```suggestion
		for i := range v.NumField()
```

JeyJeyGao (approved):
LGTM, but IANAM

cmd/oras/internal/option/parser.go
shizhMSFT: Good one as this expr was introduced in go 1.22.

Wwwsylvia (approved):
LGTM

Issue #1638, opened 2025-02-27
Improve the help doc for `oras manifest index` sub commands


Wwwsylvia:
## Problem

Currently, the help doc for the `oras manifest index create` command looks like this:

```shell
$ oras manifest index create -h
[Experimental] Create and push an index from provided manifests. All manifests should be in the same repository

Example - Create an index from source manifests tagged 'linux-amd64' and 'linux-arm64', and push without tagging:
  oras manifest index create localhost:5000/hello linux-amd64 linux-arm64

Example - Create an index from source manifests tagged 'linux-amd64' and 'linux-arm64', and push with the tag 'v1':
  oras manifest index create localhost:5000/hello:v1 linux-amd64 linux-arm64

Example - Create an index from source manifests using both tags and digests, and push with tag 'v1':
  oras manifest index create localhost:5000/hello:v1 linux-amd64 sha256:99e4703fbf30916f549cd6bfa9cdbab614b5392fbe64fdee971359a77073cdf9

Example - Create an index and push it with multiple tags:
  oras manifest index create localhost:5000/hello:tag1,tag2,tag3 linux-amd64 linux-arm64 sha256:99e4703fbf30916f549cd6bfa9cdbab614b5392fbe64fdee971359a77073cdf9

Example - Create and push an index with annotations:
  oras manifest index create localhost:5000/hello:v1 linux-amd64 --annotation "key=val"

Example - Create an index and push to an OCI image layout folder 'layout-dir' and tag with 'v1':
  oras manifest index create --oci-layout layout-dir:v1 linux-amd64 sha256:99e4703fbf30916f549cd6bfa9cdbab614b5392fbe64fdee971359a77073cdf9

Example - Create an index and save it locally to index.json, auto push will be disabled:
  oras manifest index create --output index.json localhost:5000/hello linux-amd64 linux-arm64

Example - Create an index and output the index to stdout, auto push will be disabled:
  oras manifest index create localhost:5000/hello linux-arm64 --output - --pretty

Usage:
  oras manifest index create [flags] <name>[:<tag[,<tag>][...]] [{<tag>|<digest>}...]

Aliases:
  create, pack
```

There are two problems here that could potentially confuse users:
1. The usage guide is not clear about which argument is the target to create and which are the source tags
> Usage:
  oras manifest index create [flags] <name>[:<tag[,<tag>][...]] [{<tag>|<digest>}...]

[...]

PR #1650, opened 2025-03-10
docs: update the help doc for `oras manifest index create`

RohanMishra315:
1.   Changing <name> to <target> explicitly states where the manifest index will be created.
2.   Keeping all positional arguments first improves readability and reduces the risk of users mistaking flags like --oci-layout as having their own positional arguments.

Fixes: #1638

Fixes #

### DIFF
--- a/cmd/oras/root/manifest/index/create.go
+++ b/cmd/oras/root/manifest/index/create.go
@@ -56,7 +56,7 @@ type createOptions struct {
 func createCmd() *cobra.Command {
 	var opts createOptions
 	cmd := &cobra.Command{
-		Use:   "create [flags] <name>[:<tag[,<tag>][...]] [{<tag>|<digest>}...]",
+		Use:   "create [flags] <target>[:<tag[,<tag>][...]] <sources>[{<tag>|<digest>}...]",
 		Short: "[Experimental] Create and push an index from provided manifests",
 		Long: `[Experimental] Create and push an index from provided manifests. All manifests should be in the same repository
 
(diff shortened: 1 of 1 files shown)


cmd/oras/root/manifest/index/create.go
```diff
@@ -56,7 +56,7 @@ type createOptions struct {
 func createCmd() *cobra.Command {
 	var opts createOptions
 	cmd := &cobra.Command{
-		Use:   "create [flags] <name>[:<tag[,<tag>][...]] [{<tag>|<digest>}...]",
+		Use:   "create [flags] <target>[:<tag[,<tag>][...]] <sources>[{<tag>|<digest>}...]",
```
Wwwsylvia, line 59: Just realized that maybe we should still call it `<name`> to be consistent with other commands. What do you think? @wangxiaoxuan273

cmd/oras/root/manifest/index/create.go
Wwwsylvia, line 1: Can we apply similar update to `oras manifest index update` as well?

cmd/oras/root/manifest/index/create.go
wangxiaoxuan273: I think `<target>` is better since it pairs with `sources`.

wangxiaoxuan273 (approved):
Good work! Thanks for the contribution!

cmd/oras/root/manifest/index/create.go
```diff
@@ -76,10 +76,10 @@ Example - Create and push an index with annotations:
   oras manifest index create localhost:5000/hello:v1 linux-amd64 --annotation "key=val"
 
 Example - Create an index and push to an OCI image layout folder 'layout-dir' and tag with 'v1':
-  oras manifest index create --oci-layout layout-dir:v1 linux-amd64 sha256:99e4703fbf30916f549cd6bfa9cdbab614b5392fbe64fdee971359a77073cdf9
-
+  oras manifest index create layout-dir:v1 linux-amd64 sha256:99e4703fbf30916f549cd6bfa9cdbab614b5392fbe64fdee971359a77073cdf9 --oci-layout
+
```
TerryHowe, line 80: Need this space gone
```suggestion
 
```

I kind of think having the `--oci-layout` next to `layout-dir:v1` is more clear.

Wwwsylvia: The reason we changed this was because someone thought that `layout-dir:v1` was a argument to `--oci-layout`, but `--oci-layout` is actually a boolean flag and does not take any argument.

TerryHowe: Even though it isn't required, I think CLI examples in the form `cmd --flags arg1 arg2 ..` are clearer.

Wwwsylvia:
Let's merge this PR first and finish the remaining part in a separate PR.

Wwwsylvia (approved):
LGTM, but we will need another PR for `oras manifest index update`

PR #1660, opened 2025-03-18
fix: bring back `--verbose` flag

### LINKED ISSUE #1640: --verbose flag removed in oras 1.3.0-beta.2

The ```--verbose flag``` was removed in 1.3.0-beta.2:

Image

In previous versions (e.g. 1.1.0) it used to work:

Image

```--verbose``` should continue to work in version 1.3.0-beta.2.

### How can we reproduce it?

Try to use any subcommand with the ```--verbose``` flag.

Version:        1.3.0-beta.2
Go version:     go1.24.0
OS/Arch:        darwin/amd64
Git commit:     87379d22f42b5132ed583587766f0a780f9bd322
Git tree state: clean

Mac OS Sonoma (14.7.2)

### PR DESCRIPTION
**Pull request by wangxiaoxuan273**

This PR brings back the `--verbose` flag to the following commands and mark it as deprecated.

oras blob delete
oras blob fetch
oras login
oras manifest delete
oras manifest fetch
oras manifest fetch-config
oras repo ls
oras repo tags
oras resolve
oras tag

Fixes #1640 

### DIFF
--- a/cmd/oras/internal/option/verbose.go
+++ b/cmd/oras/internal/option/verbose.go
@@ -1,0 +1,25 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package option
+
+import "github.com/spf13/pflag"
+
+// AddDeprecatedVerboseFlag adds the deprecated verbose flag to a command
+func AddDeprecatedVerboseFlag(flags *pflag.FlagSet) {
+	// ignoring the variable of the verbose flag, since we will not use it
+	_ = flags.BoolP("verbose", "v", true, "verbose output")
+	_ = flags.MarkDeprecated("verbose", "and will be removed in a future release.")
+}
--- a/cmd/oras/internal/option/verbose.go
+++ b/cmd/oras/internal/option/verbose.go
@@ -20,6 +20,6 @@ import "github.com/spf13/pflag"
 // AddDeprecatedVerboseFlag adds the deprecated verbose flag to a command
 func AddDeprecatedVerboseFlag(flags *pflag.FlagSet) {
 	// ignoring the variable of the verbose flag, since we will not use it
-	_ = flags.BoolP("verbose", "v", true, "verbose output")
+	_ = flags.BoolP("verbose", "v", false, "verbose output")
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: cmd/oras/root/login.go [hunk 1]
```diff
@@ -73,6 +75,8 @@ Example - Log in with username and password in an interactive terminal and no TL
 			return runLogin(cmd, opts)
 		},
 	}
+	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", true, "print status output for unnamed blobs")
```
Wwwsylvia, line 78: nit: Although the description is hidden, how about making the description simply "verbose output"? The original description looks a bit misleading in code.

Wwwsylvia (approved):
LGTM with nit

cmd/oras/root/login.go
wangxiaoxuan273: Thanks for catching! It should be "verbose output", I forgot to change the description when pasting.

Wwwsylvia (approved):
LGTM

cmd/oras/root/blob/delete.go
```diff
@@ -37,6 +37,8 @@ type deleteBlobOptions struct {
 	option.Descriptor
 	option.Pretty
 	option.Target
+	// Deprecated: verbose is deprecated and will be removed in the future.
```
TerryHowe, line 40: Didn't @shizhMSFT make a comment in a previous PR about not commenting private members? That potentially would apply here as well.

cmd/oras/root/blob/delete.go
```diff
@@ -69,6 +71,8 @@ Example - Delete a blob and print its descriptor:
 		},
 	}
 
+	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", true, "verbose output")
+	_ = cmd.Flags().MarkDeprecated("verbose", "and will be removed in a future release.")
```
shizhMSFT, line 75: Can we refactor it out? Since the verbose variable completely has no effect, can we just point it to nowhere?

We can have a quick discuss offline for the trick.

shizhMSFT (changes requested):

cmd/oras/root/blob/delete.go
wangxiaoxuan273: removed this field after refactoring

cmd/oras/root/blob/delete.go
wangxiaoxuan273: refactored after offline discussion

Wwwsylvia (approved):
LGTM

cmd/oras/internal/option/verbose.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package option
+
+import "github.com/spf13/pflag"
+
+// AddDeprecatedVerboseFlag adds the deprecated verbose flag to a command
+func AddDeprecatedVerboseFlag(flags *pflag.FlagSet) {
+	// ignoring the variable of the verbose flag, since we will not use it
+	_ = flags.BoolP("verbose", "v", true, "verbose output")
```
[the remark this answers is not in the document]
wangxiaoxuan273: Updated

shizhMSFT (approved):
LGTM

PR #1673, opened 2025-03-29
refactor: simplify manifest index metadata handler

TerryHowe:

This is kind of an example of why composition can be better than inheritance.

### DIFF
--- a/cmd/oras/internal/display/metadata/discard.go
+++ b/cmd/oras/internal/display/metadata/discard.go
@@ -48,7 +48,7 @@ func (Discard) OnTagged(ocispec.Descriptor, string) error {
 }
 
 // OnIndexCreated implements ManifestIndexCreateHandler.
-func (Discard) OnIndexCreated(ocispec.Descriptor) {}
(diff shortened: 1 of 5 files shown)


wangxiaoxuan273:
Hi @TerryHowe, thanks for your contribution.

Having separate `OnIndexCreated` and `Render` looks unnecessary, but this is our design to ensure future extensibility. Let me explain.

First of all, we need `metadata.Render` method for almost all commands' metadata handler, to signal that no more operations are coming and metadata can be printed. It has to be implemented by `discard`, and as go does not support method overloading, the `Render` of every command has to have the same parameters. As a result, we made it an interface and almost all metadata handlers implement it. 

But we do need to pass in data for some commands, that's why we have functions like `OnIndexCreated`.

So we are not going to proceed with this PR, but thanks for your work anyway.

TerryHowe:
I think there are a couple more that do not need the `Render` method. There is no need to rigidly follow the same pattern for all metadata handlers when it adds no value. It adds complexity and confusion.

PR #1675, opened 2025-03-31
fix: use test Setenv during tests

TerryHowe:

Use test environment for cache test for lint and reliability.


Wwwsylvia (approved):
LGTM

PR #1676, opened 2025-03-31
fix: staticcheck fails for embedded member

TerryHowe:

These embedded members fail staticcheck for newest lint because the embedded member cannot be extracted. This PR is for tests only. There is another PR for non-tests.


Wwwsylvia (approved):
LGTM

PR #1677, opened 2025-03-31
fix: discover handler errors for lint

TerryHowe:

New lint fails without this better error handling. Actually pay attention to some error returns and ignore others.


cmd/oras/internal/display/metadata/table/discover.go
```diff
@@ -55,21 +55,30 @@ func (h *discoverHandler) OnDiscovered(referrer, subject ocispec.Descriptor) err
 }
 
 // Render implements metadata.DiscoverHandler.
-func (h *discoverHandler) Render() error {
+func (h *discoverHandler) Render() (err error) {
 	if n := len(h.referrers); n > 1 {
-		fmt.Fprintln(h.out, "Discovered", n, "artifacts referencing", h.rawReference)
+		_, err = fmt.Fprintln(h.out, "Discovered", n, "artifacts referencing", h.rawReference)
```
Wwwsylvia, line 60: Wondering if the message can be simplified to something like "artifact(s)" so that we don't need the if-else clause🤔

cmd/oras/internal/option/packer.go
```diff
@@ -120,6 +120,6 @@ func decodeJSON(filename string, v interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
+	defer func() { _ = file.Close() }()
```
Wwwsylvia, line 123: We should return the close error

cmd/oras/internal/option/format.go
```diff
@@ -102,7 +102,7 @@ func (opts *Format) ApplyFlags(fs *pflag.FlagSet) {
 func (opts *Format) Parse(cmd *cobra.Command) error {
 	// print deprecation message for table format
 	if opts.FormatFlag == FormatTypeTable.Name {
-		fmt.Fprint(cmd.ErrOrStderr(), "Format \"table\" is deprecated and will be removed in a future release.\n")
+		_, _ = fmt.Fprint(cmd.ErrOrStderr(), "Format \"table\" is deprecated and will be removed in a future release.\n")
```
Wwwsylvia, line 105: Just wondering when do we return the print error and when we do not...

cmd/oras/root/push.go
```diff
@@ -185,7 +185,7 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 	if err != nil {
 		return err
 	}
-	defer store.Close()
+	defer func() { _ = store.Close() }()
```
Wwwsylvia, line 188: We need to return the error if `Close()` fails

cmd/oras/internal/option/format.go
TerryHowe: Moved this to https://github.com/oras-project/oras/pull/1683

cmd/oras/internal/option/packer.go
TerryHowe: Moved this to https://github.com/oras-project/oras/pull/1684

cmd/oras/internal/display/metadata/table/discover.go
TerryHowe: Someone thought this was important sometime. I think a topic of another PR though, this is just for lint

cmd/oras/internal/display/metadata/table/discover_test.go
```diff
@@ -138,3 +139,60 @@ func TestTableDiscoverHandler_Failure(t *testing.T) {
 		t.Errorf("Expected error <%s> actual error <%s>", expected, err.Error())
 	}
 }
+
+type writeFailure struct {
```
qweeah, line 143: Should we add it to test utility? Should be frequently reused for mocking writing errors.

TerryHowe: Done.

Wwwsylvia (approved):
LGTM

PR #1678, opened 2025-03-31
fix: calls to close in tests unchecked

TerryHowe:

Ignore some error returns for tests to fix newer lint.

### DIFF
--- a/cmd/oras/internal/option/pretty_test.go
+++ b/cmd/oras/internal/option/pretty_test.go
@@ -45,7 +45,7 @@ func TestPretty_Output(t *testing.T) {
 	if err != nil {
 		t.Fatal("error calling os.Create(), error =", err)
 	}
-	defer fp.Close()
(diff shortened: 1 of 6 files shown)


cmd/oras/internal/option/pretty_test.go
```diff
@@ -45,7 +45,7 @@ func TestPretty_Output(t *testing.T) {
 	if err != nil {
 		t.Fatal("error calling os.Create(), error =", err)
 	}
-	defer fp.Close()
+	defer func() { _ = fp.Close() }()
```
Wwwsylvia, line 48: If `Close()` fails, should we fail the test, or at least log the error?

Wwwsylvia: Just realized this is a test file. Never mind.

Wwwsylvia (approved):
LGTM

PR #1679, opened 2025-03-31
fix: staticcheck lint failures of embedded member

TerryHowe:

Newer lint fails on these embedded struct members reference.


Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1680, opened 2025-03-31
fix: unneeded error check

TerryHowe:

This error check will always be false.


Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1683, opened 2025-04-02
fix: explicitly ignore return for newest lint

TerryHowe:

Explicitly ignore return from this print for compatibility with newest version of lint. This message is sent to stderr and will not impact the important output of commands. We should care more for data output to stdout.


Wwwsylvia (approved):
LGTM

PR #1687, opened 2025-04-03
fix: blob file close handling for lint

TerryHowe:

Lint was complaining about the closing handling here and I noticed some issues with this method.


Wwwsylvia (approved):
LGTM

PR #1688, opened 2025-04-03
fix: packer decode json lint failure

TerryHowe:

Attempt to fix options packer decode json lint failure with tests.

### DIFF
--- a/cmd/oras/internal/option/packer.go
+++ b/cmd/oras/internal/option/packer.go
@@ -114,12 +114,16 @@ func (opts *Packer) parseAnnotations(cmd *cobra.Command) error {
 	return nil
 }
 
-// decodeJSON decodes a json file v to filename.
+// decodeJSON decodes file contents into json.
 func decodeJSON(filename string, v interface{}) error {
 	file, err := os.Open(filename)
 	if err != nil {
 		return err
 	}
-	defer file.Close()
-	return json.NewDecoder(file).Decode(v)
+	err = json.NewDecoder(file).Decode(v)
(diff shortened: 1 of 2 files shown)


cmd/oras/internal/option/packer.go
```diff
@@ -114,12 +114,17 @@ func (opts *Packer) parseAnnotations(cmd *cobra.Command) error {
 	return nil
 }
 
-// decodeJSON decodes a json file v to filename.
+// decodeJSON decodes file contents into json.
 func decodeJSON(filename string, v interface{}) error {
```
Wwwsylvia, line 118: We need to make `error` a named return value
```suggestion
func decodeJSON(filename string, v interface{}) (err error) {
```

qweeah (approved):
LGTM

cmd/oras/internal/option/packer.go
```diff
@@ -114,12 +114,16 @@ func (opts *Packer) parseAnnotations(cmd *cobra.Command) error {
...
 		return err
 	}
-	defer file.Close()
-	return json.NewDecoder(file).Decode(v)
+	err = json.NewDecoder(file).Decode(v)
+	if err != nil {
+		_ = file.Close()
+		return err
+	}
+	return file.Close()
```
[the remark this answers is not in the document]
TerryHowe: Done.

cmd/oras/internal/option/packer.go
TerryHowe: Done.

PR #1712, opened 2025-04-25
Add test coverage for discover handler

TerryHowe:

Test coverage for the discover handler.


qweeah (approved):
LGTM. Good to know that a plural form should be used for 0.

PR #1722, opened 2025-05-08
refactor: move different targets to different files

TerryHowe:

Working on backup/restore I found the target object a bit rigid and I need to make some changes. Some non-controversial clean-up (I hope) should make that easier to review.


cmd/oras/internal/option/interfaces.go
qweeah, line 1: Naming of this file should be `interface.go` without `s`

qweeah (approved):
LGTM with minor suggestion.

Wwwsylvia (approved):
LGTM

PR #1726, opened 2025-05-10
refactor: rename Target AnnotatedReference method

TerryHowe:

Every time I see this method, I wonder what it does. This name makes it clear.

### DIFF
--- a/cmd/oras/internal/display/metadata/text/attach.go
+++ b/cmd/oras/internal/display/metadata/text/attach.go
@@ -27,9 +27,9 @@ import (
 
 // AttachHandler handles text metadata output for attach events.
 type AttachHandler struct {
-	printer            *output.Printer
-	subjectRefByDigest string
-	root               ocispec.Descriptor
+	printer                 *output.Printer
+	subjectDisplayReference string
+	root                    ocispec.Descriptor
 }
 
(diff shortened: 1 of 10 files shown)


qweeah (approved):
LGTM with minor suggestions.

cmd/oras/internal/display/metadata/text/attach.go
```diff
@@ -43,12 +43,12 @@ func NewAttachHandler(printer *output.Printer) metadata.AttachHandler {
 func (ah *AttachHandler) OnAttached(target *option.Target, root ocispec.Descriptor, subject ocispec.Descriptor) {
 	ah.root = root
 	if strings.HasSuffix(target.RawReference, subject.Digest.String()) {
-		ah.subjectRefByDigest = target.AnnotatedReference()
+		ah.subjectRefByDigest = target.GetDisplayString()
```
qweeah, line 46: How about `GetDisplayReference`

qweeah: BTW looks like the `subjectRefByDigest` can be changed to `subjectDisplayReference`

TerryHowe: Done.

qweeah (approved):
LGTM

PR #1666, opened 2025-03-22
fix: Change receiver name for target

TerryHowe:
**What this PR does / why we need it**:

Every time I'm in this file, I get confused about what opts is because that is used outside this everywhere for options

PR #1727, opened 2025-05-11
fix: Change receiver name for Remote

TerryHowe:

Every time I'm in this file, I get confused about what opts is because that is used outside this everywhere for options

Similar to https://github.com/oras-project/oras/pull/1666


qweeah (approved):
LGTM. `r` would be even better.

PR #1729, opened 2025-05-12
refactor(perf): optimize cases with no child referrers in recursiveCopy

qweeah:
This PR improves the `recursiveCopy` function to handle cases where there are no child manifests or referrers, ensuring the function exits early in such scenarios. This change improves the function's efficiency and robustness.


cmd/oras/root/cp.go
```diff
@@ -265,3 +237,50 @@ func recursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.T
 	}
 	return err
 }
+
+func prepareCopyOption(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.Target, root ocispec.Descriptor, opts *oras.ExtendedCopyOptions) error {
```
shizhMSFT, line 241: It would be better to have this pattern
```suggestion
func prepareCopyOption(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.Target, root ocispec.Descriptor, opts oras.ExtendedCopyOptions) (oras.ExtendedCopyOptions, error) {
```

cmd/oras/root/cp.go
```diff
@@ -265,3 +237,50 @@ func recursiveCopy(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.T
...
+	if root.MediaType != ocispec.MediaTypeImageIndex && root.MediaType != docker.MediaTypeManifestList {
+		return nil
+	}
```
shizhMSFT, line 244: nit: it is better to have an empty line after line 244.

shizhMSFT (approved):
LGTM

cmd/oras/root/cp.go
qweeah: Good catch.

PR #1730, opened 2025-05-13
fix: rename interfaces file

TerryHowe:

Related to comment in https://github.com/oras-project/oras/pull/1722 renameing interfaces.go to interface.go


qweeah (approved):
LGTM

PR #1733, opened 2025-05-14
refactor: ApplyFlags with prefix for Remote and Target

TerryHowe:

It kind of looks like this Target/Remote code was in the middle of some refactor that was never completed. This doesn't address all the issues, but it does clean up flag/prefix/description handling a bit.

### DIFF
--- a/cmd/oras/internal/option/binary_target.go
+++ b/cmd/oras/internal/option/binary_target.go
@@ -48,8 +48,10 @@ func (target *BinaryTarget) EnableDistributionSpecFlag() {
 
 // ApplyFlags applies flags to a command flag set fs.
 func (target *BinaryTarget) ApplyFlags(fs *pflag.FlagSet) {
-	target.From.ApplyFlagsWithPrefix(fs, "from", "source")
-	target.To.ApplyFlagsWithPrefix(fs, "to", "destination")
+	target.From.setFlagDetails("from", "source")
+	target.From.ApplyFlags(fs)
+	target.To.setFlagDetails("to", "destination")
+	target.To.ApplyFlags(fs)
 	fs.StringArrayVarP(&target.resolveFlag, "resolve", "", nil, "base DNS rules formatted in `host:port:address[:address_port]` for --from-resolve and --to-resolve")
 }
 
--- a/cmd/oras/internal/option/remote.go
+++ b/cmd/oras/internal/option/remote.go
@@ -87,15 +87,12 @@ func (remo *Remote) EnableDistributionSpecFlag() {
 // ApplyFlags applies flags to a command flag set.
 func (remo *Remote) ApplyFlags(fs *pflag.FlagSet) {
 	remo.ApplyFlagsWithPrefix(fs, "", "")
-	fs.BoolVar(&remo.secretFromStdin, passwordFromStdinFlag, false, "read password from stdin")
-	fs.BoolVar(&remo.secretFromStdin, identityTokenFromStdinFlag, false, "read identity token from stdin")
+	remo.applyStdinFlags(fs)
 }
 
-func applyPrefix(prefix, description string) (flagPrefix, notePrefix string) {
-	if prefix == "" {
-		return "", ""
-	}
-	return prefix + "-", description + " "
+func (remo *Remote) applyStdinFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&remo.secretFromStdin, passwordFromStdinFlag, false, "read password from stdin")
+	fs.BoolVar(&remo.secretFromStdin, identityTokenFromStdinFlag, false, "read identity token from stdin")
 }
 
 // ApplyFlagsWithPrefix applies flags to a command flag set with a prefix string.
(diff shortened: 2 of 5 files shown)


Wwwsylvia:
The E2E test failed, oops

cmd/oras/internal/option/remote.go
```diff
@@ -84,53 +84,44 @@ func (remo *Remote) EnableDistributionSpecFlag() {
 	remo.applyDistributionSpec = true
 }
 
-// ApplyFlags applies flags to a command flag set.
```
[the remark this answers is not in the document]
TerryHowe: Brought that back, thanks

TerryHowe:
Fixed this a while back

cmd/oras/internal/option/remote.go
```diff
@@ -105,32 +96,37 @@ func (remo *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
...
-	fs.StringVarP(&remo.Username, remo.flagPrefix+usernameFlag, shortUser, "", notePrefix+"registry username")
-	fs.StringVarP(&remo.Secret, remo.flagPrefix+passwordFlag, shortPassword, "", notePrefix+"registry password or identity token")
-	fs.StringVar(&remo.Secret, remo.flagPrefix+identityTokenFlag, "", notePrefix+"registry identity token")
-	fs.BoolVar(&remo.Insecure, remo.flagPrefix+"insecure", false, "allow connections to "+notePrefix+"SSL registry without certs")
+	if fs.Lookup(passwordFromStdinFlag) == nil {
+		fs.BoolVar(&remo.secretFromStdin, passwordFromStdinFlag, false, "read password from stdin")
+	}
+	if fs.Lookup(identityTokenFromStdinFlag) == nil {
+		fs.BoolVar(&remo.secretFromStdin, identityTokenFromStdinFlag, false, "read identity token from stdin")
+	}
```
Wwwsylvia: Why moving it here? Looks like it's nothing to do with the prefix.

TerryHowe: For some reason, these flags do not have prefixes and so this just makes sure they are only added once.

Wwwsylvia: Should we keep these in `ApplyFlags`?

TerryHowe: Done. I forgot why I did that, but there may have been another reason.

TerryHowe: It doesn't properly add the flags when they are in ApplyFlags. I reverted that change and I'll try to understand why later.

The lookup of the flags is a pretty harmless action. It just checks if they've been added.

cmd/oras/internal/option/remote.go
```diff
@@ -105,32 +96,37 @@ func (remo *Remote) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description
...
 	}
-	fs.StringVarP(&remo.Username, remo.flagPrefix+usernameFlag, shortUser, "", notePrefix+"registry username")
-	fs.StringVarP(&remo.Secret, remo.flagPrefix+passwordFlag, shortPassword, "", notePrefix+"registry password or identity token")
-	fs.StringVar(&remo.Secret, remo.flagPrefix+identityTokenFlag, "", notePrefix+"registry identity token")
-	fs.BoolVar(&remo.Insecure, remo.flagPrefix+"insecure", false, "allow connections to "+notePrefix+"SSL registry without certs")
+	if fs.Lookup(passwordFromStdinFlag) == nil {
+		fs.BoolVar(&remo.secretFromStdin, passwordFromStdinFlag, false, "read password from stdin")
+	}
+	if fs.Lookup(identityTokenFromStdinFlag) == nil {
+		fs.BoolVar(&remo.secretFromStdin, identityTokenFromStdinFlag, false, "read identity token from stdin")
```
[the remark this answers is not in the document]
TerryHowe: No, two command line flags set the same value which doesn't sound great.

cmd/oras/internal/option/target.go
```diff
@@ -81,17 +86,10 @@ func (target *Target) GetDisplayReference() string {
...
-
-// ApplyFlagsWithPrefix applies flags to a command flag set with a prefix string.
-// Commonly used for non-unary remote targets.
-func (target *Target) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description string) {
-	target.applyFlagsWithPrefix(fs, prefix, description)
-	target.Remote.ApplyFlagsWithPrefix(fs, prefix, description)
+func (target *Target) ApplyFlags(fs *pflag.FlagSet) {
+	target.ApplyFlagsWithPrefix(fs, target.prefix, target.description)
+	fs.BoolVarP(&target.IsOCILayout, target.prefix+"oci-layout", "", false, "set "+target.description+"target as an OCI image layout")
+	fs.StringVar(&target.Path, target.prefix+"oci-layout-path", "", "[Experimental] set the path for the "+target.description+"OCI image layout target")
```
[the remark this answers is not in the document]
TerryHowe: Problem is Remote is the base class Copilot

cmd/oras/internal/option/remote.go
TerryHowe: Okay, I took a different approach which is a bit more explicit. Let me know what you think.

cmd/oras/internal/option/binary_target.go
```diff
@@ -48,8 +48,11 @@ func (target *BinaryTarget) EnableDistributionSpecFlag() {
 
 // ApplyFlags applies flags to a command flag set fs.
 func (target *BinaryTarget) ApplyFlags(fs *pflag.FlagSet) {
-	target.From.ApplyFlagsWithPrefix(fs, "from", "source")
-	target.To.ApplyFlagsWithPrefix(fs, "to", "destination")
+	target.From.setFlagDetails("from", "source")
+	target.From.ApplyFlags(fs)
+	target.From.applyStdinFlags(fs)
```
Wwwsylvia: It looks like this has already been covered in `target.From.ApplyFlags`. Also, `--identity-token-stdin` and `--password-stdin` are not applicable to `oras cp`.

TerryHowe: Ah, good point. Fixed!

cmd/oras/internal/option/target.go
```diff
@@ -81,17 +86,13 @@ func (target *Target) GetDisplayReference() string {
...
 // the full form is not implemented until a new type comes in.
-func (target *Target) applyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description string) {
-	flagPrefix, notePrefix := applyPrefix(prefix, description)
-	fs.BoolVarP(&target.IsOCILayout, flagPrefix+"oci-layout", "", false, "set "+notePrefix+"target as an OCI image layout")
-	fs.StringVar(&target.Path, flagPrefix+"oci-layout-path", "", "[Experimental] set the path for the "+notePrefix+"OCI image layout target")
-}
-
-// ApplyFlagsWithPrefix applies flags to a command flag set with a prefix string.
-// Commonly used for non-unary remote targets.
-func (target *Target) ApplyFlagsWithPrefix(fs *pflag.FlagSet, prefix, description string) {
```
shizhMSFT, line 92: The `Target.ApplyFlagsWithPrefix` is to override `Remote.ApplyFlagsWithPrefix`. Will it be dangerous to call `Target.ApplyFlagsWithPrefix` without calling `Target.setFlagDetails`?

TerryHowe: ApplyFlagsWithPrefix can be called with without calling setFlagDetails. By default, strings are empty

PR #1734, opened 2025-05-14
fix(lint): ignore returns from test defer calls

TerryHowe:

Ignoring two test defer calls for lint


Wwwsylvia (approved):
LGTM

PR #1735, opened 2025-05-14
fix: lint error for pull close

TerryHowe:

I'm not sure what happened here, but the close in pull was not properly reporting errors. Lint complained.


cmd/oras/root/pull.go
```diff
@@ -206,7 +206,7 @@ func doPull(ctx context.Context, src oras.ReadOnlyTarget, dst oras.GraphTarget,
 			}
 			defer func() {
 				if fetchErr != nil {
-					rc.Close()
+					_ = rc.Close()
```
Wwwsylvia, line 209: Looks like this can be tested by mocking src.Fetch() to return an error producing rc?

TerryHowe: True.

PR #1736, opened 2025-05-15
fix: replace slave var for child

TerryHowe:

Rename variable called slave with the name child


Wwwsylvia (approved):
LGTM

PR #1738, opened 2025-05-17
fix: rename slave to child for cp test

TerryHowe:

These tests don't work on mac, so running here, but just renaming slave to child


Wwwsylvia (approved):
LGTM

PR #1758, opened 2025-06-26
feat: support `--format` for `oras repo ls`

Wwwsylvia:

Fixes #


cmd/oras/internal/display/metadata/json/repo_list.go
```diff
@@ -0,0 +1,49 @@
...
+func NewRepoListHandler(out io.Writer) metadata.RepoListHandler {
+	return &repoListHandler{
+		out:   out,
+		model: model.NewRepositories(),
+	}
+}
+
+// OnRepositoryListed implements metadata.RepoListHandler.
+func (h *repoListHandler) OnRepositoryListed(repo string) error {
+	h.model.AddRepository(repo)
```
TerryHowe, line 43: This pattern is going to be slow on very large lists. It would be pretty trivial to generate json for repos and tags, but future problem I guess.

TerryHowe (commented):
/lgtm

shizhMSFT:
> Do we want to update the [spec](https://github.com/oras-project/oras/blob/main/docs/proposals/formatted-output.md) first? This implementation is not aligned to #733. E.g. if a sub namespace is used for MCR, the output is not repository but the output is saying it is
> 
> ```
> # E.g. `oras` is not a repository but `oss/v2/oras-project/oras` is
> > oras repo ls mcr.microsoft.com/oss/v2/oras-project --format json
> {
>   "repositories": [
>     "oras"
>   ]
> }
> ```

Nice catch, @qweeah !

I think we should have a refined model like
```console
$ oras repo ls mcr.microsoft.com/oss/v2/oras-project --format json
{
  "registry": "mcr.microsoft.com",
  "repositories": [
    "oss/v2/oras-project/oras"
  ]
}
```
so that it is consistent with the output of the [distribution catalog API](https://distribution.github.io/distribution/spec/api/#listing-repositories).

Wwwsylvia:
> > Do we want to update the [spec](https://github.com/oras-project/oras/blob/main/docs/proposals/formatted-output.md) first? This implementation is not aligned to #733. E.g. if a sub namespace is used for MCR, the output is not repository but the output is saying it is
> > ```
> > # E.g. `oras` is not a repository but `oss/v2/oras-project/oras` is
> > > oras repo ls mcr.microsoft.com/oss/v2/oras-project --format json
> > {
> >   "repositories": [
> >     "oras"
> >   ]
> > }
> > ```
> 
> Nice catch, @qweeah !
> 
> I think we should have a refined model like
> 
> ```
> $ oras repo ls mcr.microsoft.com/oss/v2/oras-project --format json
> {
>   "registry": "mcr.microsoft.com",
>   "repositories": [
>     "oss/v2/oras-project/oras"
>   ]
> }
> ```
> 
> so that it is consistent with the output of the [distribution catalog API](https://distribution.github.io/distribution/spec/api/#listing-repositories).

SGTM

cmd/oras/internal/display/metadata/json/repo_list.go
Wwwsylvia: >  It would be pretty trivial to generate json for repos and tags

Yeah indeed, but printing the JSON struct line-by-line doesn't form a valid JSON object until the end. For example:

```json
{
"repositories": [
    "foo"
    "bar"
```

But we might want to limit the max number of repos/tags to avoid OOM

Wwwsylvia: > But we might want to limit the max number of repos/tags to avoid OOM

Opened an issue to track: #1762

cmd/oras/internal/display/metadata/interface.go
```diff
@@ -143,5 +143,5 @@ type RepoListHandler interface {
 	Renderer
 
 	// OnRepositoryListed is called for each repository that is listed.
-	OnRepositoryListed(repo string) error
+	OnRepositoryListed(fullRepo, subRepo string) error
```
shizhMSFT: This abstraction confuses me. `OnRepositoryListed` should pass the full repo only. The namespace should be passed to `NewRepoListHandler`.

Wwwsylvia: Nice catch! Updated.

cmd/oras/internal/display/handler.go
```diff
@@ -266,3 +266,19 @@ func NewRepoTagsHandler(out io.Writer, format option.Format) (metadata.RepoTagsH
 	}
 	return handler, nil
 }
+
+// NewRepoListHandler returns a repo ls handler.
+func NewRepoListHandler(out io.Writer, format option.Format, registry string, namespace string) (metadata.RepoListHandler, error) {
```
shizhMSFT: nit:
```suggestion
func NewRepoListHandler(out io.Writer, format option.Format, registry, namespace string) (metadata.RepoListHandler, error) {
```

shizhMSFT (approved):
LGTM

qweeah (approved):
LGTM

PR #1759, opened 2025-06-27
feat: show repo tags with `--oci-layout-path` in `repo tags` command

### LINKED ISSUE #1612: Bad UX of `--oci-layout-path` flag working with `oras repo tags`

I have an OCI layout that contains several tags, some are in format of FQDN.  I can list the tags by running `oras repo tags --oci-layout $dirName`, for example:

```shell
$ oras repo tags --oci-layout ./repo
example.registry.com/foo:latest
test.com/bar:v1
v2
```

However, if I replace `--oci-layout` with  `--oci-layout-path` like this, there will be an error:

```shell
$ oras repo tags --oci-layout-path ./repo
Error: "oras repo tags" requires exactly 1 argument but got 0
Usage: oras repo tags [flags] <name>
Please specify exactly 1 argument as the target repository to list tags from. Run "oras repo tags -h" for more options and examples
```

The right way is to add one more parameter as the repository name, which should be an empty string in this case:

```shell
oras repo tags --oci-layout-path ./repo ""
example.registry.com/foo:latest
test.com/bar:v1
v2
```

This UX is strange and not intuitive.

The `--oci-layout=path` flag may not be applicable to the `oras repo tags` command.

### How can we reproduce it?

1. Run `oras repo tags --oci-layout $dirName`
2. Run `oras repo tags --oci-layout-path $dirName ""`

Version:        1.3.0-beta.1
Go version:     go1.23.4
OS/Arch:        linux/amd64
Git commit:     2ca02f6f6514200366df26ba9ef6d1b311629dad
Git tree state: clean

Ubuntu 24.04 in WSL

wangxiaoxuan273:

This PR implements the new feature of showing repo tags with `--oci-layout-path` in `repo tags` command. This feature applies to the scenario where the user stores images from multiple repositories in the same OCI layout directory, and put the full references of the images as tags. For example, an OCI layout may contain the following tags:
```
example.registry.com/foo:latest
test.com/bar:v1
v2
```
In this case, given the command `oras repo tags --oci-layout-path layout-dir test.com/bar`, the output would be
 ```
v1
```
as `v1` is the tag under the repository `test.com/bar`.

`oras repo tags` command now has three functionalities (list tags, list associated tags and showing repo tags) and they are implemented in the code in a mixed way. I think it would be better if we can refactor the code.

Fixes #1612 

### CODE REVIEW

File: cmd/oras/root/repo/tags.go [hunk 1]
```diff
@@ -95,8 +99,21 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
 	if err != nil {
 		return err
 	}
+	// parse opts.Reference, check if it is a directory
+	registryName := ""
+	repositoryName := ""
+	referenceName := ""
+	if cmd.Flags().Changed("oci-layout-path") && opts.Reference != "" {
```
shizhMSFT: Should those stuffs be based on `Type` and `Reference`?

cmd/oras/root/repo/tags.go
```diff
@@ -70,6 +72,9 @@ Example - [Experimental] Show tags of the target repository in JSON view:
 
 Example - [Experimental] Show tags of the target repository with Go template:
   oras repo tags localhost:5000/hello --format go-template --template "{{.tags}}"
+
+Example - [Experimental] Show tags of a specific repository in OCI layout:
+  oras repo tags --oci-layout-path layout-dir localhost:5000/hello
```
TerryHowe, line 77: It seems like a `--filter` option would have made more sense. I really hadn't thought about it before, but this is an odd syntax with `--oci-layout` and `--oci-layout-path`. Not for this PR though.

cmd/oras/root/repo/tags.go
```diff
@@ -95,8 +100,25 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
 	if err != nil {
 		return err
 	}
+
+	// if a repository path is given, filter the tags under the repository
+	var repoFilter func(ref string) (string, bool)
+	if opts.Target.Type == option.TargetTypeOCILayout {
+		ref, err := registry.ParseReference(opts.Reference)
```
TerryHowe, line 107: Should this be limited to filtering only tags that are in the format of registry/repo? Why not `HasPrefix` or `Contains` on any string?

cmd/oras/root/repo/tags.go
```diff
@@ -115,6 +137,13 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
 	}
 	err = finder.Tags(ctx, opts.last, func(tags []string) error {
 		for _, tag := range tags {
+			if repoFilter != nil {
+				if scopedTag, ok := repoFilter(tag); ok {
+					tag = scopedTag
```
TerryHowe: Why modify the tags matching the filter? That seems like it would be confusing.

cmd/oras/root/repo/tags.go
```diff
@@ -95,8 +100,25 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
...
 		return err
 	}
+
+	// if a repository path is given, filter the tags under the repository
+	var repoFilter func(ref string) (string, bool)
+	if opts.Target.Type == option.TargetTypeOCILayout {
+		ref, err := registry.ParseReference(opts.Reference)
+		if err == nil && ref.Reference == "" {
+			prefix := fmt.Sprintf("%s/%s:", ref.Registry, ref.Repository)
+			repoFilter = func(ref string) (string, bool) {
```
TerryHowe: Seems like it would be clearer just to put this code below

cmd/oras/root/repo/tags.go
wangxiaoxuan273: We decide to make it limited to filtering only tags that are in the format of registry/repo, as this is the common use case we see from the users.

cmd/oras/root/repo/tags.go
wangxiaoxuan273: We want to show the tags under a specific repository namespace. For example, given that there exists the tag `test.com/bar:v1` in our oci-layout, the command `oras repo tags --oci-layout-path layout-dir test.com/bar` would give `v1`,

cmd/oras/root/repo/tags.go
wangxiaoxuan273: Do you suggest we should remove one of the oci flags and add a filter flag?

cmd/oras/root/repo/tags.go
wangxiaoxuan273: Where do you suggest to put this code?

wangxiaoxuan273:
Need to consider the case that `--exclude-digest-tags` is used together with `--oci-layout-path`. We need to add an extra E2E test case.

cmd/oras/root/repo/tags.go
wangxiaoxuan273: Do you mean we should show the intact tag, instead of trimming it?

TerryHowe: Yeh, trimming the tags seems odd to me. I see how it is useful, but it is such a specific use case. I wonder if changing to a `--registry-filter <filter>` option at least would make it clearer.

TerryHowe: Although `--registry-filter` doesn't make any sense for remote registries

wangxiaoxuan273: Thanks for your input Terry.

wangxiaoxuan273: We had an offline discussion and here's our reasoning of showing the trimmed tags: users use tags like `test.com/bar:v1` only when they put images from multiple repositories in one oci layout, and in this case the put the full reference of the image as the tag. So it's reasonable to interpret `test.com/bar` as repository and `v1` as the tag. And we want to display only the tags.

cmd/oras/root/repo/tags.go
```diff
@@ -115,10 +131,17 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
 	}
 	err = finder.Tags(ctx, opts.last, func(tags []string) error {
 		for _, tag := range tags {
+			if targetPrefix != "" {
```
Wwwsylvia, line 136: Can we add a few comments for each if statement like this for better maintainability?

cmd/oras/root/repo/tags.go
```diff
@@ -115,10 +131,17 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
...
+				if scopedTag, ok := strings.CutPrefix(tag, targetPrefix); ok {
+					tag = scopedTag
+				} else {
+					continue
+				}
+			}
```
Wwwsylvia, line 142: nit: Might be helpful to add an empty line between these if statements to make it more clear.

Wwwsylvia (commented):
Overall LGTM

cmd/oras/root/repo/tags.go
TerryHowe: No, I'm suggesting a `--filter` flag instead of supporting another argument when OCI layout.

cmd/oras/root/repo/tags.go
wangxiaoxuan273: Added empty lines.

cmd/oras/root/repo/tags.go
wangxiaoxuan273: Added comments for each if statement

cmd/oras/root/repo/tags.go
```diff
@@ -115,10 +131,23 @@ func showTags(cmd *cobra.Command, opts *showTagsOptions) error {
...
 	err = finder.Tags(ctx, opts.last, func(tags []string) error {
 		for _, tag := range tags {
+			// if --oci-layout-path is used with a repository path, filter the
+			// tags under the repository.
+			if targetPrefix != "" {
+				if scopedTag, ok := strings.CutPrefix(tag, targetPrefix); ok {
+					tag = scopedTag
+				} else {
+					continue
+				}
```
shizhMSFT: nit: code style: less cognitive load.
```suggestion
				scopedTag, ok := strings.CutPrefix(tag, targetPrefix)
				if !ok {
					continue
				}
				tag = scopedTag
```

shizhMSFT (approved):
LGTM with suggestions

shizhMSFT (approved):
LGTM

Issue #1547, opened 2024-11-26
Support showing more information of tags in formatted output


FeynmanZhou:
### What is the version of your ORAS CLI

v1.2.0

### What would you like to be added?

When listing the tags in an OCI image layout or an repo within OCI registry. The digest and platform information can be listed alongside the tags in the output. 

```bash
$ oras repo tags --oci-layout layout-dir --format json
```

```
{
  "reference": "localhost:5000/oras@sha256:7414904f07f515f48fe4afeaf876e3151039a81e7177b9c66e9e7ed6dd186111",
  "tags": [
    {
      "tag": "v1"
      "digest": "sha256:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "size": 820
    },
     {
      "tag": "v1-linux-amd64"
      "digest": "sha256:yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy",
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 200
    },
    {
      "tag": "v1-linux-arm64"
      "digest": "sha256:zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz",
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 200
    },
  ]
}
```

### Why is this needed for ORAS?

- With the media type of each tag showing in the formatted output, users should be able to differentiate from an image index and image manifest in a repo
- Easier to users to list all available tags with their descriptor in a repo, so that users can compose an image index by selecting from a few arch-specific tags
- For scripting and automation scenario

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

shizhMSFT:
Besides the output formatting minor issues, there is a perf concern: if a repo has hundreds of tags, this command will appear to hang without any response.

Wwwsylvia:
+1 on supporting `--format json` for `oras repo tags`.

> Besides the output formatting minor issues, there is a perf concern: if a repo has hundreds of tags, this command will appear to hang without any response.

We can start with outputting the raw JSON response from the registry. For example:

```console
$ oras repo tags docker.io/library/hello-world --format json
{
    "name": "library/hello-world",
    "tags": [
        "latest",
        "linux",
        "nanoserver",
        "nanoserver-1709",
        "nanoserver-1803",
        "nanoserver-1809",
[...]

PR #1763, opened 2025-07-02
docs(spec): define output formats for oras repo commands

Wwwsylvia:
This PR adds documentation for the output formats for `oras repo ls` and `oras repo tags` commands.

Related to #1547 
Related to #1758 

### DIFF
--- a/docs/proposals/formatted-output.md
+++ b/docs/proposals/formatted-output.md
@@ -10,12 +10,19 @@
   - [Scenarios](#scenarios)
     - [Scripting](#scripting)
     - [CI/CD](#cicd)
+    - [Verify local files](#verify-local-files)
   - [Proposal and desired user experience](#proposal-and-desired-user-experience)
     - [oras manifest fetch](#oras-manifest-fetch)
     - [oras pull](#oras-pull)
     - [oras push](#oras-push)
     - [oras attach](#oras-attach)
     - [oras discover](#oras-discover)
+      - [Tree format of `oras discover` output](#tree-format-of-oras-discover-output)
+      - [JSON output of `oras discover`](#json-output-of-oras-discover)
+      - [Table format of `oras discover` output](#table-format-of-oras-discover-output)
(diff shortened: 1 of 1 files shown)


docs/proposals/formatted-output.md
```diff
@@ -10,12 +10,19 @@
   - [Scenarios](#scenarios)
     - [Scripting](#scripting)
     - [CI/CD](#cicd)
+    - [Verify local files](#verify-local-files)
   - [Proposal and desired user experience](#proposal-and-desired-user-experience)
     - [oras manifest fetch](#oras-manifest-fetch)
     - [oras pull](#oras-pull)
     - [oras push](#oras-push)
     - [oras attach](#oras-attach)
     - [oras discover](#oras-discover)
```
TerryHowe, line 19: None of the other commands have links to the format specific sections. Should this be scipped?

docs/proposals/formatted-output.md
```diff
@@ -556,6 +565,117 @@ localhost:5000/kubernetes/kubectl@sha256:bece4f4746a39cb39e38451c70fa5a1e5ea4fa2
...
+{
+  "registry": "localhost:5000",
+  "repositories": [
+    "dev/foo",
+    "dev/bar"
+  ]
+}
+```
+
+The default text output for the same command omits the namespace prefix for better readability:
```
TerryHowe, line 629: seems like it would be more consistent if text had the namespace

Wwwsylvia: Yes I agree. But the concern is that changing the existing default output would be breaking. Here is the historical discussion on why the namespace is omitted at the beginning: https://github.com/oras-project/oras/pull/768#issuecomment-1407609355

docs/proposals/formatted-output.md
Wwwsylvia: I think the ToC is supposed to flatten all levels of titles? The other commands don't have format specific sections.

FeynmanZhou (approved):
LGTM. Thanks @Wwwsylvia

PR #1766, opened 2025-07-03
Added --simple-auth flag  FIXES #1239

### LINKED ISSUE #1239: Optionally support crane/docker style auth for greater compatibility

main

I would like a flag called `--simple-auth`.  If the flag is not present then the existing behavior is used.  Specifically creating a single `auth.newCache()` and used for all targets (source and destination).   If the flag is present then a new `auth.NewSingleContextCache()` ([definition](https://github.com/oras-project/oras-go/blob/c34d275e1ae7d6f2e017e72e0ff88b5525118640/registry/remote/auth/cache.go#L222)) is created for each target (source and destination).

See this [issue](https://github.com/oras-project/oras-go/issues/650) for the details.

tanyabhatnagar:
**Description**

This PR introduces a new CLI flag --simple-auth that enables per-target credential caching using auth.NewSingleContextCache() instead of the default shared cache via auth.NewCache(). This behaviour improves compatibility with tools like crane and docker, which expect scoped credential separation.

 **Changes Made**

Added --simple-auth flag to Remote options.
Updated the authClient() method in cmd/oras/internal/option/remote.go to select the appropriate cache based on this flag.
Injected tlsConfigFn to make authClient() more testable.
Added unit tests in remote_authclient_test.go to verify cache behaviour based on the flag.

**Tests**

authClient() returns auth.NewSingleContextCache() when UseSimpleAuth = true
authClient() returns auth.NewCache() when UseSimpleAuth = false (default behaviour)

### DIFF
--- a/cmd/oras/internal/option/target.go
+++ b/cmd/oras/internal/option/target.go
@@ -59,6 +59,7 @@ type Target struct {
 	Path string
 
 	IsOCILayout bool
+	UseSimpleAuth bool
 }
 
 // ApplyFlags applies flags to a command flag set for unary target
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,3 +30,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "docker"
+    directory: "/" 
+    schedule:
+      interval: "weekly"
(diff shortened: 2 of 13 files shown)


shizhMSFT:
Manually linking to #1239

cmd/oras/internal/option/target.go
```diff
@@ -59,6 +59,7 @@ type Target struct {
 	Path string
 
 	IsOCILayout bool
+	UseSimpleAuth bool
```
TerryHowe, line 62: This would probably belong in `Remote`

PR #1767, opened 2025-07-03
doc: follow up PR to enrich oras backup and oras restore

FeynmanZhou:

This is a follow up PR to enrich oras backup and oras restore

Fixes #


docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,260 @@
...
+oras copy --recursive --to-oci-layout registry.example.com/app/backend:v1.0.0 ./image-backup:v1.0.0 
+```
+
+To restore from an OCI image layout to an image:
+
+```bash
+oras copy --recursive --from-oci-layout ./image-backup:v1.0.0 registry.example.com/app/backend:v1.0.0
+```
+
+Lack of clarity and built-in commands for standardized, reliable image backup and restore causes user confusion and broken workflows. OCI image layout is not widely adopted by users. This pattern is reported repeatedly by users tracked in [GitHub Issue #1160](https://github.com/oras-project/oras/issues/1160), [GitHub Issue #1353](https://github.com/oras-project/oras/issues/1353), [GitHub Issue #1366](https://github.com/oras-project/oras/issues/1366)
```
shizhMSFT: nit:
```suggestion
Lack of clarity and built-in commands for standardized, reliable image backup and restore causes user confusion and broken workflows. OCI image layout is not widely adopted by users. This pattern is reported repeatedly by users tracked in [GitHub Issue #1160](https://github.com/oras-project/oras/issues/1160), [GitHub Issue #1353](https://github.com/oras-project/oras/issues/1353), [GitHub Issue #1366](https://github.com/oras-project/oras/issues/1366).
```

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,260 @@
...
+
+**Output:**
+An OCI image layout directory or `.tar` archive containing the images, artifacts, their metadata, and optional referrers.
+
+**New Flags:**
+
+* `--include-referrers`: Back up the image and its linked referrers (e.g., attestations, SBOMs).
+* `--output <path>`: Required. Target file path or archive tarball file to write in local filesystem.
+
+> [!NOTE] The file extension determines the output format. If the user does not provide an extension and output is a folder path, it defaults to directory. For Windows terminal, `oras` should prompt users that `oras` will generate a tarball file instead of a folder path when specifying a ".tar" extension. When specifying an extention with `.zip` or `.tar.gz`, `oras` should output warning message and prompt that `.zip` or `.tar.gz` is not supported, then create a folder path with given extention as a part of the folder name.
```
shizhMSFT:
```suggestion
> [!NOTE]
> The file extension determines the output format. If the user does not provide an extension and output is a folder path, it defaults to directory. For Windows terminal, `oras` should prompt users that `oras` will generate a tarball file instead of a folder path when specifying a ".tar" extension. When specifying an extention with `.zip` or `.tar.gz`, `oras` should output warning message and prompt that `.zip` or `.tar.gz` is not supported, then create a folder path with given extention as a part of the folder name.
```
So that it renders as
> [!NOTE]
> The file extension determines the output format. If the user does not provide an extension and output is a folder path, it defaults to directory. For Windows terminal, `oras` should prompt users that `oras` will generate a tarball file instead of a folder path when specifying a ".tar" extension. When specifying an extention with `.zip` or `.tar.gz`, `oras` should output warning message and prompt that `.zip` or `.tar.gz` is not supported, then create a folder path with given extention as a part of the folder name.

shizhMSFT: typo: `extention` --> `extension`
```suggestion
> [!NOTE] The file extension determines the output format. If the user does not provide an extension and output is a folder path, it defaults to directory. For Windows terminal, `oras` should prompt users that `oras` will generate a tarball file instead of a folder path when specifying a ".tar" extension. When specifying an extension with `.zip` or `.tar.gz`, `oras` should output warning message and prompt that `.zip` or `.tar.gz` is not supported, then create a folder path with given extension as a part of the folder name.
```

shizhMSFT: Why Windows terminal is special to Linux or macOS? They are the same, aren't they?

shizhMSFT: Consider the following wording:
```suggestion
> [!NOTE]
> The file extension determines the output format. If the output path does not include a file extension, it is assumed that the output should be a directory. When an unsupported extension such as `.zip` or `.tar.gz` is specified, `oras` should display a warning indicating that the format is not supported. In such cases, it will proceed to create a directory at the specified path instead.
```
> [!NOTE]
> The file extension determines the output format. If the output path does not include a file extension, it is assumed that the output should be a directory. When an unsupported extension such as `.zip` or `.tar.gz` is specified, `oras` should display a warning indicating that the format is not supported. In such cases, it will proceed to create a directory at the specified path instead.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,260 @@
...
+* `--distribution-spec string`: [Preview] set OCI distribution spec version and API option for target. Options: v1.1-referrers-tag, v1.1-referrers-api
+* `--no-tty`: Disable progress bars.
+
+### User Experience in the CLI
+
+**Backup and Restore an Entire Repository and Tagged Artifacts**
+
+```bash
+# Backup a repository from a registry to a local compressed tarball
+oras backup --include-referrers registry.acme.com/repo:v1 --output multi-backup.tar
```
shizhMSFT: To backup the entire repository, tags should not be provided.
> [!NOTE]
> Tags are automatically discovered.
```suggestion
oras backup --output multi-backup.tar --include-referrers registry.acme.com/repo
```

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,260 @@
...
+
+# Transfer the backup file to new environment via secure channels (e.g., BitLocker-enabled removable drives)
+
+# Restore images and referrer artifacts from a local backup file to a target registry
+oras restore registry.backup.com/repo:v1  --input multi-backup.tar
```
shizhMSFT: To restore the entire repository, tags should not be provided.
> [!NOTE]
> Tags are automatically discovered.
```suggestion
oras restore --input multi-backup.tar registry.backup.com/repo 
```

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,260 @@
...
+# Backup multiple repositories from a registry to a local OCI image layout
+$ oras backup registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0 --output k8s-control-plane
+
+# List the tags in the OCI image layout
+$ oras repo tags --oci-layout k8s-control-plane
+registry.k8s.io/kube-apiserver-arm64:v1.31.0
+registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+
+# Restore them to two repositories in different registries
+$ oras restore localhost:5000/kube-apiserver-arm64:v1.31.0 localhost:5000/kube-controller-manager-arm64:v1.31.0 --input k8s-control-plane:v1.31.0
```
shizhMSFT: The `restore` command example is ambiguous. You might need to re-think about it.

docs/proposals/backup-restore.md
```diff
@@ -0,0 +1,260 @@
...
+
+**Offline Snapshot for Air-Gapped Environments**
+
+Create a snapshot of images and referrers for an air-gapped environment:
+
+```bash
+oras backup secure.registry/repo:stable --output airgap-snapshot.tar
+```
+
+By default, the referrers will be restored along with the image to the registry. User could use the flag `--exclude-referrers` to remove the linked referrers.
```
shizhMSFT: This statement is for `oras backup` or `oras restore`?

docs/proposals/backup-restore.md
FeynmanZhou: Agree with you. Removed Windows.

FeynmanZhou: Updated

docs/proposals/backup-restore.md
FeynmanZhou: Good catch. It is for oras backup

docs/proposals/backup-restore.md
FeynmanZhou: Updated the example.

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
+### Scenario 1: Creating Offline Snapshots for Air-Gapped Environments
 
-## Proposal 
+Dave, a security engineer at a FinTech company. To create a snapshot of the image and its referrers in an air-gapped environment, Dave needs to run the following flow:
 
-The backup and restore commands will support reading and writing multiple files to and from a registry.
-As well as the flags described here, the commands will support the normal set of flags to support TLS and authentication.
+1. Packages the image and its referrers from an OCI image layout into a `.tar` for portability.
+2. Copies compressed files via secured channels to the air-gapped network.
+3. Restore all artifacts from a compressed file to an OCI registry in an air-gapped environment.
```
shizhMSFT: nit: consistency: grammar: `Restore` -> `Restores`.

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
 
+No unified snapshot solution available in `docker` or `oras`. Blobs duplicated across files and no assurance of artifact integrity and completeness causes problems to users. See [GitHub Issue #730](https://github.com/oras-project/oras/issues/730)
```
shizhMSFT: nit:
```suggestion
No unified snapshot solution available in `docker` or `oras`. Blobs duplicated across files and no assurance of artifact integrity and completeness causes problems to users. See [GitHub Issue #730](https://github.com/oras-project/oras/issues/730).
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
 ```
 
-Each image will be stored in a subdirectory which matches the repository name.
+Lack of clarity and built-in commands for standardized, reliable image backup and restore causes user confusion and broken workflows. OCI image layout is not widely adopted by users. This pattern is reported repeatedly by users tracked in [GitHub Issue #1160](https://github.com/oras-project/oras/issues/1160), [GitHub Issue #1353](https://github.com/oras-project/oras/issues/1353), [GitHub Issue #1366](https://github.com/oras-project/oras/issues/1366).
+
+## Existing Solutions or Expectations
+
+* `docker save/load` supports exporting and importing images but not referrers or OCI artifacts.
+* `oras pull/push` handles single artifacts, but not repository-level operations.
+* There is no built-in way to persist multiple artifacts in OCI layout format via `oras`.
```
shizhMSFT: It's not true. `oras copy` does that. ... just not that convenient. You know there are lot's scripting jobs to do.

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
 
-#### Optimize blobs
+**New Flags:**
 
-A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
+* `--include-referrers`: Back up the image and its linked referrers (e.g., attestations, SBOMs).
+* `--output <path>`: Required. Target file path or archive tarball file to write in local filesystem.
 
+> [!NOTE] 
+> > The file extension determines the output format. If the output path does not include a file extension, it is assumed that the output should be a directory. When an unsupported extension such as `.zip` or `.tar.gz` is specified, `oras` should display a warning indicating that the format is not supported. In such cases, it will proceed to create a directory at the specified path instead.
```
shizhMSFT: You don't need double quoting.
```suggestion
> The file extension determines the output format. If the output path does not include a file extension, it is assumed that the output should be a directory. When an unsupported extension such as `.zip` or `.tar.gz` is specified, `oras` should display a warning indicating that the format is not supported. In such cases, it will proceed to create a directory at the specified path instead.
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
-The file name does NOT need to end in `tgz`, but file output will be compressed tar file.
+**Output:**
+An OCI image layout directory or `.tar` archive containing the images, artifacts, their metadata, and optional referrers.
 
-#### Optimize blobs
+**New Flags:**
 
-A further enhancement is to create blobs that are duplicated between images as hard links in the directory output and compressed tar files.
+* `--include-referrers`: Back up the image and its linked referrers (e.g., attestations, SBOMs).
+* `--output <path>`: Required. Target file path or archive tarball file to write in local filesystem.
```
shizhMSFT:
```suggestion
* `--output <path>`: Required. Target directory path or tar file path to write in local filesystem.
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
+
+**Offline Snapshot for Air-Gapped Environments**
+
+Create a snapshot of a sample image `registry-a.k8s.io/kube-apiserver:v1` and its referrer (e.g. signature) for an air-gapped environment:
+
+```bash
+oras backup registry-a.k8s.io/kube-apiserver:v1 --include-referrers --output airgap-snapshot.tar
+```
+
+By default, referrers are included in the backup along with the image. Users can use the `--exclude-referrers` flag to omit linked referrers from the backup.
```
shizhMSFT: This statement is not true. It's hallucination.

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
-localhost:15000/my-mirror/kube-controller-manager-arm64:v1.31.0
+oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
 ```
 
-A namespace in the registry will be optional.
-The registry in the above example could be specified as `localhost:15000`.
+The image and linked referrers are reliably restored to another registry with minimal steps.
 
-Any directory in the input that does not contain an `index.json` shall be silently ignored.
+```bash
```
shizhMSFT, line 124:
```suggestion
```console
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
-oras restore --input ./mirror.tgz localhost:15000/my-mirror
+oras restore --input backup.tar registry-b.k8s.io/kube-apiserver
 ```
 
-If the specified source is a file, the format is assumed to be a compressed tar file.
-There will be no validation of file name format.
+List all tags from the repo `registry-b.k8s.io/kube-apiserver`
+
+```bash
+oras repo tags registry-b.k8s.io/kube-apiserver
```
shizhMSFT:
```suggestion
$ oras repo tags registry-b.k8s.io/kube-apiserver
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
 ```bash
-oras restore --input ./mirror.tgz localhost:15000/my-mirror
+oras restore --input backup.tar registry-b.k8s.io/kube-apiserver
 ```
 
-If the specified source is a file, the format is assumed to be a compressed tar file.
-There will be no validation of file name format.
+List all tags from the repo `registry-b.k8s.io/kube-apiserver`
+
+```bash
```
shizhMSFT:
```suggestion
```console
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
+Backup multiple repositories from a registry to a local OCI image layout
+
+```bash
+$ oras backup registry.k8s.io/kube-apiserver registry.k8s.io/kube-controller-manager --output k8s-control-plane
+```
+
+List the repositories in the OCI image layout
+
+```bash
+$ oras repo list --oci-layout k8s-control-plane
```
shizhMSFT: This feature is not there yet. See also #1770

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
+```bash
+$ oras backup registry.k8s.io/kube-apiserver registry.k8s.io/kube-controller-manager --output k8s-control-plane
+```
+
+List the repositories in the OCI image layout
+
+```bash
+$ oras repo list --oci-layout k8s-control-plane 
+kube-apiserver
+kube-controller-manager
```
shizhMSFT:
```suggestion
registry.k8s.io/kube-apiserver
registry.k8s.io/kube-controller-manager
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
+```
+
+Restore them to two repositories in a registry
+
+```bash
```
shizhMSFT:
```suggestion
```console
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
...
+$ oras restore localhost:5000/kube-apiserver localhost:5000/kube-controller --input k8s-control-plane
```
shizhMSFT: We can iterate on the output of `oras backup/restore` in the next PR.

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,291 @@
-# ORAS Backup and Restore Commands
```
Wwwsylvia, line 1: nit: **I**mages?

docs/proposals/backup-restore.md
Wwwsylvia: nit: Can we put the `--output` flag first in the list as it is required?

docs/proposals/backup-restore.md
Wwwsylvia: Do you mean something like: "Users can specify the `--exclude-referrers` flag to skip linked referrers when restoring the backup" ?

docs/proposals/backup-restore.md
FeynmanZhou: I changed to "It's inefficient to persist multiple artifacts in OCI layout format via `oras copy`"

docs/proposals/backup-restore.md
FeynmanZhou: Good catch. The `--exclude-referrers` statement should be moved to `oras restore`.

docs/proposals/backup-restore.md
FeynmanZhou: This is a desired end-to-end user experience. As we already planning to support it in https://github.com/oras-project/oras/issues/1770, does it make sense to describe it in the spec?

docs/proposals/backup-restore.md
FeynmanZhou: Updated

docs/proposals/backup-restore.md
```diff
@@ -123,7 +123,7 @@ Lack of clarity and built-in commands for standardized, reliable image backup an
 
 * `docker save/load` supports exporting and importing images but not referrers or OCI artifacts.
 * `oras pull/push` handles single artifacts, but not repository-level operations.
-* There is no built-in way to persist multiple artifacts in OCI layout format via `oras`.
+* It's inefficient to persist multiple artifacts in OCI layout format via `oras copy`.
```
shizhMSFT: It's still efficient using `oras copy`. It's just error-prone for users themselves to writing scripts to do so.

docs/proposals/backup-restore.md
shizhMSFT: I think we should remove it in this spec.

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,289 @@
...
+
+```console
+$ oras repo list --oci-layout k8s-control-plane 
+registry.k8s.io/kube-apiserver
+registry.k8s.io/kube-controller-manager
+```
+
+Restore them to two repositories in a registry
+
+```bash
```
shizhMSFT:
```suggestion
```console
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,289 @@
...
+$ oras repo tags registry-b.k8s.io/kube-apiserver
+v1
+v2
+```
+
+**Backup and Restore Multiple Repositories**
+
+Backup multiple repositories from a registry to a local OCI image layout
+
+```bash
```
shizhMSFT, line 208:
```suggestion
```console
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,289 @@
...
+
+#### Command: `oras backup`
+
+**Short summary:**
+Backup OCI artifacts and repositories from a registry into a structured, portable OCI image layout or archive tarball file locally.
+
+**Syntax:**
 ```bash
-oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+oras backup [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
TerryHowe:
```suggestion
oras backup  --output ./mirror.tgz [optional-flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
The output option is required

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,289 @@
...
 ```
 
-If the output specified is an existing directory, the output will be written in that directory in OCI layout format.
-If the output specified is an existing file, it will be overwritten.
-If the output specified is neither a file or a directory, file output is assumed.
-The file name does NOT need to end in `tgz`, but file output will be compressed tar file.
+**Output:**
+An OCI image layout directory or `.tar` archive containing the images, artifacts, their metadata, and optional referrers.
```
TerryHowe, line 42:
```suggestion
An OCI image layout directory or `.tgz` archive containing the images, artifacts, their metadata, and optional referrers.
```

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,289 @@
...
+#### Command: `oras restore`
 
-An example of restoring from a directory:
+**Short summary:**
+Restore OCI artifacts or images from a local OCI image layout or archive into a registry.
 
+**Syntax:**
 ```bash
-oras restore --input ./mirror localhost:15000/my-mirror
+oras restore [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
TerryHowe: the restore comman only specifies the destination like `localhost:15000/my-mirror` We could add a feature to filter the input for certain things, but that is another topic

docs/proposals/backup-restore.md
```diff
@@ -1,154 +1,289 @@
...
-localhost:15000/my-mirror/kube-controller-manager-arm64:v1.31.0
+oras backup registry-a.k8s.io/kube-apiserver:v1 --include-referrers --output airgap-snapshot.tar
 ```
 
-A namespace in the registry will be optional.
-The registry in the above example could be specified as `localhost:15000`.
+Transfer the `.tar` file to the air-gapped system via a secured channel. Restore the tarball from local to another registry:
+
+```bash
+oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
```
TerryHowe:
```suggestion
oras restore my-registry.localdomain --input airgap-snapshot.tar
```

You don't specify the object you are restoring and this is a bit odd of an example. My example isn't great here, but it should be some local domain on a private network for this to make sense

docs/proposals/backup-restore.md
FeynmanZhou: Without this step, I think it would be hard to inspect what have been stored in the OCI image layout after `oras backup`. My intention is to show how a desired end-to-end looks like in the proposal doc. How about we state that this will be supported in https://github.com/oras-project/oras/issues/1770 in  v1.4.0?

TerryHowe: The tags will contain the contents for example:

```
oras repo tags --oci-layout ./mirror
./mirror
registry.k8s.io/kube-apiserver-arm64:v1.31.0
registry.k8s.io/kube-controller-manager-arm64:v1.31.0
```

docs/proposals/backup-restore.md
FeynmanZhou: @TerryHowe  Yes, `--output` option is required as mentioned in the [new flag part below](https://github.com/oras-project/oras/pull/1767/files#diff-1c2a449182a03a4abeb6ccb6c5a77b5527b69df26109c4865d41e83be4b83bfaR149). I think it should be flexible to use `[flags]` in the Syntax. There might be other required options (flags) in the future so no need to limit to `--outpu` only. Note that this line is to define the syntax not the example.

docs/proposals/backup-restore.md
FeynmanZhou: @TerryHowe  OCI and Docker ecosystem uses `.tar`. Tools like docker save/load rely on `.tar` formats. 
Would it be more common and benefitial for users if we make `oras backup/restore` supports `.tar` by default? If users want to compress the tarball into `.tgz` or `.tar.gz`, they could use `gzip` on demand.

docs/proposals/backup-restore.md
FeynmanZhou: @TerryHowe  As this example is to demonstrate "Backup multiple repositories from a registry to a local OCI image layout", `oras repo list` would be more approriate than `oras repo tags` in this example.

docs/proposals/backup-restore.md
FeynmanZhou: If `oras` only requires a `registry` domain to restore, how can users restore to multiple repositories?

docs/proposals/backup-restore.md
TerryHowe:
```suggestion
$ oras repo tags --oci-layout ./k8s-control-plane 
```
So people know it is a directory

docs/proposals/backup-restore.md
TerryHowe: I haven't used a tar utility in go, but in other languages, compressing it is just an option. It might worth looking at the level of effort because for something like this, compressing makes sense.

docs/proposals/backup-restore.md
TerryHowe: The square brackets on `[flags]` means they are optional

docs/proposals/backup-restore.md
```diff
@@ -22,7 +22,7 @@ Yet today, developers and users resort to fragmented, CLI tools like:
 
 * `docker save/load` for container images.
 * `oras pull/push` for OCI artifacts.
-* `oras copy` for copying a single image with artifacts
+* `oras copy` for copying a single image with artifacts.
```
Wwwsylvia: Is "image with artifacts" indended? Or do you mean "artifact with referrers" or "image with referrers"?

FeynmanZhou: I meant images with referrers. Updated in the scenario doc: https://github.com/oras-project/oras/pull/1778

docs/proposals/backup-restore.md
FeynmanZhou: @TerryHowe I removed square brackets and highlighted `--output` is a madatory flag.

docs/proposals/backup-restore.md
FeynmanZhou: Per our discussion in the community meeting on July 8, 2025, the consensus is to support  `.tar` as the default format for `oras backup` and `oras restore` since OCI and Docker ecosystem uses `.tar`.

docs/proposals/backup-restore.md
FeynmanZhou: Updated.

FeynmanZhou:
Hi @TerryHowe @sabre1041 @sajayanto @shizhMSFT @Wwwsylvia , @oras-project/oras-cli ,

I've resolved the conflicts and updated the proposal doc based on our community meeting discussion on July 8, 2025.

- Changed to use `.tar` archive as the default format rather than `.tgz` since OCI and Docker ecosystem uses `.tar` archive.
- Moved the scenarios to another PR https://github.com/oras-project/oras/pull/1778. This doc focuses only on the CLI design improvement
- Add successful output to each command
- Moved backup and restore of multi-repo experience to the next iteration (v1.4.0-alpha.1). It will be tracked in a separate issue

Would you mind reviewing it again? Thanks

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,258 @@
...
 ```
 
-The backup command will also have the ability to write output to a new compressed tar file where the contents are in a single oci-layout
-format. For example:
+Upon success, the output will be:
 
-```bash
-oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+```console
+Pulled 1 descriptor(s) from registry-a.k8s.io/kube-apiserver:v1
```
Wwwsylvia: What does "Pulled descriptors" mean here?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,258 @@
...
+Found 1 linked referrer(s)
+Included referrers in backup: application/vnd.cncf.notary.signature
```
Wwwsylvia: What if there are multiple referrers?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,258 @@
...
 
-### oras restore
+```console
+oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
+```
+
+Upon success, the output will be:
+
+```console
+Loaded 2 artifact(s) from airgap-snapshot.tar
```
Wwwsylvia: How does the "artifact" count here? The number of manifests?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,258 @@
...
+Pushed image to registry-b.k8s.io/kube-apiserver:v1
+Pushed linked referrer: sha256:78833f9c870...
```
Wwwsylvia: Do we want to output the artifact type of the referrer as well? Something like:
> Pushed referrer: application/vnd.cncf.notary.signature: sha256:78833f9c870...

Same for the backup command

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,258 @@
...
+registry-b.k8s.io/kube-apiserver@sha256:9081a6f83f4febf47369fc46b6f0f7683c7db243df5b43fc9defe51b0471a950
+└── application/vnd.cncf.notary.signature
+    └── sha256:78833f9c870d3b069cdd998cae33b935629399f24743e680ab3bebb90de76589
+        └── [annotations]
+            ├── org.opencontainers.image.created: "2025-06-10T20:25:53Z"
+            └── io.cncf.notary.x509chain.thumbprint#S256: '["xxxxxx"]'
+```
 
-#### Restore from a directory
+#### Backup Artifacts to a Directory and Restore to Another Registry
```
Wwwsylvia: The previous example also restores repositories to another registry

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,258 @@
...
 
-An example of restoring from a directory:
+Backing up multiple artifacts to a directory:
+
+```console
+oras backup --output ./mirror registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
```
Wwwsylvia: Do we support backing up artifacts from different registries? If so, we should show it in the example.

FeynmanZhou: Backing up artifacts from different registries is not a common scenario so far and it is not listed in the scenario doc https://github.com/oras-project/oras/pull/1778. I don't think it's required for the 1st iteration right now. I am opening to hear suggestions and requirements to see if we could support it in future iterations.

docs/proposals/backup-restore.md
FeynmanZhou: These two examples are a bit different. The first example focuses on backing up a single image with an referrer to a tarball, while the second example focuses on backing up two artifacts with referrers from different repositories and write to a directory.

docs/proposals/backup-restore.md
FeynmanZhou: Good point. Added the artifact type to each referrer digest

docs/proposals/backup-restore.md
FeynmanZhou: It's the number of manifest. I feel it would be more accurate to use the term `manifest` rather than artifact in the output as it represents not only a single image manifest, but also an image index (manifest list).

docs/proposals/backup-restore.md
FeynmanZhou: I included the sample output for multiple referrers in the third example.

docs/proposals/backup-restore.md
FeynmanZhou: I changed to `manifest`. I feel it would be more accurate to use the term `manifest` in the output as it represents not only a single image manifest, but also an image index (manifest list). What do you think?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
 
-* https://github.com/oras-project/oras/issues/1366
-* https://github.com/oras-project/oras/issues/730
+The [scenario doc](https://github.com/oras-project/oras/pull/1778) illustrates real-world scenarios highlighting these challenges and how unified, structured backup and restore functionality built into `oras` can significantly improve user experience, operational efficiency, and supply chain security.
 
+## Existing Solutions
 
-## Scenarios 
+* `docker save/load` supports exporting and importing images but not referrers or OCI artifacts.
+* `oras pull/push` handles single artifacts, but not repository-level operations.
```
shizhMSFT, line 22: note: Also, `oras pull/push` does not handle those custom artifacts where the artifacts are not named like helm ones.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+* Users can write scripts to persist multiple artifacts and repositories in local OCI layout and distribute to registries via `oras copy`, but it's error-prone for users to do so.
 
-This feature will be useful for mirroring, air gapped registries, and disaster recovery.
+## Proposal
 
-Users often need to mirror registries for performance and reliability reasons.
-Pulling images over the Internet can be significantly slower than pulling images from a local registry.
-If network connections are unreliable, a local registry will potentially be a lot more reliable.
+This document proposes two new command sets, `oras backup` and `oras restore`, to address the identified problems and support the scenarios outlined above. It also describes the desired user experience for backing up and restoring artifacts, images, and repositories between a registry and the local environment. This proposal meets user expectations of portability, structure, and artifact completeness using OCI specifications.
```
shizhMSFT: They are commands, not command sets. A command set is a group of commands such as `oras repo`.
```suggestion
This document proposes two new commands, `oras backup` and `oras restore`, to address the identified problems and support the scenarios outlined above. It also describes the desired user experience for backing up and restoring artifacts, images, and repositories between a registry and the local environment. This proposal meets user expectations of portability, structure, and artifact completeness using OCI specifications.
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
 
-This document outlines various scenarios related to backing up and restoring OCI images from registries and local files. It covers a common workflow for downloading and uploading one or more image resources to and from disk, identifies limitations and challenges in existing solutions, and presents proposals for improvements to enhance usability and portability of managing OCI images.
+By providing native support for comprehensive backup and restore workflows, this enhancement improves user experience, simplifies operational tooling, and ensures that all artifacts including linked referrers are preserved with integrity and fidelity according to OCI specifications.
 
+## Problem Statement & Motivation
 
-## Problem Statement & Motivation 
+Problem statement and motivation are documented in the [scenario doc](https://github.com/oras-project/oras/pull/1778).
```
shizhMSFT: After #1778 is merged, the link needs to be updated.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
-It may be significantly easier to restore a registry from a backup rather than recreate the artifacts.
+#### Command: `oras backup`
 
+**Short summary:**
+Backup OCI artifacts and repositories from a registry into a structured, portable OCI image layout or archive tarball file locally.
 
-## Existing Solutions or Expectations
+**Syntax:**
+```console
+oras backup --flags <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
shizhMSFT: typo: `--flags` --> [flags]
```suggestion
oras backup [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
 
-For example, backing up artifacts specified on the command line to a directory:
+**Short summary:**
+Restore OCI artifacts or images from a local OCI image layout or archive into a registry.
 
-```bash
-oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+**Syntax:**
+```console
+oras restore --flags <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
shizhMSFT: typo: `--flags` --> [flags]
```suggestion
oras restore [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
-The backup and restore commands will also help users wanting to copy artifacts from a registry for disaster recovery.
-It may be significantly easier to restore a registry from a backup rather than recreate the artifacts.
+#### Command: `oras backup`
 
+**Short summary:**
+Backup OCI artifacts and repositories from a registry into a structured, portable OCI image layout or archive tarball file locally.
 
-## Existing Solutions or Expectations
+**Syntax:**
+```console
```
shizhMSFT:
```suggestion
```bash
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+#### Command: `oras restore`
 
-For example, backing up artifacts specified on the command line to a directory:
+**Short summary:**
+Restore OCI artifacts or images from a local OCI image layout or archive into a registry.
 
-```bash
-oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+**Syntax:**
+```console
```
shizhMSFT:
```suggestion
```bash
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+
+### User Experience in the CLI
+
+The desired end-to-end user experience of using `oras backup` and `oras restore` to address the identified problems and support the outlined user scenarios is illustrated below.
+
+#### Offline Snapshot for Air-Gapped Environments
+
+Create a snapshot of a sample image `registry-a.k8s.io/kube-apiserver:v1` and its referrer (e.g. signature) for an air-gapped environment:
+
+```console
```
shizhMSFT:
```suggestion
```bash
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
-format. For example:
+Upon success, the output will be:
+
+```console
+Pulled 1 manifest(s) from registry-a.k8s.io/kube-apiserver:v1
+Found 1 linked referrer(s)
+Included referrers in backup: [application/vnd.cncf.notary.signature] sha256:78833f9c870...
+Exported backup to airgap-snapshot.tar
+Backup completed: 2 manifest(s) written
+```
```
shizhMSFT: Manifests or tags? What if we want to dump the entire repository? What if the tagged manifest is an index? The output here is a guesswork.

shizhMSFT: Also, the progress bar is not involved.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+
+Transfer the `.tar` file to the air-gapped system via a secured channel. Restore the tarball from local to another registry:
+
+```console
+oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
```
shizhMSFT: curious if we need a `--dry-run` mode or not.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+### User Experience in the CLI
+
+The desired end-to-end user experience of using `oras backup` and `oras restore` to address the identified problems and support the outlined user scenarios is illustrated below.
+
+#### Offline Snapshot for Air-Gapped Environments
+
+Create a snapshot of a sample image `registry-a.k8s.io/kube-apiserver:v1` and its referrer (e.g. signature) for an air-gapped environment:
+
+```console
+oras backup registry-a.k8s.io/kube-apiserver:v1 --include-referrers --output airgap-snapshot.tar
```
shizhMSFT: It is better to provide a full repo dump scenario instead of a single image dump. This is because a single image dump can be easily done by a single `oras copy` command that cannot highlight the necessity of `oras backup`.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+```console
+oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
+```
+
+Upon success, the output will be:
+
+```console
+Loaded 2 manifest(s) from airgap-snapshot.tar
+Pushed image to registry-b.k8s.io/kube-apiserver:v1
+Pushed linked referrer: [application/vnd.cncf.notary.signature] registry-b.k8s.io/kube-apiserver@sha256:78833f9c870...
```
shizhMSFT: Are they progress bar? Can they be consistent with the existing ones?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,261 @@
...
+$ oras discover registry-b.k8s.io/kube-apiserver:v1
+registry-b.k8s.io/kube-apiserver@sha256:9081a6f83f4febf47369fc46b6f0f7683c7db243df5b43fc9defe51b0471a950
+└── application/vnd.cncf.notary.signature
+    └── sha256:78833f9c870d3b069cdd998cae33b935629399f24743e680ab3bebb90de76589
+        └── [annotations]
+            ├── org.opencontainers.image.created: "2025-06-10T20:25:53Z"
+            └── io.cncf.notary.x509chain.thumbprint#S256: '["xxxxxx"]'
+```
+
+#### Backup Artifacts to a Directory and Restore to Another Registry
```
shizhMSFT: This is a multi-resp scenario. Do you want to keep it or leave it to the next iteration?

docs/proposals/backup-restore.md
shizhMSFT: I noticed that you have the section "Backup an Entire Repository to a Tarball and Restore to Another Registry" below. Do you want to bring that section to the top and remove the "Offline Snapshot for Air-Gapped Environments" section?

docs/proposals/backup-restore.md
FeynmanZhou: It's a tag in this example. For index manifest, does it make sense to print out it is an image index? I believe `oras` should be able to identify the type of manifest.

docs/proposals/backup-restore.md
FeynmanZhou: I think `--dry-run` is valuable as it enables preview before actual restore and avoid accidental overwrites in the target registry. I added as follows:

`--dry-run`: Optional. Simulate the restore process without actually uploading any artifacts to the target registry.

docs/proposals/backup-restore.md
FeynmanZhou: I moved the section "Backup an Entire Repository to a Tarball and Restore to Another Registry" to the bottom. Backup of a single is still a necessary example as it aligns with the scenario 1 and scenario 2.

docs/proposals/backup-restore.md
FeynmanZhou: Progress bar is dynamic information which is difficult to presented in the spec. We could keep it as the same format as existing progress bar design.

docs/proposals/backup-restore.md
FeynmanZhou: Multi-repo is complicated as we have to consider how to effectively mapping multiple repos when restoring them to the target registry. Suggest us moving to `v.1.4.0-alpha` and thinking thoroughly about its scenarios and CLI design. I just removed this section.

docs/proposals/backup-restore.md
FeynmanZhou: Updated to the file name.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,206 @@
...
-It may be significantly easier to restore a registry from a backup rather than recreate the artifacts.
+#### Command: `oras backup`
 
+**Short summary:**
+Backup OCI artifacts and repositories from a registry into a structured, portable OCI image layout or archive tarball file locally.
 
-## Existing Solutions or Expectations
+**Syntax:**
+```bash
+oras backup --output [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
shizhMSFT:
```suggestion
oras backup [flags] --output <path> <registry>/<repository>[:<ref1>[,<ref2>...]]
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,206 @@
...
+#### Command: `oras restore`
 
-For example, backing up artifacts specified on the command line to a directory:
+**Short summary:**
+Restore OCI artifacts or images from a local OCI image layout or archive into a registry.
 
+**Syntax:**
 ```bash
-oras backup --output ./mirror  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+oras restore --input [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]
```
shizhMSFT:
```suggestion
oras restore [flags] --input <path> <registry>/<repository>[:<ref1>[,<ref2>...]]
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,206 @@
...
+
+### User Experience in the CLI
+
+The desired end-to-end user experience of using `oras backup` and `oras restore` to address the identified problems and support the outlined user scenarios is illustrated below.
+
+#### Backup an Entire Repository to a Tarball and Restore to Another Registry
+
+Assume two tags `v1` and `v2` are stored in a repository `registry.k8s.io/kube-apiserver` and each tag has one referrer. Backup the entire repo to a tarball and restore it to another registry:
+
+```console
```
shizhMSFT:
```suggestion
```bash
```

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,206 @@
...
-```bash
-oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+```console
+Pulled 1 tag from registry-a.k8s.io/kube-apiserver:v1
+Included referrer in backup: [application/vnd.cncf.notary.signature]: sha256:a18210fdb52...
+Pulled 1 tag from registry-a.k8s.io/kube-apiserver:v2
+Included referrer in backup: [application/vnd.cyclonedx+json]: sha256:4fc21fe77e83...
+Exported backup to backup.tar
+Backup completed: 4 manifest(s) written
+```
```
shizhMSFT: The output looks strange to me.
It makes more sense like
```
Discovered 2 tag(s) in registry-a.k8s.io/kube-apiserver
- v1
- v2
<progress bar here>
Exporting to backup.tar
Successfully backed up 2 tag(s) in registry-a.k8s.io/kube-apiserver to backup.tar
```

shizhMSFT: Similarly for `oras restore`.

FeynmanZhou: Updated the output in both two commands.

docs/proposals/backup-restore.md
```diff
@@ -108,149 +109,97 @@ Artifacts are uploaded to the target registry/registries as specified.
...
-```
+#### Backup an Entire Repository to a Tarball and Restore to Another Registry
 
-Transfer the `.tar` file to the air-gapped system via a secured channel. Restore the tarball from local to another registry:
+Assume two tags `v1` and `v2` are stored in a repository `registry.k8s.io/kube-apiserver` and each tag has one referrer. Backup the entire repo to a tarball and restore it to another registry:
 
-```console
-oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
+```bash
+## Backup a repository from a registry to a local compressed tarball. All tags and their referrers will be included.
```
Wwwsylvia: The tarball is not compressed, isn't it?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,207 @@
...
+List all tags from the repo `registry-b.k8s.io/kube-apiserver`:
+
+```console
+$ oras repo tags registry-b.k8s.io/kube-apiserver
+v1
+v2
+```
 
-#### Restore from a directory
+#### Backup and Restore an Image with Referrers
```
Wwwsylvia, line 204: Would it be more helpful to demonstrate the scenario of backing up multiple artifacts by specifying tags? Such as 

`oras backup registry-a.k8s.io/kube-apiserver:v1,v2 --include-referrers --output airgap-snapshot.tar`

FeynmanZhou: Good call out. I added the example of backing up multiple artifacts by specifying tags in the section Backup an Entire Repository to a Tarball and Restore to Another Registry.

docs/proposals/backup-restore.md
FeynmanZhou: Removed `compressed`

docs/proposals/backup-restore.md
```diff
@@ -140,12 +146,12 @@ oras restore --input backup.tar registry-b.k8s.io/kube-apiserver
...
-Pushed image to registry-b.k8s.io/kube-apiserver:v1
-Pushed linked referrer to [application/vnd.cncf.notary.signature]: registry-b.k8s.io/kube-apiserver@sha256:a18210fdb52...
-Pushed image to registry-b.k8s.io/kube-apiserver:v2
-Pushed linked referrer to [application/vnd.cyclonedx+json]: registry-b.k8s.io/kube-apiserver@sha256:4fc21fe77e83...
-Restore completed successfully
+Found 2 tag(s) in backup.tar
+- v1
+- v2
+<progress bar>...
+Pushing to registry-b.k8s.io/kube-apiserver
```
shizhMSFT: nit: this line is included in the progress.

docs/proposals/backup-restore.md
```diff
@@ -183,10 +189,11 @@ oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
 Upon success, the output will be:
 
 ```console
-Loaded 2 manifest(s) from airgap-snapshot.tar
-Pushed image to registry-b.k8s.io/kube-apiserver:v1
-Pushed linked referrer: [application/vnd.cncf.notary.signature] registry-b.k8s.io/kube-apiserver@sha256:78833f9c870...
-Restore completed successfully
+Found 1 tag(s) in airgap-snapshot.tar
+v1
```
shizhMSFT:
```suggestion
- v1
```

docs/proposals/backup-restore.md
```diff
@@ -183,10 +189,11 @@ oras restore registry-b.k8s.io/kube-apiserver:v1 --input airgap-snapshot.tar
...
+<progress bar>...
+Pushing to registry-b.k8s.io/kube-apiserver
```
shizhMSFT: nit: this line is included in the progress.

FeynmanZhou: @shizhMSFT I polished the output of `oras backup` and `oras restore` with ChatGPT's assistance. Would you mind reviewing it again? Thanks

docs/proposals/backup-restore.md
```diff
@@ -121,18 +121,35 @@ oras backup --output backup.tar --include-referrers registry-a.k8s.io/kube-apise
...
+✓ Pulled  application/vnd.oci.image.layer.v1.tar+gzip                                                            28.3/28.3 MB  100.00%     3s
+  └─ sha256:b08e2ff4391ef70ca747960a731d1f21a75febbd86edc403cd1514a099615808
+✓ Pulled  application/vnd.cncf.notary.signature                                                                 1.85/1.85 MB  100.00%  890ms
+  └─ sha256:854ad9e87ce93dae54ae1699837b2c812d2f373c3fb62625ea6992efa8f023c4
+
+## <status output>
+Found 2 tag(s): v1, v2 in registry-a.k8s.io/kube-apiserver
+Pulled tag v1 and 1 referrer(s)
+Pulled tag v2 and 1 referrer(s)
+Exported to backup.tar (58.8 MB)
```
shizhMSFT, line 155: It takes time to exporting to a tarball as well as loading from a tarball. Therefore, it is better to have **Exporting** and **Exported**.

docs/proposals/backup-restore.md
```diff
@@ -121,18 +121,35 @@ oras backup --output backup.tar --include-referrers registry-a.k8s.io/kube-apise
...
 
 ```console
-Found 2 tag(s) in registry-a.k8s.io/kube-apiserver
-- v1
-- v2
-<progress bar>...
-Exporting to backup.tar
-Successfully backed up 2 tag(s) from registry-a.k8s.io/kube-apiserver to backup.tar
+## <progress_bar>
+✔ Pulled tag v1 and 1 referrer(s)
```
shizhMSFT: Is the referrer information important as its implementation cost a lot.

docs/proposals/backup-restore.md
```diff
@@ -121,18 +121,35 @@ oras backup --output backup.tar --include-referrers registry-a.k8s.io/kube-apise
...
+✔ Pulled tag v2 and 1 referrer(s)
+✓ Pulled  application/vnd.oci.image.config.v1+json                                                               2.24/2.24 KB  100.00%  353ms
+  └─ sha256:f9248aac10f2f82e0970222e36cc7b71215b88e974e001282e5cd89797a82218
+✓ Pulled  application/vnd.oci.image.layer.v1.tar+gzip                                                            28.3/28.3 MB  100.00%     3s
+  └─ sha256:b08e2ff4391ef70ca747960a731d1f21a75febbd86edc403cd1514a099615808
+✓ Pulled  application/vnd.cncf.notary.signature                                                                 1.85/1.85 MB  100.00%  890ms
+  └─ sha256:854ad9e87ce93dae54ae1699837b2c812d2f373c3fb62625ea6992efa8f023c4
+
+## <status output>
+Found 2 tag(s): v1, v2 in registry-a.k8s.io/kube-apiserver
```
shizhMSFT: This line is better to be the first line.

shizhMSFT (commented):
Note: The comment I have is for both restore and backup.

docs/proposals/backup-restore.md
FeynmanZhou: Updated

docs/proposals/backup-restore.md
FeynmanZhou: Moved to the first line in each output.

docs/proposals/backup-restore.md
FeynmanZhou: Without the referrer information, how can users know whether each referrer is backed up or restored during the file transfering?

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,265 @@
...
-registry.k8s.io/kube-apiserver-arm64:v1.31.0
-registry.k8s.io/kube-apiserver-arm64:v1.32.0
-registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+## Backup a repository from a registry to a local tarball. All tags and their referrers will be included.
+oras backup --output backup.tar --include-referrers registry-a.k8s.io/kube-apiserver
 ```
 
-The backup command will also have the ability to write output to a new compressed tar file where the contents are in a single oci-layout
-format. For example:
+Alternatively, `oras` should also enables users to choose which tags within a repository to back up. For example, back up `v1` and `v2` tags with their referrers:
```
Wwwsylvia, line 122: Do we support specifying one or more digests as well? Is it possible to do something like: 
```bash
oras backup registry-a.k8s.io/kube-apiserver:v1,@abc123,@def456
``` 
?

If so, we need to put some examples here.

shizhMSFT: I don't think we should support digests here. The `oras backup` and `oras restore` are to cover the most common scenarios and we should leave edge scenarios to `oras copy`.

docs/proposals/backup-restore.md
shizhMSFT: Just an update: its implementation won't cost a lot as we can get the info from the local copy. So... it's up to you.

shizhMSFT (approved):
LGTM

docs/proposals/backup-restore.md
FeynmanZhou: @Wwwsylvia  It makes sense to support with backup and restore tags in the 1st iteration as tags are used the most for image manifest and image index. Digest-based backup and restore could be considered in v1.4.0-alpha if we received users inquires.

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,265 @@
...
-oras backup --output ./mirror.tgz  registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
+oras backup registry-a.k8s.io/kube-apiserver:v1,v2 --include-referrers --output backup.tar
+```
+
+Upon success, the output will be:
+
+```console
+## <progress_bar>
+Found 2 tag(s): v1, v2 in registry-a.k8s.io/kube-apiserver
+✔ Pulled tag v1 and 1 referrer(s)
```
Wwwsylvia: This should be outputted after the progress bar. That is:

✓ Pulled  application/vnd.oci.image.config.v1+json                                                               2.26/2.26 KB  100.00%  447ms
  └─ sha256:45a5868eb9f1dfbce42513000964664014789a43310865b0c8461e773e9972b9
✓ Pulled  application/vnd.oci.image.layer.v1.tar+gzip                                                            25.6/25.6 MB  100.00%     4s
  └─ sha256:149362fdfa6e6a5d9f009b896da3be3172c395ba2287b57d4969f3f46e573055
✓ Pulled  application/vnd.cncf.notary.signature                                                                 1.85/1.85 MB  100.00%  898ms
  └─ sha256:9b666bc868511a0f2d33a738a9ff0bd54eb750a72a832e8b59085d22bbdbaac2

✔ Pulled tag v1 and 1 referrer(s)

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,265 @@
...
+**New Flags:**
 
-## Proposal 
+* `--output <path>`: Required. Target directory path or tar file path to write in local filesystem.
+* `--include-referrers`: Optional. Back up the image and its linked referrers (e.g., attestations, SBOMs).
 
-The backup and restore commands will support reading and writing multiple files to and from a registry.
-As well as the flags described here, the commands will support the normal set of flags to support TLS and authentication.
+> [!NOTE] 
+> The file extension determines the output format. `oras` supports `.tar` archive as the default format since OCI and Docker ecosystem uses `.tar` archive. If the output path does not include a file extension, it is assumed that the output should be a directory. When an unsupported extension such as `.zip` or `.tar.gz` is specified, `oras` should display a warning indicating that the format is not supported. In such cases, it will proceed to create a directory at the specified path instead.
```
Wwwsylvia, line 50: What if the `tar` file already exists? Do we want to silently overwrite it?

docs/proposals/backup-restore.md
FeynmanZhou: Updated.

docs/proposals/backup-restore.md
FeynmanZhou: I think `oras` should overwrite the file if it already exists, but prompt the user for confirmation before doing so.

docs/proposals/backup-restore.md
```diff
@@ -260,6 +263,42 @@ registry-b.k8s.io/kube-apiserver@sha256:9081a6f83f4febf47369fc46b6f0f7683c7db243
             └── io.cncf.notary.x509chain.thumbprint#S256: '["xxxxxx"]'
 ```
 
+Back up an image from Docker Hub to local. Upon success, the output in a non-tty mode will be:
```
shizhMSFT: The `non-tty` mode is for plaintext, which means it only allows ASCII characters. Characters likes `•`, `├─`, and `✔` (emojis) are not allowed.

docs/proposals/backup-restore.md
```diff
@@ -260,6 +263,42 @@ registry-b.k8s.io/kube-apiserver@sha256:9081a6f83f4febf47369fc46b6f0f7683c7db243
...
+
+```console
+✔ Found 1 tag in docker.io/bitnami/nginx: 1.29.0
+✔ Starting backup with referrers included...
+
+Pulling content:
+  • application/jose+json                         fe00f3e4bc13    Pulled
+  • application/vnd.cncf.notary.signature         44136fa355b3    Pulled
+  • application/vnd.docker.container.image.v1+json
+    ├─ 6362258f3406                                Pulled
```
shizhMSFT: It MUST show the algorithm like `sha256` and `sha512`.

shizhMSFT: Otherwise, there will be problems with crypto agility.

shizhMSFT (changes requested):

docs/proposals/backup-restore.md
FeynmanZhou: I removed the verbose output changes. ORAS maintainers aligned that we will keep the verbose output as it is now since  verbose output it's not for terminal users.

docs/proposals/backup-restore.md
FeynmanZhou: I removed the verbose output changes. ORAS maintainers aligned that we will keep the verbose output as it is now since verbose output it's not for terminal users.

shizhMSFT (approved):
LGTM

docs/proposals/backup-restore.md
```diff
@@ -129,28 +129,30 @@ Upon success, the output will be:
 ```console
 ## <progress_bar>
 Found 2 tag(s): v1, v2 in registry-a.k8s.io/kube-apiserver
```
Wwwsylvia: nit: For better readability:

```suggestion
Found 2 tag(s) in registry-a.k8s.io/kube-apiserver: v1, v2
```

Wwwsylvia (approved):
LGTM with suggestions

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,268 @@
...
+* `--password <string>`: Password for authenticating to the registry.
+* `--password-stdin`: Read password from stdin.
+* `--identity-token <string>`: Use bearer token for authentication.
+* `--identity-token-stdin`: Read identity token from stdin.
+* `--ca-file <path>`: Path to custom CA certificate file.
+* `--cert-file <path>`: Path to client TLS certificate file.
+* `--key-file <path>`: Path to client private key file.
+* `--resolve <host:port:address[:address_port]>`: Customized DNS for registry.
+* `--debug`: Output debug logs (implies `--no-tty`).
+* `--no-tty`: Disable progress bars
```
Wwwsylvia, line 70: Just realized, we also need the `--distribution-spec` flag

docs/proposals/backup-restore.md
```diff
@@ -1,131 +1,268 @@
...
+## <progress_bar>
+Found 2 tag(s): v1, v2 in registry-a.k8s.io/kube-apiserver
+
+✓ Pulled  application/vnd.oci.image.config.v1+json                                                               2.26/2.26 KB  100.00%  447ms
+  └─ sha256:45a5868eb9f1dfbce42513000964664014789a43310865b0c8461e773e9972b9
+✓ Pulled  application/vnd.oci.image.layer.v1.tar+gzip                                                            25.6/25.6 MB  100.00%     4s
+  └─ sha256:149362fdfa6e6a5d9f009b896da3be3172c395ba2287b57d4969f3f46e573055
+✓ Pulled  application/vnd.cncf.notary.signature                                                                 1.85/1.85 MB  100.00%  898ms
+  └─ sha256:9b666bc868511a0f2d33a738a9ff0bd54eb750a72a832e8b59085d22bbdbaac2
+✔ Pulled tag v1 and 1 referrer(s)
```
Wwwsylvia, line 140: How about: 

```suggestion
Backed up v1 and 1 referrer(s)
```

So that it aligns with the final state "Successfully backed up 1 tag(s) ..."

Wwwsylvia (commented):
UX

docs/proposals/backup-restore.md
FeynmanZhou: @Wwwsylvia  The `pulled` and `backed up` are different meaning semantically. I think it's more accurate to use "pulling" and "pulled" in the progress bar to indicate the operational status while downloading content from the registry, and use backed up in the operation result as a final summary. In fact, "pulling/pulled" is straightforward to reflect the content download status and easy to understand for end users.

docs/proposals/backup-restore.md
FeynmanZhou: Added `--distribution-spec` as a preview flag to `oras backup` and `oras restore`. We may want to think about whether this flag should be kept or deprecated in v1.3.0.

docs/proposals/backup-restore.md
```diff
@@ -89,6 +90,7 @@ Artifacts are uploaded to the target registry/registries as specified.
 **Common flags:**
 
 * `--concurrency <int>`: Number of parallel upload operations. Default: `3`.
+* `--distribution-spec <string>`: [Preview] set OCI distribution spec version and API option for the registry. Options: v1.1-referrers-tag, v1.1-referrers-api.
```
Wwwsylvia, line 93: This duplicates with line 107

docs/proposals/backup-restore.md
```diff
@@ -128,7 +130,7 @@ Upon success, the output will be:
 
 ```console
 ## <progress_bar>
-Found 2 tag(s): v1, v2 in registry-a.k8s.io/kube-apiserver
+Found 2 tag(s) in registry-a.k8s.io/kube-apiserver: v1, v2
```
Wwwsylvia, line 132: Just a note: I realized that if `v1` does not exist, the output would look like this:

```console
Found 2 tag(s) in registry-a.k8s.io/kube-apiserver: v1, v2
Error: registry-a.k8s.io/kube-apiserver: v1: not found
```

It might be a confusing when the output says "Found tags ..." and then the error says "not found"

FeynmanZhou: @Wwwsylvia  I assume if `v1` does not exist, `oras` should print out the existing tag only, i.e. `Found 1 tag in registry-a.k8s.io/kube-apiserver: v1`. Is it supposed to be this logic?

Wwwsylvia: You are right, I refactored the implementation. Now it works like this:

```console
oras backup mcr.microsoft.com/oss/v2/oras-project/oras:v1.2.3,not-exist` -o backup.tar
Error: failed to resolve tag "test": mcr.microsoft.com/oss/v2/oras-project/oras:not-exist: not found
```

Wwwsylvia (approved):
LGTM

PR #1768, opened 2025-07-04
feat(repo): added --limit flag to repo ls and tags   FIXES #1762

### LINKED ISSUE #1762: Add capability to limit the max number of results for `oras repo ls` and `oras repo tags`

v1.3.0-beta.3

I would like to suggest adding a flag to the `oras repo ls` and `oras repo tags` commands to limit the maximum number of results returned. This flag would allow users to specify a maximum number of results, preventing the client from being overwhelmed by too many results.

Security: Helps prevent malicious servers from dumping excessive results, which could lead to an Out Of Memory (OOM) situation.
Usability: When using `--json` format, users see nothing until the entire result is loaded and rendered. Limiting results improves performance when only the first N tags are needed.

tanyabhatnagar:
**Description**

This PR adds support for a --limit flag to both oras repo ls and oras repo tags commands.
It allows users to limit the number of repositories or tags returned, which improves performance and avoids overwhelming output — especially useful in JSON or template formats.

**What’s Added**

--limit flag in both repo ls and repo tags
Early return when the limit is reached
Validation to prevent negative values (--limit < 0)

**Example Usage**

-  Limit to 2 repositories
   ./oras repo ls localhost:5000 --limit 2
-  Limit -1 is invalid and will return an error
    ./oras repo ls localhost:5000 --limit -1
   Error: --limit must be 0 or a positive number

 **Notes**

Happy to add a test case if needed — just let me know!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,3 +30,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "docker"
+    directory: "/" 
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: ['1.24']
+        go-version: ['1.22']
       fail-fast: true
     steps:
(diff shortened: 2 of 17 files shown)


shizhMSFT:
Manually linking to #1762

PR #1769, opened 2025-07-04
feature: basic backup and restore

TerryHowe:

This PR adds a very limited backup and restore command. Backup and restore in this PR are only supported in OCI Layout format. For example:

```
oras backup --output ./mirror registry.k8s.io/kube-apiserver-arm64:v1.31.0 registry.k8s.io/kube-apiserver-arm64:v1.32.0 registry.k8s.io/kube-controller-manager-arm64:v1.31.0
```

And restore to Zot:

```
oras restore --input ./mirror localhost:15000
```


shizhMSFT:
I wrote a [bash-based prototype](https://github.com/shizhMSFT/oras-bash/blob/322f58e40f5bd284db10eeea1ade8da3979a09ca/oras-ext.sh). Hope it can give some insights.

TerryHowe:
> I wrote a [bash-based prototype](https://github.com/shizhMSFT/oras-bash/blob/322f58e40f5bd284db10eeea1ade8da3979a09ca/oras-ext.sh). Hope it can give some insights.

Yes.

The only thing you have that I was not thinking of is `# usage: oras restore [flags] <registry>/<repository>[:<ref1>[,<ref2>...]] [...]` I think that should be `# usage: oras restore [flags] <registry>[/<repository>[:<ref1>[,<ref2>...]] [...]]` adding the capability to copy everything. That would be how you'd want it to work for mirroring for example. In the specification right now, it only supports `# usage: oras restore [flags] <registry>`

PR #1773, opened 2025-07-04
chore(ux): improve copy error message

Wwwsylvia:

Fixes #

### DIFF
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -104,20 +104,17 @@ func Command(cmd *cobra.Command, handler Modifier) *cobra.Command {
 	return cmd
 }
 
-// TrimErrResp tries to trim toTrim from err.
-func TrimErrResp(err error, toTrim error) error {
-	// var inner error
-	if errResp, ok := toTrim.(*errcode.ErrorResponse); ok {
-		if len(errResp.Errors) == 0 {
-			// return fmt.Errorf("no recognizable error message found: %w", toTrim)
-			return toTrim
-		}
-		return errResp.Errors
-		// inner = errResp.Errors
-	} else {
-		return err
+// ReportErrResp returns the inner error message from errResp.Errors.
+// If errResp.Errors is empty, it returns the original errResp.
+func ReportErrResp(errResp *errcode.ErrorResponse) error {
+	if len(errResp.Errors) == 0 {
+		// Example error string:
+		// GET "registry.example.com/v2/_catalog": response status code 401: 401
+		return errResp
 	}
-	// return reWrap(err, toTrim, inner)
+	// Example error string:
+	// unauthorized: authentication required
+	return errResp.Errors
 }
 
 // TrimErrBasicCredentialNotFound trims the credentials from err.
--- a/cmd/oras/internal/errors/errors.go
+++ b/cmd/oras/internal/errors/errors.go
@@ -106,16 +106,18 @@ func Command(cmd *cobra.Command, handler Modifier) *cobra.Command {
 
 // TrimErrResp tries to trim toTrim from err.
 func TrimErrResp(err error, toTrim error) error {
-	var inner error
+	// var inner error
 	if errResp, ok := toTrim.(*errcode.ErrorResponse); ok {
 		if len(errResp.Errors) == 0 {
-			return fmt.Errorf("recognizable error message not found: %w", toTrim)
+			// return fmt.Errorf("no recognizable error message found: %w", toTrim)
+			return toTrim
 		}
-		inner = errResp.Errors
+		return errResp.Errors
+		// inner = errResp.Errors
 	} else {
 		return err
 	}
-	return reWrap(err, toTrim, inner)
+	// return reWrap(err, toTrim, inner)
 }
 
 // TrimErrBasicCredentialNotFound trims the credentials from err.
--- a/cmd/oras/internal/option/remote.go
+++ b/cmd/oras/internal/option/remote.go
@@ -407,10 +407,9 @@ func (remo *Remote) Modify(cmd *cobra.Command, err error) (error, bool) {
 	}
 
 	if errors.As(err, &errResp) {
-		// TODO: handle this
 		cmd.SetErrPrefix(oerrors.RegistryErrorPrefix)
 		return &oerrors.Error{
-			Err: oerrors.TrimErrResp(err, errResp),
+			Err: oerrors.ReportErrResp(errResp),
 		}, true
 	}
 	return err, false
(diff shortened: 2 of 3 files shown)


cmd/oras/internal/errors/errors.go
```diff
@@ -85,37 +86,47 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
 	}
 }
 
-// Modifier modifies the error during cmd execution.
-type Modifier interface {
-	Modify(cmd *cobra.Command, err error) (modifiedErr error, modified bool)
+// ErrModifier modifies the error during cmd execution.
+type ErrModifier interface {
```
shizhMSFT: The name `ErrModifier` looks like an `error` type instead of an `interface` type. You may consider `ErrorModifier` as an alternative.

shizhMSFT: Then `ModifyErr` will become `ModifyError`.

shizhMSFT (approved):
LGTM with suggestions

cmd/oras/internal/errors/errors.go
Wwwsylvia: Sounds good. Updated.

cmd/oras/internal/errors/errors.go
```diff
@@ -85,37 +86,47 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
 	}
 }
 
-// Modifier modifies the error during cmd execution.
-type Modifier interface {
-	Modify(cmd *cobra.Command, err error) (modifiedErr error, modified bool)
+// ErrorModifier modifies the error during cmd execution.
+type ErrorModifier interface {
+	ModifyError(cmd *cobra.Command, err error, canSetPrefix bool) (modifiedErr error, modified bool)
```
qweeah: Would Modifier be a better interface name? According to [Effective Go](https://go.dev/doc/effective_go#:~:text=exported%20names%20in%20the%20package%20can%20use%20that%20fact%20to%20avoid%20repetition.) `exported names in the package can use that fact to avoid repetition.`

cmd/oras/internal/errors/errors.go
```diff
@@ -85,37 +86,47 @@ func CheckArgs(checker func(args []string) (bool, string), Usage string) cobra.P
...
 
 // Command returns an error-handled cobra command.
-func Command(cmd *cobra.Command, handler Modifier) *cobra.Command {
+func Command(cmd *cobra.Command, handler ErrorModifier) *cobra.Command {
 	runE := cmd.RunE
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
 		err := runE(cmd, args)
 		if err != nil {
-			err, _ = handler.Modify(cmd, err)
+			err, _ = handler.ModifyError(cmd, err, true)
```
qweeah: Why `canSetPrefix` is needed when it's always true?

cmd/oras/internal/option/binary_target.go
```diff
@@ -64,9 +67,40 @@ func (target *BinaryTarget) Parse(cmd *cobra.Command) error {
...
+func (target *BinaryTarget) ModifyError(cmd *cobra.Command, err error, canSetPrefix bool) (error, bool) {
+	var copyErr *oras.CopyError
+	if !errors.As(err, &copyErr) {
+		return target.modifyError(cmd, err, canSetPrefix)
+	}
+
+	err = copyErr.Err // extract inner error
+	if !canSetPrefix {
+		err, _ := target.modifyError(cmd, err, canSetPrefix)
+		return err, true
```
qweeah: This block can never be hit

cmd/oras/internal/option/binary_target.go
```diff
@@ -64,9 +67,40 @@ func (target *BinaryTarget) Parse(cmd *cobra.Command) error {
...
+	}
+
+	// Example: Error from source registry for "localhost:5000/test:v1":
+	// Example: Error from destination oci-layout for "oci-dir:v1":
+	cmd.SetErrPrefix(fmt.Sprintf("Error from %s %s for %q:", copyErr.Origin, errTarget.Type, errTarget.RawReference))
+	canSetPrefix = false // do not set prefix again
+	err, _ = target.modifyError(cmd, err, canSetPrefix)
+	return err, true
```
qweeah: Why not removing `canSetPrefix` and set error prefix after modifying?

Like
```golang
	err, _ = target.modifyError(cmd, err)
	cmd.SetErrPrefix(fmt.Sprintf("Error from %s %s for %q:", copyErr.Origin, errTarget.Type, errTarget.RawReference))
```

cmd/oras/internal/errors/errors.go
Wwwsylvia: Makes sense. Updated to `Modifier` but keep `ModifyError` as it will be used like `handler.ModifyError`.

cmd/oras/internal/option/binary_target.go
Wwwsylvia: Looks good. Updated the related bits.

cmd/oras/root/pull.go
```diff
@@ -167,8 +166,12 @@ func runPull(cmd *cobra.Command, opts *pullOptions) (pullError error) {
 
 	desc, err := doPull(ctx, src, dst, copyOptions, metadataHandler, statusHandler, opts)
 	if err != nil {
+		err = oerrors.UnwrapCopyError(err)
```
qweeah: Should this be called right after using `oras.Copy` in `doPull`?

cmd/oras/root/attach.go
```diff
@@ -213,7 +213,7 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
 	// Attach
 	root, err := doPush(dst, stopTrack, pack, copy)
 	if err != nil {
-		return err
+		return oerrors.UnwrapCopyError(err)
```
qweeah: Should `UnwrapCopyError` be used right after `oras.CopyGraph`?

cmd/oras/root/push.go
```diff
@@ -272,7 +272,7 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 	// Push
 	root, err := doPush(dst, stopTrack, pack, copyWithScopeHint)
 	if err != nil {
-		return err
+		return oerrors.UnwrapCopyError(err)
```
qweeah: Should this be used within `copyWithScopeHint`?

qweeah (approved):
LGTM with minor suggestions

cmd/oras/root/pull.go
Wwwsylvia: Good catch, updated.

cmd/oras/root/pull.go
```diff
@@ -167,8 +166,12 @@ func runPull(cmd *cobra.Command, opts *pullOptions) (pullError error) {
...
 		if errors.Is(err, file.ErrPathTraversalDisallowed) {
```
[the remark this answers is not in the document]
Wwwsylvia: This is no applicable

shizhMSFT (approved):
LGTM

PR #1777, opened 2025-07-08
feat: add `--preserve-permissions` to `pull`

### LINKED ISSUE #1776: Add a --preserve-permissions flag for pull/push commands.

1.3.0-beta.3+unreleased

In this [issue](https://github.com/oras-project/oras-go/issues/886) and subsequent [PR](https://github.com/oras-project/oras-go/pull/891), @sammy-da kindly added support for file.Store to correctly unpack file permissions.

I would like this to be added as a CLI option on the pull/push sub-commands (as appropriate).

Ive attempted to add this in a fork by doing the following: https://github.com/Chronojam/oras/pull/1/files, and while this compiles, it does not do what id expect. 

I think the issue is part of the push command, I also attempted to add the  `PreservePermissions` field in the file.Store in push.go.

I suspect that the manipulation with the copies around memory.Store might be to blame, but if im being honest I dont truly understand enough about exactly is happening here to isolate it.

My tested with the following:

``` bash
touch foo.sh
chmod 777 foo.sh
ls -lrt foo.sh
-rwxrwxrwx 1 calliope users 0 Jul  7 16:40 foo.sh

./oras push --preserve-permissions localhost:5000/test-script:latest foo.sh 
rm foo.sh
./oras pull --preserve-permissions localhost:5000/test-script:latest foo.sh

ls -lrt foo.sh
-rw-r--r--  1 calliope users        0 Jul  7 16:42 foo.sh
```

Hopefully thats enough detail, happy to answer questions if there are any

This is a nice QOL feature for the CLI, and would be a nice addition to solve workarounds like storing file metadata and doing file-permission repair commands after a pull.

- [x] Yes, I am willing to implement it.

Chronojam:

This adds the `--preserve-permissions` flag to the `pull` sub-command.

Fixes #1776

  Yes, although my approach might be a bit heavy-handed, feedback welcome.
  Yes, we'll want to update the flag list on the pull command docs page
  No, the default behavior is false, so existing scripts and code will continue to function as-is


FeynmanZhou:
Hi @Chronojam , thanks for your contribution. Would you mind signing off your git commits using the `-s` flag (i.e., `git commit -s`)?

cmd/oras/root/cp_test.go
```diff
@@ -120,6 +121,13 @@ func TestMain(m *testing.M) {
 	defer ts.Close()
 	uri, _ := url.Parse(ts.URL)
 	host = "localhost:" + uri.Port()
+
+	// genericRegistry is a test fixture designed to function
+	// exactly like a normal registry would
+	genericRegistry := NewMockOCIRegistry()
+	defer genericRegistry.Close()
+	genericUri, _ := url.Parse(genericRegistry.URL())
+	genericHost = "localhost:" + genericUri.Port()
```
Wwwsylvia: Is this related to the PR?

cmd/oras/root/pull.go
```diff
@@ -120,6 +124,7 @@ Example - Pull artifact files tagged 'example.com:v1' from an OCI image layout f
 	cmd.Flags().BoolVarP(&opts.KeepOldFiles, "keep-old-files", "k", false, "do not replace existing files when pulling, treat them as errors")
 	cmd.Flags().BoolVarP(&opts.PathTraversal, "allow-path-traversal", "T", false, "allow storing files out of the output directory")
 	cmd.Flags().BoolVarP(&opts.IncludeSubject, "include-subject", "", false, "[Preview] recursively pull the subject of artifacts")
+	cmd.Flags().BoolVarP(&opts.PreservePermissions, "preserve-permissions", "", false, "preserve permissions when pulling artifacts from registry.")
```
Wwwsylvia: We need make it clearer on the effect of the flag. How about something like "Preserve original file permissions when extracting tar archives, ignoring the current umask."?

cmd/oras/root/pull.go
```diff
@@ -70,6 +71,9 @@ Example - Pull artifact files from a registry:
 Example - Recursively pulling all files from a registry, including subjects of hello:v1:
   oras pull --include-subject localhost:5000/hello:v1
 
+Example - Pull artifact files from a registry, preserving file permissions:
+  oras pull --preserve-permissions -o out localhost:5000/hello:v1
```
Wwwsylvia: How about:

Example - Pull files from a directory-based artifact in a registry, preserving original permissions
oras pull --preserve-permissions -o out localhost:5000/hello:v1

```suggestion
Example - Pull files from a directory-based artifact in a registry, preserving original permissions
oras pull --preserve-permissions -o out localhost:5000/hello-dir:v1
```

cmd/oras/root/pull_test.go
```diff
@@ -41,3 +44,86 @@ func Test_runPull_errType(t *testing.T) {
 		t.Fatalf("got %v, want %v", got, want)
 	}
 }
+
+func Test_doPull(t *testing.T) {
+	t.Run("PreservePermissions", func(t *testing.T) {
+		repo := "pp"
+		tagReference := "dppp"
```
Wwwsylvia: nit: The values are a bit confusing. Can we make the it a bit more meaningful?

cmd/oras/root/pull_test.go
```diff
@@ -41,3 +44,86 @@ func Test_runPull_errType(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		oldWd, err := os.Getwd()
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer func() { _ = os.Chdir(oldWd) }()
+
+		//fmt.Println(filepath.Base(srcDir))
```
Wwwsylvia: should we remove this?

cmd/oras/root/pull_test.go
```diff
@@ -41,3 +44,86 @@ func Test_runPull_errType(t *testing.T) {
...
+		subDir := filepath.Base(srcDir)
+
+		pushArgs := []string{
+			"push",
+			"--insecure",
+			"--plain-http",
+			"--no-tty",
+			"--format",
+			"go-template=\n",
```
Wwwsylvia: Do we need the formatting?

cmd/oras/root/pull_test.go
```diff
@@ -41,3 +44,86 @@ func Test_runPull_errType(t *testing.T) {
 		t.Fatalf("got %v, want %v", got, want)
 	}
 }
+
+func Test_doPull(t *testing.T) {
```
Wwwsylvia: This test might now work as expected on Windows.

Wwwsylvia: We can consider add e2e test instead of unit tests if it's easier.

Wwwsylvia (changes requested):

cmd/oras/root/cp_test.go
Chronojam: Thanks for looking at this, 

So the reason this generic server exists is because the way im testing this is basically to just fully run the push/pull commands here: https://github.com/oras-project/oras/pull/1777/files#diff-df51e0a44832011d99dd8dc06b63a89aa35891b86b36ba0256ed1f6a5d1b1953R81

The registry server [above](https://github.com/oras-project/oras/blob/f44081dc5527c275f1fcf42115d7d78e874f73ec/cmd/oras/root/cp_test.go#L77) uses precomputed values specifically for the `cp` test, which I think is a better way of handling it but I don't know how those values are computed.

cmd/oras/root/pull_test.go
Chronojam: I couldn't find a way to pass an empty value here for the template, this felt like the easiest way to squash the output when running as a test.

If you dont pass the formatting you get the full output, but maybe that's ok? Ill update this to remove it

cmd/oras/root/pull_test.go
Chronojam: Probably, as you can see it does a bunch of `os.Chdir` all over the place to temporary directories which feels like just a bad idea in a unit test. Ill see if I can move it over to an e2e test instead

cmd/oras/root/pull.go
```diff
@@ -70,6 +71,9 @@ Example - Pull artifact files from a registry:
 Example - Recursively pulling all files from a registry, including subjects of hello:v1:
   oras pull --include-subject localhost:5000/hello:v1
 
+Example - Pull files from a directory-based artifact in a registry, preserving original permissions
+oras pull --preserve-permissions -o out localhost:5000/hello-dir:v1
```
shizhMSFT: nit: formatting. Also, remove the non-relevant flag `-o`.
```suggestion
  oras pull --preserve-permissions localhost:5000/hello-dir:v1
```

cmd/oras/root/pull.go
```diff
@@ -120,6 +124,7 @@ Example - Pull artifact files tagged 'example.com:v1' from an OCI image layout f
 	cmd.Flags().BoolVarP(&opts.KeepOldFiles, "keep-old-files", "k", false, "do not replace existing files when pulling, treat them as errors")
 	cmd.Flags().BoolVarP(&opts.PathTraversal, "allow-path-traversal", "T", false, "allow storing files out of the output directory")
 	cmd.Flags().BoolVarP(&opts.IncludeSubject, "include-subject", "", false, "[Preview] recursively pull the subject of artifacts")
+	cmd.Flags().BoolVarP(&opts.PreservePermissions, "preserve-permissions", "", false, "Preserve original file permissions when extracting tar archives, ignoring the current umask.")
```
shizhMSFT: Should we mark this flag as `[Experimental]` or `[Preview]`?

Since it is already a stable feature in `oras-go`, a `[Preview]` mark should fit.

shizhMSFT (commented):
Overall, it LGTM but the tests need to be refined.

cmd/oras/root/cp_test.go
Chronojam: Ive moved the test to a much cleaner e2e one, so this is no longer needed

Chronojam:
Ive reworked the big bulky unit test into a nicer, smaller, e2e test instead.
I think that this is now ready for another review

cmd/oras/root/pull.go
```diff
@@ -70,6 +71,9 @@ Example - Pull artifact files from a registry:
 Example - Recursively pulling all files from a registry, including subjects of hello:v1:
   oras pull --include-subject localhost:5000/hello:v1
 
+Example - Pull files from a directory-based artifact in a registry, preserving original permissions
```
Wwwsylvia: We need a `[Preview]` label here as well. Besides, since this feature is in preview, should we consider moving it below the stable examples?"

test/e2e/internal/testdata/foobar/const.go
```diff
@@ -22,9 +22,11 @@ import (
...
 
 var (
-	Tag              = "foobar"
-	Digest           = "sha256:fd6ed2f36b5465244d5dc86cb4e7df0ab8a9d24adc57825099f522fe009a22bb"
-	ManifestStateKey = match.StateKey{Digest: "fd6ed2f36b54", Name: "application/vnd.oci.image.manifest.v1+json"}
+	Tag                = "foobar"
+	Digest             = "sha256:fd6ed2f36b5465244d5dc86cb4e7df0ab8a9d24adc57825099f522fe009a22bb"
+	ManifestStateKey   = match.StateKey{Digest: "fd6ed2f36b54", Name: "application/vnd.oci.image.manifest.v1+json"}
+	DirectoryName      = "foobar"
+	PermissionFileName = "foobar/baz"
```
Wwwsylvia: nit: Can we make the file names a bit more self-explaining? Such as:
```suggestion
	PermissionDirectoryName      = "test-permission-dir"
	PermissionFileName = "test-permission-file"
```

test/e2e/suite/command/pull.go
```diff
@@ -200,6 +200,28 @@ var _ = Describe("ORAS beginners:", func() {
...
+			// prepare
+			pushRoot := GinkgoT().TempDir()
+			tag := "pushed_preserve_permissions"
+			ref := RegistryRef(ZOTHost, ArtifactRepo, tag)
+			Expect(CopyTestFiles(pushRoot)).ShouldNot(HaveOccurred())
+			ORAS("push", ref, foobar.DirectoryName).WithWorkDir(pushRoot).Exec()
+
+			// test
+			pullRoot := GinkgoT().TempDir()
+			fmt.Println(LayoutRef(pushRoot, tag))
```
Wwwsylvia: We usually don't print stuffs in tests

test/e2e/suite/command/pull.go
```diff
@@ -200,6 +200,28 @@ var _ = Describe("ORAS beginners:", func() {
 				ExpectFailure().
 				MatchErrKeyWords("Error: ", "--allow-path-traversal").Exec()
 		})
+
+		It("should preserve permissions", func() {
```
Wwwsylvia: This test should be moved into the "OCI spec 1.1 registry users:" section and the "OCI image layout users:" section.

Wwwsylvia: It'd be nice if we also test the case where `--preserve-permissions` is set to `false` and the permissions are not preserved.

test/e2e/internal/testdata/foobar/const.go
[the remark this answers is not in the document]
Wwwsylvia: This is a valid comment.

test/e2e/internal/utils/prepare.go
```diff
@@ -115,5 +115,12 @@ func copyFile(srcFile, dstFile string) error {
 	defer from.Close()
 
 	_, err = io.Copy(to, from)
+	if err != nil {
+		return err
+	}
+
+	// Also preserve file permissions.
+	srcStat, _ := from.Stat()
```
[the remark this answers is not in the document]
Wwwsylvia: Agreed.

test/e2e/internal/utils/prepare.go
```diff
@@ -115,5 +115,12 @@ func copyFile(srcFile, dstFile string) error {
...
+	err = os.Chmod(dstFile, srcStat.Mode())
```
Wwwsylvia, line 127: Do we need to do this in the prepare function? Since we have any pre-defined original permissions, we can `chmod` to desired permissions in the actual tests after the original files are copied to the target directory. e.g. `pushRoot`.

Chronojam: Ive added two new testdata files
```
$ tree test-permission-dir/
test-permission-dir/
├── test-no-permission-file
└── test-permission-file

1 directory, 2 files
```

```
$ stat -c "%n: %a" *
test-no-permission-file: 644
test-permission-file: 755
```

I think its maybe better to have these as static testdata rather than create them as part of the test, which makes this changing of permissions as part of the copy operation needed (otherwise they are just pushed with defaults), happy to be convinced otherwise, however

cmd/oras/root/pull_test.go
Wwwsylvia, line 1: Need to clean up the imports

test/e2e/internal/utils/prepare.go
Wwwsylvia: Sounds good.

test/e2e/internal/testdata/foobar/const.go
```diff
@@ -16,15 +16,19 @@ limitations under the License.
...
 
 var (
-	Tag              = "foobar"
-	Digest           = "sha256:fd6ed2f36b5465244d5dc86cb4e7df0ab8a9d24adc57825099f522fe009a22bb"
-	ManifestStateKey = match.StateKey{Digest: "fd6ed2f36b54", Name: "application/vnd.oci.image.manifest.v1+json"}
+	Tag                     = "foobar"
+	Digest                  = "sha256:fd6ed2f36b5465244d5dc86cb4e7df0ab8a9d24adc57825099f522fe009a22bb"
+	ManifestStateKey        = match.StateKey{Digest: "fd6ed2f36b54", Name: "application/vnd.oci.image.manifest.v1+json"}
+	PermissionDirectoryName = "test-permission-dir"
+	PermissionFileName      = filepath.Join(PermissionDirectoryName, "test-permission-file")
```
Wwwsylvia, line 31: For future maintainability, can we add small comment to mention the permissions of the file? Something like:

```suggestion
	PermissionFileName      = filepath.Join(PermissionDirectoryName, "test-permission-file") // permissions: 777
```

test/e2e/suite/command/pull.go
```diff
@@ -437,6 +477,46 @@ var _ = Describe("OCI image layout users:", func() {
...
+			ORAS("push", ref, foobar.PermissionDirectoryName).WithWorkDir(pushRoot).Exec()
+
+			// test
+			pullRoot := GinkgoT().TempDir()
+			ORAS("pull", ref, "-o", pullRoot).
+				WithDescription(pullRoot).Exec()
+
+			have, err := os.Stat(filepath.Join(pullRoot, foobar.PermissionFileName))
+			Expect(err).NotTo(HaveOccurred())
+			want, err := os.Stat(filepath.Join(pushRoot, foobar.NoPermissionFileName))
```
Wwwsylvia: It looks like we don't really need the `test-no-permission-file`? We can compare the same file in the pull root and push root like we did above, and their permissions should be different.

Chronojam:
Im closing this PR, I dont think we need to set this flag at all.

While trying to figure out why a test was failing, ive realized that the process here https://github.com/oras-project/oras/issues/1776#issuecomment-3047522221 actually works as intended, and we dont need to add anything extra.

Thanks all for your support, sorry for pulling all your time looking at this 🤦🏻‍♀️

FeynmanZhou:
@Chronojam Thanks for clarifying the decision. Looking forward to your future contribution.

Wwwsylvia:
@Chronojam Sounds good! Thanks again for your contribution!

PR #1779, opened 2025-07-09
test: exclude incompatible test Go OSes

TerryHowe:

Exclude operating systems that are broken for these tests rather include operating systems we do not technically support


shizhMSFT (approved):
LGTM

PR #1790, opened 2025-07-23
feat: `oras backup` to backup artifacts to a directory or a tar

Wwwsylvia:
Add a new command `oras backup` and the corresponding tests.
Spec: https://github.com/oras-project/oras/blob/main/docs/proposals/backup-restore.md

The command looks like this:

`oras -h`:
image

`oras backup -h`:
image

`oras backup mcr.microsoft.com/oss/v2/oras-project/oras:v1.2.3 -o backup.tar`:

image

`./oras backup mcr.microsoft.com/oss/v2/oras-project/oras --include-referrers -o backup`:
image

Spec: #1767 
Resolves part of #1366


internal/io/tar_utils.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"fmt"
+	"io"
+	"os"
+)
+
+// TarDirectory creates a tar archive from the contents of sourceDir and writes it to the given writer.
+// It walks through the directory tree, adding all regular files and directory entries to the archive.
+// Symlinks, devices, pipes, and other special files are skipped.
+func TarDirectory(ctx context.Context, writer io.Writer, sourceDir string) (tarErr error) {
```
shizhMSFT: `ctx` is not used.

internal/io/tar_utils.go
```diff
@@ -0,0 +1,49 @@
...
+	// Ensure sourceDir exists and is a directory
+	fi, err := os.Stat(sourceDir)
+	if err != nil {
+		return fmt.Errorf("failed to stat source directory: %w", err)
+	}
+	if !fi.IsDir() {
+		return fmt.Errorf("source is not a directory: %s", sourceDir)
+	}
```
shizhMSFT, line 34: Is this necessary? It should be covered by `tw.AddFS(os.DirFS(sourceDir))`?

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+			}
+			defer func() {
+				stopErr := statusHandler.StopTracking()
+				if retErr == nil {
+					retErr = stopErr
+				}
+			}()
+
+			return backupTag(ctx, srcRepo, trackedDst, roots[i], t, opts.includeReferrers, copyGraphOpts, extCopyGraphOpts)
+		}(t)
```
shizhMSFT: I understand you need an anonymous function to scope the defer func. However, I think it is unnecessary to take any parameters like `(tag string)`.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+			return 0, fmt.Errorf("failed to tag %q with %q: %w", root.Digest.String(), tag, err)
+		}
+		return 0, nil
+	}
+
+	// copy with referrers
+	if err := recursiveCopy(ctx, src, dst, tag, root, extCopyGraphOpts); err != nil {
+		return 0, fmt.Errorf("failed to pull tag %q and referrers, digest %q: %w", tag, root.Digest.String(), err)
+	}
+	referrers, err := registry.Referrers(ctx, dst, root, "")
```
shizhMSFT: Does it work with index?

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func backupTag(ctx context.Context,
+	src oras.ReadOnlyGraphTarget,
+	dst oras.GraphTarget,
+	root ocispec.Descriptor,
+	tag string,
+	includeReferrers bool,
+	copyGraphOpts oras.CopyGraphOptions,
+	extCopyGraphOpts oras.ExtendedCopyGraphOptions) (int, error) {
```
shizhMSFT: It looks like two functions `backupTag` and `backupTagWithReferrers`.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func prepareBackupOutput(ctx context.Context, dstRoot string, opts *backupOptions, logger logrus.FieldLogger, metadataHandler metadata.BackupHandler) error {
+	// Remove ingest dir for a cleaner output
+	ingestDir := filepath.Join(dstRoot, "ingest")
+	if _, err := os.Stat(ingestDir); err == nil {
+		if err := os.RemoveAll(ingestDir); err != nil {
+			logger.Debugf("failed to remove ingest directory: %v", err)
+		}
+	}
```
shizhMSFT: The `ingest` folder is likely to be there. Therefore, you can try `os.RemoveAll` first. If the error is "not exist", we can ignore it.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+		return 0, fmt.Errorf("failed to pull tag %q and referrers, digest %q: %w", tag, root.Digest.String(), err)
+	}
+	referrers, err := registry.Referrers(ctx, dst, root, "")
+	if err != nil {
+		return 0, fmt.Errorf("failed to get referrers for tag %q, digest %q: %w", tag, root.Digest.String(), err)
+	}
+	return len(referrers), nil
+}
+
+func prepareBackupOutput(ctx context.Context, dstRoot string, opts *backupOptions, logger logrus.FieldLogger, metadataHandler metadata.BackupHandler) error {
```
shizhMSFT: To me, `postBackup` sounds better than `prepareBackupOutput`.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+		if removeErr != nil {
+			logger.Debugf("failed to remove temporary tar file %s: %v", tempTarPath, removeErr)
+		}
+		return err
+	}
+
+	fi, err := os.Stat(absOutput)
+	if err != nil {
+		return fmt.Errorf("failed to stat output file %s: %w", absOutput, err)
+	}
```
shizhMSFT: What's the reason to create a tar file in a temp place then move to the desired folder?

In this scenario, we don't have race condition. I don't think we need to trade off the perf for the race condition scenarios.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,390 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get absolute path for output file %s: %w", opts.output, err)
+		}
+	}
+	if err := os.MkdirAll(filepath.Dir(absOutput), 0755); err != nil {
+		return fmt.Errorf("failed to create directory for output file %s: %w", absOutput, err)
+	}
+
+	// Move the temporary tar file to the final output path
+	if err := os.Rename(tempTarPath, absOutput); err != nil {
```
shizhMSFT: What if the final output path is not writable in the first place? We shouldn't do backup at all!

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,422 @@
...
+		// resolve the specified tags
+		tags = specifiedTags
+		descs = make([]ocispec.Descriptor, 0, len(tags))
+		for _, tag := range tags {
+			desc, err := oras.Resolve(ctx, target, tag, oras.DefaultResolveOptions)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to resolve tag %q: %w", tag, err)
+			}
+			descs = append(descs, desc)
+		}
```
shizhMSFT: It seems this part can be refactored out so that we won't have duplicated code below.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,422 @@
...
+	}); err != nil {
+		return nil, nil, fmt.Errorf("failed to find tags: %w", err)
+	}
+
+	return tags, descs, nil
+}
+
+// parseArtifactReferences parses the input string into a repository
+// and a slice of tags.
+func parseArtifactReferences(artifactRefs string) (repository string, tags []string, err error) {
```
shizhMSFT: Can we reuse `registry.ParseReferenece` and work on top of it instead of manual parsing?

Think about `strings.Split(artifactRefs, ",")` first. Then do reference parsing and tag validating.

internal/io/tar_utils.go
Wwwsylvia: `AddFS` won't return error if sourceDir is not a directory, it will create a tar with no file entries in it.

cmd/oras/root/backup.go
Wwwsylvia: What do you mean by "work with index"?

cmd/oras/root/backup.go
Wwwsylvia: Good catch!

cmd/oras/root/backup.go
Wwwsylvia: Sounds good!

cmd/oras/root/backup.go
Wwwsylvia: It's for atomicity of the output file and to avoid middle states. If we create the tar file at the first place, there is a chance that users can access the file before the tar process completes.

cmd/oras/root/backup.go
Wwwsylvia: We can first create the directories and check if the file already exist?

cmd/oras/root/backup.go
Wwwsylvia: What's the benefit of this approach? `registry.ParseReferenece` can just parse the first tag and we still need to validate the rest. I find the current approach is clearer to read. 🤔

cmd/oras/root/backup.go
Wwwsylvia: My intention was to resolve the tags page-by-page, so I didn't do things like:

```go
if len(specifiedTags) > 0 {
      tags = specifiedTags
} else {
      tags, _ = registry.Tag(ctx, tagLister)
}
for _, tag := range tags {
    desc, err := oras.Resolve(ctx, target, gotTag, oras.DefaultResolveOptions)
    // ...
}
```

cmd/oras/root/backup.go
Wwwsylvia: The name `postBackup` feels like there is a `preBackup`😂 Maybe something like `handleBackupOutput`, as it actually dealing with the output of the backup?

cmd/oras/root/backup.go
Wwwsylvia: Updated.

cmd/oras/root/backup.go
Wwwsylvia: Refactored.

cmd/oras/root/backup.go
Wwwsylvia: Or `finalizeBackupOutput`?

cmd/oras/root/backup.go
Wwwsylvia: OK. I refactored it to recursively count all referrers.

internal/graph/graph.go
```diff
@@ -131,6 +137,19 @@ func FindPredecessors(ctx context.Context, src oras.ReadOnlyGraphTarget, descs [
 	return referrers, nil
 }
 
+func RecursiveFindPredecessors(ctx context.Context, src oras.ReadOnlyGraphTarget, descs []ocispec.Descriptor, opts oras.ExtendedCopyGraphOptions) ([]ocispec.Descriptor, error) {
+	var allPredecessors []ocispec.Descriptor
+	for len(descs) > 0 {
+		predecessors, err := FindPredecessors(ctx, src, descs, opts)
```
shizhMSFT: Find predecessors or referrers? They are different in some edge cases.

cmd/oras/internal/display/metadata/text/backup.go
```diff
@@ -0,0 +1,73 @@
...
+// OnArtifactPulled implements metadata.BackupHandler.
+func (bh *BackupHandler) OnArtifactPulled(tag string, referrerCount int) error {
+	// represent duration in a human-readable format
+	return bh.printer.Printf("Pulled tag %s and %d referrer(s)\n", tag, referrerCount)
+}
+
+// OnTagsFound implements metadata.BackupHandler.
+func (bh *BackupHandler) OnTagsFound(tags []string) error {
+	if len(tags) == 0 {
+		return nil
```
shizhMSFT: What's the output if no tags found?

cmd/oras/root/backup.go
shizhMSFT: I think you can have something like
```go
resolve := func(tags []string) error {
	for _, tag := range tags {
		desc, err := oras.Resolve(ctx, target, tag, oras.DefaultResolveOptions)
		if err != nil {
			return fmt.Errorf("failed to resolve tag %q: %w", tag, err)
		}
		descs = append(descs, desc)
	}
	return nil
}
```

Then you can have
```go
if len(specifiedTags) > 0 {
	if err := resolve(specifiedTags) {
		return nil, nil, err
	}
	return specifiedTags, desc, nil
}

// discover all tags in the repository and resolve them
tagLister, ok := target.(registry.TagLister)
if !ok {
	return nil, nil, errTagListNotSupported
}
if err := tagLister.Tags(ctx, "", func(gotTags []string) error {
	if err := resolve(specifiedTags) {
		return nil, nil, err
	}
	tags = append(tags, gotTags...)
	return nil
}); err != nil {
	return nil, nil, fmt.Errorf("failed to find tags: %w", err)
}
return tags, descs, nil
```

cmd/oras/root/backup.go
shizhMSFT: There is no middle state.

> there is a chance that users can access the file before the tar process completes.

That is not a common assumption in OS.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,458 @@
...
+		var err error
+		absOutput, err = filepath.Abs(opts.output)
+		if err != nil {
+			return fmt.Errorf("failed to get absolute path for output file %s: %w", opts.output, err)
+		}
+	}
+	if err := os.MkdirAll(filepath.Dir(absOutput), 0755); err != nil {
+		// ensure target directory exists
+		return fmt.Errorf("failed to create directory for output file %s: %w", absOutput, err)
+	}
```
shizhMSFT: Creating the parent folders are not necessary as other CLI tools or shells.

cmd/oras/root/backup.go
shizhMSFT: What if we don't need to create directories?

cmd/oras/root/backup.go
shizhMSFT: For example:
```console
$ ls
$ echo hello > hello.txt
$ cp hello.txt foo/bar.txt
cp: cannot create regular file 'foo/bar.txt': No such file or directory
```

cmd/oras/root/backup.go
shizhMSFT: Basically, I don't want to have parsing logics (L412-L430) here as they may be error-prone and duplicated. We should build code (e.g. reuse code) on top of `oras-go` just like we did [here](https://github.com/oras-project/oras/blob/2a3c65a67fca30cd7dfa63dbcff8508df2400c36/cmd/oras/root/push.go#L126-L127).

internal/graph/graph.go
Wwwsylvia: It's to find referrers. I renamed the method for clarity.

cmd/oras/internal/display/metadata/text/backup.go
Wwwsylvia: In reality, it's handled in the backup command. But it might be better to print something here, although this part will not be reached. Updated.

cmd/oras/root/backup.go
Wwwsylvia: OK so we should just leave it to the user

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,423 @@
...
+	return nil
+}
+
+// backupTagWithReferrers copies the artifact identified by tag and its referrers from src to dst.
+func backupTagWithReferrers(ctx context.Context, src oras.ReadOnlyGraphTarget, dst oras.GraphTarget, tag string, root ocispec.Descriptor, extCopyGraphOpts oras.ExtendedCopyGraphOptions) (int, error) {
+	if err := recursiveCopy(ctx, src, dst, tag, root, extCopyGraphOpts); err != nil {
+		return 0, fmt.Errorf("failed to pull tag %q and referrers, digest %q: %w", tag, root.Digest.String(), err)
+	}
+	return countReferrers(ctx, dst, tag, root, extCopyGraphOpts)
```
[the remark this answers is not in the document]
Wwwsylvia: it's by design for code cleanness

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,423 @@
...
+	// Prepare copy source and destination
+	srcRepo, err := opts.NewRepository(opts.repository, opts.Common, logger)
+	if err != nil {
+		return fmt.Errorf("failed to prepare repository %s for backup: %w", opts.repository, err)
+	}
+	dstOCI, err := oci.New(dstRoot)
+	if err != nil {
+		return fmt.Errorf("failed to prepare OCI store for backup: %w", err)
+	}
+	statusHandler, metadataHandler := display.NewBackupHandler(opts.Printer, opts.TTY, opts.repository, dstOCI)
```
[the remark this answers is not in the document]
Wwwsylvia: by design

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,423 @@
...
+	if opts.outputFormat != outputFormatTar {
+		// If output format is not a tar, we are done
+		return nil
+	}
+
+	// exporting the backup to a tar archive
+	if err := metadataHandler.OnTarExporting(opts.output); err != nil {
+		return err
+	}
+	tarFile, err := os.Create(opts.output)
```
shizhMSFT, line 323: Can we test whether `ots.output` is creatable before the actual backup?

shizhMSFT: so that we can fail fast as it is a common failure path.

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,423 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to resolve tag %q: %w", tag, err)
+			}
+			descs = append(descs, desc)
+		}
+		return nil
+	}
+	if len(specifiedTags) > 0 {
+		// resolve the specified tags
+		descs = make([]ocispec.Descriptor, 0, len(specifiedTags))
```
shizhMSFT, line 363: nit: do we really need this although it does no harm?

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,423 @@
...
+	parsedRepo.Reference = "" // clear the tag
+	repository = parsedRepo.String()
+
+	tags = make([]string, 0, len(extraTags)+1)
+	if mainTag != "" {
+		tags = append(tags, mainTag)
+	}
+	// Validate each tag
+	for _, tag := range extraTags {
+		tag = strings.TrimSpace(tag)
```
shizhMSFT: nit: is it necessary? in which scenario, can a user type spaces in tag?

Wwwsylvia: I‘m OK with removing it. It just depends on whether we want to be strict with inputs. For example, users might accidentally type:

`localhost:5000/hello:v1,v2,v3,v4, v5`

cmd/oras/root/backup.go
Wwwsylvia: I think it's more efficient to handle slice growth🤔 why not?

cmd/oras/root/backup.go
Wwwsylvia: Let's be strict then

cmd/oras/root/backup.go
Wwwsylvia: Updated to test at the beginning

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,439 @@
...
+	mainTag := parsedRepo.Reference
+	parsedRepo.Reference = "" // clear the tag
+	repository = parsedRepo.String()
+
+	// validate each tag
+	for _, tag := range extraTags {
+		if tag == "" {
+			return "", nil, fmt.Errorf("empty tag in reference %q", artifactRefs)
+		}
+		if !tagRegexp.MatchString(tag) {
```
shizhMSFT: Let's eliminate the use of `tagRegexp`.

```suggestion
		parsedRepo.Reference = tag
		if err := parsedRepo.ValidateReferenceAsTag(); err != nil {
```

cmd/oras/root/backup.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+	if mainTag == "" {
+		if len(extraTags) == 0 {
+			// no tags
+			return repository, nil, nil
+		}
+		// "localhost:5000/repo:,v1" is invalid
+		return "", nil, fmt.Errorf("empty tag in reference %q", artifactRefs)
+	}
```
shizhMSFT: We can do it early.

cmd/oras/root/backup.go
```diff
@@ -163,6 +163,16 @@ func runBackup(cmd *cobra.Command, opts *backupOptions) error {
 	case outputFormatDir:
 		dstRoot = opts.output
 	case outputFormatTar:
+		// test if the output file can be created and fail early if there is an issue
+		fp, err := os.Create(opts.output)
```
shizhMSFT: Is it safer to `os.OpenFile` without `O_TRUNC`?

Wwwsylvia: updated

shizhMSFT (approved):
LGTM

Issue #1791, opened 2025-07-30
Support content integrity detection


FeynmanZhou:
### What is the version of your ORAS CLI

oras v1.3.0-beta.3

### What would you like to be added?

Introduce a new `oras check` command that verifies the integrity of an artifact in a repository. This command would help users detect missing, corrupted, or tampered content, improving trust and reliability when distributing an artifact across filesystem and registries.

```
$ oras check docker.io/library/busybox:latest

▶ Checking image: docker.io/library/busybox:latest

Manifests:
✓ sha256:ec153840d1e635ac434fab5e377081f17e0e15afab27beb3f726c3265039cfff
✓ sha256:03b62250a3cb1abd125271d393fc08bf0cc713391eda6b57c02d1ef85efcc25c
✗ sha256:28d150a085778a307404359779e0b9b3c670597b4286cba4c66f052edbb1ecaa — error: not found
✓ sha256:817fa52143d152895d63824cb3be81294fb204b042d1b3706a393b7a0b6e5fb6

Blobs:
✓ sha256:74cc54e27dc41bb10dc4b2226072d469509f2f22f1a3ce74f4a59661a1d44602
✗ sha256:e6590344b1a5dc518829d6ea1524fc12f8bcd14ee9a02aa6ad8360cce3a9a9e9 — error: size mismatch (expected: 32, got: 42)
✗ sha256:04d15ddc0f179fa826db9218594a0f646c56f6e2432cb2aac0f5627fc596b904 — error: digest mismatch

❌ Integrity check failed (2 blobs, 1 manifest corrupted)
⏱️ Elapsed: 42s
```

Similar feature request is tracked in oras-go SDK https://github.com/oras-project/oras-go/issues/986

### Why is this needed for ORAS?

While ORAS provides robust capabilities to push, pull, copy, backup, and restore artifacts, there is currently no built-in mechanism to validate the integrity of artifacts by detecting their manifests and blobs. This is particularly important for:

- Air-gapped or disaster recovery workflows where artifact completeness must be guaranteed
- Debugging and remediation of corrupted image layers in registries
- Ensuring content has not been tampered with post-push
- CI/CD pipelines that need to validate content before deployment

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

PR #1803, opened 2025-08-08
[wip] `oras check` design

wangxiaoxuan273:

Fixes #

### DIFF
--- a/docs/proposals/check-design.md
+++ b/docs/proposals/check-design.md
@@ -1,0 +1,71 @@
+# oras check draft design
+
+This is a draft design for the `oras check` command.
+
+## UX
+
+Command:
+
+```
+oras check [flags] <name>{:<tag>|@<digest>}
+
+Example - validate a single image referenced by tag:
+  oras check localhost:5000/hello:v1
+
+Example - validate multiple images referenced by tags:
+  oras check localhost:5000/hello:v1,v2,v3
+
+Example - validate a single image referenced by digest:
+  oras check localhost:5000/hello@sha256:fd6ed2f36b5465244d5dc86cb4e7df0ab8a9d24adc57825099f522fe009a22bb
+
+```
+
+Flags:
+
+`--oci-layout` / `--oci-layout-path` ：OCI Layout validation
+
+`--include-referrers`: include validation of the referrers (Note: this is predecessor validation, only enabled by flag)
+
+`--concurrency`: set concurrency level
+
+## Validation behaviors (For manifest only, not index yet)
+
+**Step 1: Resolve the reference given by the user**
+
+Returns an error if the reference fails to resolve, or resolves to a blob instead of a manifest.
+
+**Step 2: Validate the manifest against the resolved descriptor**
+
+Fetch the manifest content by descriptor, and check if the **MediaType**, **Size** and **Digest** are consistent with the descriptor.
+
+**Step 3: Validate the blobs (config and layers) against the descriptors in the manifest**
+
+Fetch the blobs (config and layers) by descriptors in the manifest, and check if the **Size** and **Digest** are consistent with the descriptors.
+
+
+```mermaid
+graph TD;
+    A[referrer of v1] -- subject --> B[manifest v1]
+```
+
+**Step 4: If the subject field is not empty, validate the subject manifest**
+
+(Note: this is successor validation, enabled by default)
+
+Do step 2 with the descriptor in the subject field. Then do step 3.
+
+**Step 5: If --include-referrers flag is used, validate the referrers**
+
+(Note: this is predecessor validation, only enabled by flag)
+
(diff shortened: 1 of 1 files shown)


docs/proposals/check-design.md
```diff
@@ -0,0 +1,114 @@
...
+  oras check localhost:5000/hello@sha256:fd6ed2f36b5465244d5dc86cb4e7df0ab8a9d24adc57825099f522fe009a22bb
+```
+
+### Flags:
+
+`--oci-layout` / `--oci-layout-path` ：OCI Layout validation
+
+`--include-referrers`: include validation of the referrers (Note: this is predecessor validation, only enabled by flag)
+
+`--concurrency`: set concurrency level
```
FeynmanZhou, line 30: Why `--concurrency` is necessary of `oras check`?

docs/proposals/check-design.md
```diff
@@ -0,0 +1,114 @@
+# oras check draft design
+
+This is a draft design for the `oras check` command.
```
FeynmanZhou, line 3: Could you clarify what content will be checked? Linking to this issue https://github.com/oras-project/oras/issues/1791 can also provide some context to reviewers

docs/proposals/check-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+Verify that the subject descriptor of the referrers is consistent with the manifest.
+
+## PoC implementation
+https://github.com/oras-project/oras/pull/1801
+
+## Questions for discussion
+1. Should we support reference by digest?
+2. Should we support validating index? What should the behavior be?
+3. Should we support validating an entire repository? (`oras check localhost:5000/hello`, no image reference is given)
```
FeynmanZhou: We may want to think which use case that users want to check the entire repo?  My assumption is that users may want to check the content completeness and integrity when backup or restore an entire repo

docs/proposals/check-design.md
```diff
@@ -0,0 +1,114 @@
...
+Resolve the referrer descriptors, then do step 2 and 3.
+
+Verify that the subject descriptor of the referrers is consistent with the manifest.
+
+## PoC implementation
+https://github.com/oras-project/oras/pull/1801
+
+## Questions for discussion
+1. Should we support reference by digest?
+2. Should we support validating index? What should the behavior be?
```
FeynmanZhou: I think an image index should be considered as it will saves some time for users to check a multi-platform image; otherwise, users have to manually retrieve digest of each multi-platform image and check them one by one. 

The behavior would be checking the content integrity of each platform-specific manifest.

docs/proposals/check-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+Resolve the referrer descriptors, then do step 2 and 3.
+
+Verify that the subject descriptor of the referrers is consistent with the manifest.
+
+## PoC implementation
+https://github.com/oras-project/oras/pull/1801
+
+## Questions for discussion
+1. Should we support reference by digest?
```
FeynmanZhou: Can you clarify this question?

docs/proposals/check-design.md
wangxiaoxuan273: I think it is possible to check multiple layers concurrently? This can be a nice-to-have optimization.

docs/proposals/check-design.md
wangxiaoxuan273: What I mean is, currently we support
`oras check localhost:5000/repo:v1`

Do we want to support
`oras check localhost:5000/repo@sha256:xxxxx` ?

docs/proposals/check-design.md
wangxiaoxuan273: For each image specified by the user (by name or digest), we will check:

[for the manifest]
1. If the manifest referenced by the tag or digest exists in the repository.
2. If the manifest's size, digest and media type match the information saved on the repo server. (technically, that means if the descriptor resolved by the repo server matches the actual manifest)

[for each blob, including config and layers]
1. If the blob referenced by the manifest exists in the repository.
2. If the blob's size and digest match the descriptors in the manifest

[check the subject if present, recursively]
1. check the subject's manifest as mentioned above
2. check the subject's blobs as mentioned above

docs/proposals/check-design.md
wangxiaoxuan273: Two scenarios I can think of:

1. When a user fails to pull an image from the registry, he can run `oras check` to see if the image is corrupted. This helps the user to troubleshoot faster. 
2. User may want to set up a periodic pipeline (say once a month) to check that all current images in the production repository are intact.

docs/proposals/check-design.md
TerryHowe: Maybe just add this text  here?

PR #1832, opened 2025-08-27
chore(ux): update colors for oras discover tree view

### LINKED ISSUE #1772: Better color scheme for `oras discover` tree view?
Some users have mentioned that the current color scheme for the `oras discover --format tree` command can be a bit distracting, as the colors for digests and artifact types are both very noticeable.
Let's discuss if there are better color schemes we could use to improve readability of the output.

**Current display:**

Image

**Possible improvements:**

1. Change the digest color from light green to dark green to make it less highlighted. Change the annotations color to light black. For example:

Image

2. Use the same color for digest and artifact type at the same level, and use a different color for different levels. For example:

Image

Wwwsylvia:

This PR updates the colors of digests and the `[annotation]` nodes in the `oras discover` tree view to improve readability.

**BEFORE**:
image

image

**AFTER**:
image

image

Fixes #1772 

### DIFF
--- a/cmd/oras/internal/display/metadata/tree/discover.go
+++ b/cmd/oras/internal/display/metadata/tree/discover.go
@@ -31,8 +31,8 @@ import (
 
 var (
 	artifactTypeColor = aec.LightYellowF
-	digestColor       = aec.LightGreenF
-	annotationsColor  = aec.LightCyanF
+	digestColor       = aec.GreenF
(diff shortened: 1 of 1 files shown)


wangxiaoxuan273 (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1833, opened 2025-08-27
chore(ux): improve oras backup error message

### LINKED ISSUE #1831: Improve error message in `oras backup` when output .tar path is a directory
When using the `oras backup` command with an output path ending in .tar, the tool attempts to create a tar archive at that location. However, if a directory with the same name already exists, the command fails with an error message:

```console
$ oras backup localhost:5001/oras-cli --include-referrers --output oras-tool.tar

Error: unable to create output file oras-tool.tar: open oras-tool.tar: is a directory
```

Some users might not understand this error. We should consider enhancing the message to clearly indicate that the target path already exists as a directory and cannot be used as a tar archive output.

Wwwsylvia:

This PR improves the error message in `oras backup` for the case where the output path ends with `.tar` but is an existing directory.

**BEFORE**:

> Error: unable to create output file test.tar: open test.tar: is a directory

image

**AFTER**:

> Error: the output path test.tar already exists and is a directory: open test.tar: is a directory
> To back up to a tar archive, please specify a different output file name or remove the existing directory.

image

Fixes #1831

### DIFF
--- a/cmd/oras/root/backup.go
+++ b/cmd/oras/root/backup.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -167,6 +168,13 @@ func runBackup(cmd *cobra.Command, opts *backupOptions) error {
 		// test if the output file can be created and fail early if there is an issue
 		fp, err := os.OpenFile(opts.output, os.O_CREATE|os.O_WRONLY, 0666)
 		if err != nil {
+			var pe *fs.PathError
+			if errors.As(err, &pe) && errors.Is(pe.Err, syscall.EISDIR) {
+				return &oerrors.Error{
+					Err:            fmt.Errorf("the output path %s already exists and is a directory: %w", opts.output, err),
+					Recommendation: "To back up to a tar archive, please specify a different output file name or remove the existing directory.",
+				}
+			}
 			return fmt.Errorf("unable to create output file %s: %w", opts.output, err)
 		}
 		if err := fp.Close(); err != nil {
--- a/cmd/oras/root/backup.go
+++ b/cmd/oras/root/backup.go
@@ -167,7 +167,7 @@ func runBackup(cmd *cobra.Command, opts *backupOptions) error {
 		// test if the output file can be created and fail early if there is an issue
 		fp, err := os.OpenFile(opts.output, os.O_CREATE|os.O_WRONLY, 0666)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/oras/root/backup.go
```diff
@@ -167,6 +168,13 @@ func runBackup(cmd *cobra.Command, opts *backupOptions) error {
 		// test if the output file can be created and fail early if there is an issue
 		fp, err := os.OpenFile(opts.output, os.O_CREATE|os.O_WRONLY, 0666)
 		if err != nil {
+			var pe *fs.PathError
+			if errors.As(err, &pe) && errors.Is(pe.Err, syscall.EISDIR) {
```
shizhMSFT: Can we avoid using `syscall` since it is not portable?

shizhMSFT: Let's consider [os.Stat()](https://pkg.go.dev/os#Stat) instead.

shizhMSFT (changes requested):

cmd/oras/root/backup.go
Wwwsylvia: updated

cmd/oras/root/backup.go
```diff
@@ -164,9 +164,14 @@ func runBackup(cmd *cobra.Command, opts *backupOptions) error {
 	case outputFormatDir:
 		dstRoot = opts.output
 	case outputFormatTar:
-		// test if the output file can be created and fail early if there is an issue
 		fp, err := os.OpenFile(opts.output, os.O_CREATE|os.O_WRONLY, 0666)
 		if err != nil {
+			if fi, err := os.Stat(opts.output); err == nil && fi.IsDir() {
+				return &oerrors.Error{
+					Err:            fmt.Errorf("the output path %q already exists and is a directory", opts.output),
```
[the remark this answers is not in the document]
Wwwsylvia: This is by design. Users might find the underlying error distracting.

cmd/oras/root/backup.go
```diff
@@ -167,6 +167,12 @@ func runBackup(cmd *cobra.Command, opts *backupOptions) error {
 		// test if the output file can be created and fail early if there is an issue
 		fp, err := os.OpenFile(opts.output, os.O_CREATE|os.O_WRONLY, 0666)
 		if err != nil {
+			if fi, err := os.Stat(opts.output); err == nil && fi.IsDir() {
```
shizhMSFT: for readability
```suggestion
			if fi, statErr := os.Stat(opts.output); statErr== nil && fi.IsDir() {
```

shizhMSFT (approved):
LGTM

PR #1850, opened 2025-09-11
chore: convert console output to creack/pty

### LINKED ISSUE #1449: containerd/console is broken for mac consider creack/pty

1.2.0

Working console

Tests are broken for mac. They will not fix.

https://github.com/containerd/console/issues/79

TerryHowe:
The https://github.com/containerd/console we are using is pretty much a dead project. There has been a ticket about mac failures for over a year and no action over there other than bug fixes. The project is small time:
Screenshot 2025-09-10 at 6 10 00 PM

The creack/pty project is a lot more active:
Screenshot 2025-09-10 at 6 10 11 PM

Closes: https://github.com/oras-project/oras/issues/1449

### DIFF
--- a/cmd/oras/internal/display/status/console/console.go
+++ b/cmd/oras/internal/display/status/console/console.go
@@ -36,19 +36,11 @@ const (
 
 // Console is a wrapper around PTY and ANSI escape codes.
 type Console interface {
-	Write(p []byte) (n int, err error)
-	Size() (*WinSize, error)
 	GetHeightWidth() (height, width int)
 	Save()
 	NewRow()
 	OutputTo(upCnt uint, str string)
 	Restore()
-}
-
-// WinSize represents the size of a terminal window.
-type WinSize struct {
-	Height uint16
-	Width  uint16
 }
 
 type console struct {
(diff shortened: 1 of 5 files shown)


Wwwsylvia:
Should the PR title start with "refactor"?

cmd/oras/internal/display/status/console/console.go
```diff
@@ -33,27 +34,47 @@ const (
 	Restore = "\0338"
 )
 
-// Console is a wrapper around containerd's Console and ANSI escape codes.
+// Console is a wrapper around PTY and ANSI escape codes.
 type Console interface {
-	containerd.Console
+	Write(p []byte) (n int, err error)
+	Size() (*WinSize, error)
```
shizhMSFT: nit: better perf
```suggestion
	Size() (WinSize, error)
```

internal/testutils/console.go
```diff
@@ -25,34 +25,30 @@ import (
...
 	if err != nil {
 		return nil, nil, err
 	}
-	device, err := os.OpenFile(devicePath, os.O_RDWR, 0)
-	if err != nil {
-		return nil, nil, err
-	}
-	return pty, device, nil
+	return ptmx, pts, nil
 }
```
shizhMSFT, line 35: nit: should that be simpler?
```suggestion
func NewPty() (*os.File, *os.File, error) {
	return pty.Open()
}
```

shizhMSFT (approved):
LGTM with nits

cmd/oras/internal/display/status/console/console.go
qweeah: Looks like `Size` function doesn't need to be added to `Console` interface?

cmd/oras/internal/display/status/console/console.go
```diff
@@ -33,27 +34,47 @@ const (
 	Restore = "\0338"
 )
 
-// Console is a wrapper around containerd's Console and ANSI escape codes.
+// Console is a wrapper around PTY and ANSI escape codes.
 type Console interface {
-	containerd.Console
+	Write(p []byte) (n int, err error)
```
qweeah: No need to add Write to `Console` interface.

qweeah (approved):
LGTM with minor suggestions.

qweeah:
It would be a good start on extending test coverage in Darwin. Thanks for your contribution. @TerryHowe

qweeah (changes requested):
Taking approvement back. I tested on Windows and the tty fails. Confirming offline.

qweeah (approved):
LGTM. Sry I used a bad build to test on Windows. The actual build in this PR works fine.

qweeah:
The width on Window not the actual size though, the width is always 80. @Wwwsylvia @shizhMSFT What do you think?

image

TerryHowe:
> Should the PR title start with "refactor"?

Good point, fixed. This happenwd to be a case where the code was nicely encapsulated, so went easier than expected!

shizhMSFT (changes requested):
> The width on Window not the actual size though, the width is always 80

I've reproduced that. Good catch. It's a regression. We need to find out a way to fix it.

shizhMSFT:
@TerryHowe After investigating `github.com/creack/pty`, I found that it does not support Windows. 😢 
When calling `pty.Getsize(c.file)`, `creack/pty` returns `ErrUnsupported` on Windows. Then `oras` fallback to a `80x10` window.

TerryHowe:
Just fixing conflicts

TerryHowe:
Giving up on creack/pty

PR #1852, opened 2025-09-11
chore: update linting scope

TerryHowe:

Add some additional lint checks.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,70 @@
+# Copyright The ORAS Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+formatters:
+  enable:
+    - gofmt
+    - goimports
+
+  exclusions:
+    generated: lax
+
+  settings:
+    gofmt:
+      simplify: true
+
+linters:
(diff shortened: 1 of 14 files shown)


TerryHowe:
codecov/patch is not useful here

cmd/oras/internal/option/remote.go
```diff
@@ -288,7 +288,7 @@ func (remo *Remote) authClient(registry string, debug bool) (client *auth.Client
 		}
 		client.Credential = credentials.Credential(remo.store)
 	}
-	return
+	return client, nil
```
shizhMSFT: Should we just change the named return values to unnamed at line 252?

internal/graph/graph.go
```diff
@@ -78,30 +78,30 @@ func Successors(ctx context.Context, fetcher content.Fetcher, node ocispec.Descr
...
-			return
+			return nil, nil, nil, err
 		}
 		nodes = index.Manifests
 		subject = index.Subject
 	default:
 		nodes, err = content.Successors(ctx, fetcher, node)
 	}
-	return
+	return nodes, subject, config, err
```
shizhMSFT: Let's not use named return values by changing line 62.

TerryHowe: It isn't the named return values that lint is complaining about so much as the naked returns. I went through and changed this to remove the named returned values, but I don't think that is really what you are looking for, so I didn't push that change.

Named return values are kind of handy for documentation so people understand what is being returned if it isn't clear.

cmd/oras/internal/option/remote.go
TerryHowe: I removed the rule for naked return values.

internal/graph/graph.go
TerryHowe: I removed the rule for naked return values.

FeynmanZhou:
Hi @TerryHowe, are we able to merge this PR? You may want to rebase it before merging. Thanks

TerryHowe:
New problem was introduced, so I've had to update.

PR #1855, opened 2025-09-12
fix: codeql was missing this configuration file

TerryHowe:

Looks like CodeQL is not running without this file

For example:
https://github.com/oras-project/oras/pull/1854/checks?check_run_id=50246044453


TerryHowe:
I think CodeQL is still concerned about go 1.24. Ignore this I guess.

PR #1912, opened 2025-11-24
fix: CodeCov patch configuration

TerryHowe:
I had a PR where CodeCov Patch was failing because I had e2e tests. My changes had 89% coverage, but the e2e tests brought the coverage down.


sajayantony (approved):
LGTM

PR #1914, opened 2025-11-26
chore: move Steve Lasker to emeritus

TerryHowe:
Steve Lasker has not been active for over six months. While Steve's contributions are greatly appreciated, priorities change.


sabre1041 (approved):
LGTM

Issue #1893, opened 2025-11-05
Store full image reference during oras backup


AustinAbro321:
### What is the version of your ORAS CLI

1.3.0

### What would you like to be added?

I would like the full image reference to be stored in the image during oras backup.

I can see a few ways of accomplishing this:
- We could add the entire image reference to `org.opencontainers.image.title`. 
- If we have an appetite for breaking changes, we could simply add the entire image reference to `org.opencontainers.image.ref.name`
- We could add the tag to `org.opencontainers.image.ref.name` and the base name to `org.opencontainers.image.base.name`

### Why is this needed for ORAS?

As a maintainer of an air-gap packaging tool called [Zarf](https://github.com/zarf-dev/zarf), I would like Zarf to be able to automatically determine the name of images stored in image tars that my user's add to their package. The `docker save` command for instance always stores the full image reference under `io.containerd.image.name` so even if users have images from 10 different repositories, Zarf will know the reference for each when it pushes these images to the air-gap registry.

I would be happy to implement this, however I do not know which annotation strategy the oras project would like to use. 

### Are you willing to submit PRs to contribute to this feature?

- [x] Yes, I am willing to implement it.

wangxiaoxuan273:
@shizhMSFT @Wwwsylvia @sajayantony Maintainers, please take a look. cc: @FeynmanZhou 

This issue should be marked with "feature request".

shizhMSFT:
@TerryHowe @FeynmanZhou It is a sub scenario of backing up / restoring multiple repositories across registries.

TerryHowe:
Interesting alternatives proposed.

To review the alternatives I attempted:

1. Tag each image with original registry/repository:tag or repository:tag
2. Store artifacts backed up in directories based on repository name

One other alternative I was going to look at is add an SBOM to the backup that would correlate the images to the repository or registry/repository.

Pretty much a duplicate of https://github.com/oras-project/oras/issues/730 but this is a smaller subset of the problem in #730

FeynmanZhou:
Sounds like a valid scenario! @AustinAbro321 thanks for proposing this. We will assign this issue to you.

I personally vote for adding the entire image reference to `org.opencontainers.image.title`. This looks like more consistent with the OCI image spec.

AustinAbro321:
I went with a simple solution in #1904 of only adding the annotation during backup. This is good enough for my purposes, and to close this issue. 

[...]

Issue #1770, opened 2025-07-04
Support `oras repo ls --oci-layout-path`


Wwwsylvia:
### What is the version of your ORAS CLI

v1.3.0-beta.3

### What would you like to be added?

With the support of [oras backup and restore](https://github.com/oras-project/oras/blob/main/docs/proposals/backup-restore.md), users are able to back up artifacts from multiple registries/repositories to an OCI layout. 

We should consider supporting OCI layout for `oras repo ls` to allow users to list all repositories inside the OCI layout for better usability.

Possible UX: 

```console
$ oras repo ls --oci-layout-path oci-dir/
foo.example.com/test
bar.example.com/dev/hello-world
bar.example.com/something/else
```

### Why is this needed for ORAS?

This feature would enhance the usability of the ORAS backup and restore functionality by providing users with a clear overview of all repositories contained within an OCI layout. 

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

shizhMSFT:
Additional possible UX:
```console
$ oras repo ls --oci-layout-path oci-dir/ bar.example.com
dev/hello-world
something/else
```

TerryHowe:
Should we consider making the user specify the regular expression like the ls command? `bar.example.com*`

TerryHowe:
If this was to act like ls, users could in theory specify multiple regular expressions. `oras repo ls --oci-layout-path oci-dir/ *v1.31* *v.1.32*`

sajayantony:
I'm actually leaning towards pipe and filter instead of supporting a built-in filter just to keep the code simpler here.

PR #1917, opened 2025-11-30
docs: add multi-repository backup and restore proposal

shizhMSFT:
## Summary

Adds a comprehensive proposal for extending `oras backup` and `oras restore` commands to support multi-repository operations within a single registry. This enhancement builds upon existing single-repository functionality while maintaining full backward compatibility.

## Changes

- **New file**: `docs/proposals/backup-restore-multi.md` (881 lines)
  - Multi-repository backup syntax and behavior specification
  - Multi-repository restore with flexible mapping capabilities
  - Detailed examples for common enterprise scenarios
  - Implementation plan and user experience considerations

## Key Features Proposed

- Backup artifacts from multiple repositories in the same registry into a single OCI layout
- Restore with repository mapping and restructuring capabilities
- Support for selective tag backup per repository
- Blob deduplication across repositories
- Dry-run mode for validation before restore operations
- Progressive disclosure: simple cases remain simple, complexity only when needed

## Out of Scope

- Backing up entire registries (all repositories)
- Cross-registry backup in a single operation

## Related Issues

- Depends on #1770 - Support `oras repo ls --oci-layout-path` for listing repositories in OCI Image Layout backups
- Addresses #1893 - Use cases for full reference tracking in backups for automation tools

This proposal addresses enterprise scenarios including unified backups for microservices, repository migration between environments, and disaster recovery for specific repository sets.


docs/proposals/backup-restore-multi.md
```diff
@@ -0,0 +1,881 @@
...
+- Enables automation tools (e.g., air-gap packaging tools like [Zarf](https://github.com/zarf-dev/zarf)) to programmatically determine the original location of backed-up artifacts
+- Addresses use cases described in [#1893](https://github.com/oras-project/oras/issues/1893)
+- **Multi-repository backups always use full references** to disambiguate artifacts from different repositories
+
+> [!WARNING]
+> Backups created with `--full-reference` are **not compatible** with ORAS CLI versions prior to v1.4.0 or tools that expect tag-only references in `org.opencontainers.image.ref.name`. Use this flag only when you need full reference information for automation or multi-registry scenarios.
+
+**Example `index.json` formats:**
+
+**Default single-repository backup (tag-only):**
```
TerryHowe, line 436: Seems like it would be better to make full-reference default and something else to make these short references.

We haven't had a release out yet with these short references yet, have we? Not to mention this feature is experimental

FeynmanZhou: I think the short references (tags) are generated in the image annotations since `oras v1.3.0`.

```
"annotations":{"org.opencontainers.image.ref.name":"v1.2.3-2"}
```

docs/proposals/backup-restore-multi.md
```diff
@@ -0,0 +1,881 @@
...
+    // Change path structure for alpine
+    {
+      "source": "docker.io/library/alpine",
+      "target": "registry.mycompany.com/images/alpine"
+    },
+    // Selective tags with mapping
+    {
+      "source": "docker.io/library/nginx",
+      "target": "registry.mycompany.com/webservers/nginx",
+      "tags": ["1.25-alpine"]
```
FeynmanZhou, line 262: Is digest allowed to be specified here?

docs/proposals/backup-restore-multi.md
TerryHowe: I thought that change went in after 1.3.0, but I am probably wrong.

docs/proposals/backup-restore-multi.md
TerryHowe: digest should be support

shizhMSFT: Digest support is ruled out intentionally for this iteration (same as the single repo implementation). The principal is that the `oras backup/restore` is not going to solve 100% scenarios (that's the job of `oras copy`) but 80+% or 90+% common scenarios.

shizhMSFT: We definitely can have an issue for the digest support.

docs/proposals/backup-restore-multi.md
shizhMSFT: I think there is a backward compatible issue, not just for current implementation of `oras backup`/`oras restore` but `oras repo tags` and other command like `oras copy`, if we use full-reference by default for single repo.

shizhMSFT: The backward combability issue not just impact `oras` and it may also impact other OCI-layout compatible tools.

docs/proposals/backup-restore-multi.md
```diff
@@ -0,0 +1,881 @@
...
+
+**Behavior:**
+- `docker.io/library/busybox` → `registry.mycompany.com/library/busybox`
+- `docker.io/library/alpine` → `registry.mycompany.com/library/alpine`
+- `docker.io/library/nginx` → `registry.mycompany.com/library/nginx`
+
+**Output:**
+```console
+Loaded backup archive: base-images.tar (42.8 MB)
+Found 3 repository(ies) with 3 tag(s):
```
sabre1041, line 180: Printing out each tag could be a challenge to read if a repository has a large number of tags

docs/proposals/backup-restore-multi.md
```diff
@@ -0,0 +1,881 @@
...
+```
+
+This approach:
+- Aligns with the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/v1.1.1/annotations.md#pre-defined-annotation-keys), which allows both tag-only and full reference formats in `org.opencontainers.image.ref.name`
+- Enables automation tools (e.g., air-gap packaging tools like [Zarf](https://github.com/zarf-dev/zarf)) to programmatically determine the original location of backed-up artifacts
+- Addresses use cases described in [#1893](https://github.com/oras-project/oras/issues/1893)
+- **Multi-repository backups always use full references** to disambiguate artifacts from different repositories
+
+> [!WARNING]
+> Backups created with `--full-reference` are **not compatible** with ORAS CLI versions prior to v1.4.0 or tools that expect tag-only references in `org.opencontainers.image.ref.name`. Use this flag only when you need full reference information for automation or multi-registry scenarios.
```
AustinAbro321:
```suggestion
> Backups created with `--full-reference` are **not compatible** with ORAS CLI versions prior to v1.4.0 or tools that expect tag-only references in `org.opencontainers.image.ref.name`. Use this flag only when you need full reference information for automation or multi-repository scenarios.
```
I believe the intent here is multi repository since elsewhere in the proposal it says that "all repositories in a backup operation must be from the same registry"

docs/proposals/backup-restore-multi.md
shizhMSFT: Deferred to #1948

PR #1918, opened 2025-12-02
chore: add Terry Howe to owners list

sajayantony:
There seems to a mismatch between owners and maintainers of this sub project. 
We should consider reconciling the 2 at some point.


TerryHowe (approved):
/lgtm

I'm already in CODEOWNERS

PR #1923, opened 2025-12-06
feat: save context between calls

TerryHowe:

This change saves authentication cache between calls so ORAS go scripting will auth once for most usages.

I created a test application to functional test this and it only authenticates the once.

```
With debug on, the second in-process call does not try to authenticate — no anonymous request, no 401 challenge. It attaches credentials proactively from the shared cache and goes straight to 200.

┌──────────────────────────────────┬───────────┬───────────┬─────────────┬───────────────┐
│              Build               │  CALL 1   │  CALL 2   │ Client 401s │ Registry 401s │
├──────────────────────────────────┼───────────┼───────────┼─────────────┼───────────────┤
│ Parent 7383f87 (no shared cache) │ 401 → 200 │ 401 → 200 │ 2           │ 2             │
├──────────────────────────────────┼───────────┼───────────┼─────────────┼───────────────┤
│ PR #1923 46fffb6 (shared cache)  │ 401 → 200 │ 200       │ 1           │ 1             │
└──────────────────────────────────┴───────────┴───────────┴─────────────┴───────────────┘

Client trace for the PR build, verbatim — CALL 2 has a single request and it's already authenticated:
CALL 1: GET .../manifests/v1  -> "401 Unauthorized"
        GET .../manifests/v1  -> "200 OK"
CALL 2: GET .../manifests/v1  -> "200 OK"
Parent build, same program — CALL 2 re-challenges from scratch:
CALL 2: GET .../manifests/v1  -> "401 Unauthorized"
        GET .../manifests/v1  -> "200 OK"
```


cmd/oras/internal/option/remote.go
```diff
@@ -45,6 +45,38 @@ import (
...
+// GetSharedClient returns the shared auth client if one has been set.
+// Returns nil if no shared client has been configured.
+func GetSharedClient() *auth.Client {
+	sharedClientLock.RLock()
+	defer sharedClientLock.RUnlock()
+	return sharedClient
+}
+
+// ClearSharedClient removes the shared auth client.
+func ClearSharedClient() {
```
sajayantony, line 74: Do we need coverage for this?

sajayantony (approved):

cmd/oras/internal/option/remote.go
TerryHowe: Created issue to track this https://github.com/oras-project/oras/issues/2121

PR #1934, opened 2025-12-10
feat: add static security scanning to CI

### LINKED ISSUE #1905: Add static security check
Add a static security check to CI and the makefile.

I think there is a gosec module to golangci

bupd:

Adds vulnerability scanning to the ORAS CI pipeline:

- **govulncheck GitHub Action**: New workflow (`.github/workflows/vulnerability-check.yml`) that runs on PRs and pushes to main/release branches, scanning Go dependencies for known vulnerabilities
- **Local security checks**: New `make govulncheck` target for running vulnerability scans locally before pushing
- **Go version update**: Bumps Go to 1.25.5 to address vulnerabilities in the Go toolchain itself

This complements the existing CodeQL analysis by adding Go-specific vulnerability detection for dependencies.

**Which issue(s) this PR fixes** 
- Fixes #1905

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,18 +27,28 @@ linters:
   default: standard
 
   enable:
+    - bodyclose
+    - copyloopvar
     - depguard
     - dupl
+    - errorlint
     - goheader
     - gomodguard
+    - gosec
     - govet
     - ineffassign
     - misspell
+    - nilerr
+    - nilnil
+    - noctx
     - staticcheck
     - thelper
+    - unconvert
+    - unparam
     - unused
     - usestdlibvars
     - usetesting
+    - whitespace
 
   exclusions:
 
@@ -67,6 +77,12 @@ linters:
 
     goheader:
       template-path: copyright.tmpl
+
+    gosec:
+      excludes:
+        - G306
+        - G115
+        - G402
 
 run:
   timeout: 10m
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ WORKDIR ${ORASPKG}
 RUN make "build-$(echo $TARGETPLATFORM | tr / -)"
 RUN mv ${ORASPKG}/bin/${TARGETPLATFORM}/oras /go/bin/oras
 
-FROM docker.io/library/alpine:3.22.2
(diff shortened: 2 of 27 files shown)


cmd/oras/internal/display/metadata/text/restore_test.go
```diff
@@ -46,10 +46,6 @@ func TestNewRestoreHandler(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			printer := output.NewPrinter(&bytes.Buffer{}, &bytes.Buffer{})
 			handler := NewRestoreHandler(printer, tt.dryRun)
-
```
TerryHowe, line 49: Why is this removed?

cmd/oras/internal/display/metadata/text/restore_test.go
```diff
@@ -46,10 +46,6 @@ func TestNewRestoreHandler(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			printer := output.NewPrinter(&bytes.Buffer{}, &bytes.Buffer{})
 			handler := NewRestoreHandler(printer, tt.dryRun)
```
TerryHowe, line 48:
```suggestion
			handler := NewRestoreHandler(printer, tt.dryRun)
			if handler == nil {
			    t.Fatal("expected a non-nil handler")
			}
```

TerryHowe: I suppose this must be an impossible situation?

bupd: The check `if handler == nil` will never be true. and keeping this creates a dead code.

I have removed this to satisfy the linter. since this is a impossible case. and the function will never return nil.

bupd: Thanks for the review.

cmd/oras/internal/display/metadata/text/restore_test.go
bupd: same impossible condition. handler can never be nil.

bimakw:
@TerryHowe Thank you for the approval! All CI checks are passing. Could this be merged when you have a moment? 🙏

PR #1936, opened 2025-12-10
feat: add golangci-lint configuration with additional linters

bupd:
**What this PR does / why we need it**:
- Add comprehensive golangci-lint configuration with additional linters
- Add goheader linter to enforce license headers in source files
- Fix lint issues across the codebase

**Which issue(s) this PR fixes** 
Fixes #1935

**Please check the following list**:
- [ ]  Does the affected code have corresponding tests, e.g. unit test, E2E test?
- [ ]  Does this change require a documentation update?
- [ ]  Does this introduce breaking changes that would require an announcement or bumping the major version?
- [x]  Do all new files have an appropriate license header?

PR #1942, opened 2025-12-14
chore: expand golangci-lint configuration with additional linters

### LINKED ISSUE #1935: Expand golangci-lint configuration with additional linters

The codebase currently lacks comprehensive linting coverage. Several Go anti-patterns exist that could introduce subtle bugs over time
  - Unchecked errors
  - Improper error handling
  Current `.golangci.yml` has limited linters enabled.

  The project should have a comprehensive linting setup that:
  - Catches common Go anti-patterns early
  - Maintain consistent code quality

### How can we reproduce it?

1. Run `golangci-lint run` with additional linters like errcheck, staticcheck, errorlint, nilerr.
2. Observe various warnings and issues reported

main

ArchLinux BTW

bimakw:
## Summary

This PR expands the golangci-lint configuration by adding additional linters to improve code quality and catch common Go anti-patterns.

### Added Linters:
- **errcheck**: Checks for unchecked errors in the codebase
- **errorlint**: Checks for proper error handling patterns (e.g., using `errors.Is()` instead of `==`)
- **nilerr**: Checks for functions that return nil error when they should return an error

### Why These Linters?

As mentioned in the issue, the current configuration has limited linting coverage. These linters help:
- Catch unchecked errors that could introduce subtle bugs
- Ensure proper error handling patterns are followed
- Maintain consistent code quality

### Note

The `staticcheck` linter mentioned in the issue is already enabled in the configuration.

## Test Plan
- [x] Config syntax is valid (verified by go build)
- [ ] CI will run the linters and report any issues

Fixes #1935

### DIFF
--- a/cmd/oras/internal/option/target_test.go
+++ b/cmd/oras/internal/option/target_test.go
@@ -249,7 +249,7 @@ func TestTarget_ModifyError_errResponse(t *testing.T) {
 	if !modified {
 		t.Errorf("expected error to be modified but received %v", modified)
 	}
-	if got == nil || got.Error() != errResp.Errors.Error() {
+	if got != nil && got.Error() != errResp.Errors.Error() {
 		t.Errorf("unexpected output from Target.ModifyError() = %v", got)
 	}
 	if cmd.ErrPrefix() != oerrors.RegistryErrorPrefix {
@@ -331,10 +331,10 @@ func TestTarget_ModifyError_dockerHint(t *testing.T) {
 		},
 	}
 	tests := []struct {
-		name        string
-		fields      fields
-		err         error
-		modifiedErr *oerrors.Error
+		name          string
+		fields        fields
+		err           error
+		expectedError *oerrors.Error
 	}{
 		{
 			"namespace already exists",
@@ -410,8 +410,8 @@ func TestTarget_ModifyError_dockerHint(t *testing.T) {
 			if !errors.As(got, &gotErr) {
 				t.Errorf("expecting error to be *oerrors.Error but received %T", got)
 			}
-			if gotErr.Err.Error() != tt.modifiedErr.Err.Error() || gotErr.Usage != tt.modifiedErr.Usage || gotErr.Recommendation != tt.modifiedErr.Recommendation {
-				t.Errorf("Target.ModifyError() error = %v, wantErr %v", gotErr, tt.modifiedErr)
+			if gotErr.Err.Error() != tt.expectedError.Err.Error() || gotErr.Usage != tt.expectedError.Usage || gotErr.Recommendation != tt.expectedError.Recommendation {
+				t.Errorf("Target.ModifyError() error = %v, wantErr %v", gotErr, tt.expectedError)
 			}
 			if !modified {
 				t.Errorf("Failed to modify %v", tt.err)


cmd/oras/internal/option/target_test.go
```diff
@@ -249,7 +249,7 @@ func TestTarget_ModifyError_errResponse(t *testing.T) {
 	if !modified {
 		t.Errorf("expected error to be modified but received %v", modified)
 	}
-	if got.Error() != errResp.Errors.Error() {
+	if got == nil || got.Error() != errResp.Errors.Error() {
```
TerryHowe: This doesn't  look right, I'd think something like
```suggestion
	if got != nil && got.Error() != errResp.Errors.Error(){
```

cmd/oras/internal/option/target_test.go
```diff
@@ -406,8 +406,8 @@ func TestTarget_ModifyError_dockerHint(t *testing.T) {
 				IsOCILayout:  tt.fields.IsOCILayout,
 			}
 			got, modified := opts.ModifyError(cmd, tt.err)
-			gotErr, ok := got.(*oerrors.Error)
-			if !ok {
+			var gotErr *oerrors.Error
```
TerryHowe, line 409: Maybe this could be called something like "expectedError" to be more clear. Also, I haven't messed with this, but you are creating a pointer to an object and then using the address of it which doesn't seem right. Same thing below. Can't you just use `errors.Is` with oerrors.Error?

cmd/oras/root/attach.go
```diff
@@ -106,9 +106,10 @@ Example - Attach file to the manifest tagged 'example.com:v1' in an OCI image la
 			}
 			if len(opts.FileRefs) == 0 {
 				// no file argument provided
-				if err, ok := err.(*oerrors.Error); ok && err.OperationType == oerrors.OperationTypeParseArtifactReference {
```
TerryHowe, line 109: Same here

bimakw:
Thanks for the review @TerryHowe!

I've addressed the feedback:
- Fixed the error check logic in line 252 as suggested
- Renamed `modifiedErr` to `expectedError` for clarity

Regarding the `errors.As` pattern with `var oErr *oerrors.Error` followed by `errors.As(err, &oErr)` - this is actually the idiomatic Go pattern for extracting typed errors. From the [Go documentation](https://pkg.go.dev/errors#As):

> `As` finds the first error in err's tree that matches target, and if one is found, sets target to that error value and returns true.

The pattern `var target *MyError; errors.As(err, &target)` is the standard way to use it, as `errors.As` requires a pointer to the target variable to set the value. `errors.Is` is for equality comparison only and cannot extract the error value.

Let me know if you have any other concerns!

TerryHowe (approved):
/lgtm

bimakw:
Hi @TerryHowe,

CI is failing because the new linters found 8 issues in the existing codebase (not from my changes).

How do you want me to handle this? I can:
1. Fix all issues in this PR
2. Fix them in separate PR(s)
3. Add exclusions for now

Let me know!

bupd:
@bimakw how does this PR differ from https://github.com/oras-project/oras/pull/1936

bimakw:
@bupd oh didn't notice yours. yours is more complete anyway, I'll close this one

PR #1954, opened 2026-01-08
feat(copy): Support copying artifacts across multiple platforms.

### LINKED ISSUE #1945: Support multiple `--platform` where appropriate

1.3.0

Since `oras` is already capable of handling multiple platforms where appropriate (such as `oras cp`), it would be ideal if `--platform` could support a list of preferred platforms.

---

Potentially support a wildcard syntax as a shorthand for filtering a group of platforms. For example when parsing an image index, filtering platforms returned by applying the wildcard pattern. For example:
- All archs for a given OS.
- All variants of an arch.
- All OS with a specific arch.

```bash
# Single platform (already supported):
oras cp docker.io/library/caddy:latest --to-oci-layout oras-caddy --platform linux/arm64

# Multiple platforms (unsupported, requires repeating the command with a different `--platform`):
oras cp docker.io/library/caddy:latest --to-oci-layout oras-caddy --platform 'linux/amd64,linux/arm64'
oras cp docker.io/library/caddy:latest --to-oci-layout oras-caddy --platform linux/amd64 --platform linux/arm64

# Multiple platforms with wildcard (unsupported):
oras cp docker.io/library/caddy:latest --to-oci-layout oras-caddy --platform 'linux/*'
oras cp docker.io/library/caddy:latest --to-oci-layout oras-caddy --platform 'linux/arm*'
oras cp docker.io/library/caddy:latest --to-oci-layout oras-caddy --platform '*/arm64'
```

I don't have a specific requirement for the related wildcard filtering, it is just something familiar with Docker Bake `--set` and HCL "splat", along with other tools that support a similar filter/selection pattern.

When using `oras cp` the [default behaviour for an image index is to pull all platforms](https://github.com/oras-project/oras/issues/1456).

When using `--platform` explicitly, whichever the last specified option was is the platform that will be stored. The command must be invoked multiple times with a different platform each time.

For an image index like `docker.io/library/caddy:latest` this is not only multi-arch but multi-platform, with the windows images variants using several GB while the Alpine linux images are considerably smaller (approx 20MB each). I was interested in using `oras cp` with only the linux platform as I had no need for the windows platform to be copied (_nor the spare disk space in this case_).

---

With a single `--platform`, the `oras cp` results in making a copy of just the image manifest (_eg: with an OCI layout `index.json`_) rather than an implicitly creating an image index (_thus would need to be manually constructed afterwards to support tags referring to both archs_).

Likewise for images built via Docker, their `docker build` attestation feature attaches `unknown` platforms to the image index which would also be lost:

<details>
<summary>Example for AMD64 platform with associated attestation manifest</summary>

```json
{
  "manifests": [
    {
      "annotations": {
        "com.docker.official-images.bashbrew.arch": "amd64",
        "org.opencontainers.image.base.digest": "sha256:85f2b723e106c34644cd5851d7e81ee87da98ac54672b29947c052a45d31dc2f",
        "org.opencontainers.image.base.name": "alpine:3.22",
        "org.opencontainers.image.created": "2025-10-08T23:21:09Z",
        "org.opencontainers.image.revision": "5572371a83e48fd0368a4917d0fc48e44ef30582",
        "org.opencontainers.image.source": "https://github.com/caddyserver/caddy-docker.git#5572371a83e48fd0368a4917d0fc48e44ef30582:2.10/alpine",
        "org.opencontainers.image.url": "https://hub.docker.com/_/caddy",
        "org.opencontainers.image.version": "2.10.2-alpine"
      },
      "digest": "sha256:5c18f4b845cd8f2cbb344ffa6fe0391d8c975ed921848c9968298f1eb8940a97",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 1916
    },
    {
      "annotations": {
        "com.docker.official-images.bashbrew.arch": "amd64",
        "vnd.docker.reference.digest": "sha256:5c18f4b845cd8f2cbb344ffa6fe0391d8c975ed921848c9968298f1eb8940a97",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "digest": "sha256:c49b5bb9fafca5664bab24f64ba8ed9dbf93ef119619a617a6452dbf94aaec01",
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      },
      "size": 840
    },
```

</details>

I wouldn't expect that to be supported by this feature request, since filtering by `platform` would lack context (_technically the image manifest is referenced via `vnd.docker.reference.digest`_) and either result in none or all (_IIRC the Docker maintainers were opposed to supporting the OCI tag fallback for when the Referrer API was unavailable, which is the case with GHCR last I checked_).

hellocn9:
In certain scenarios, only the images for the `linux/amd64` and `linux/arm64` platforms need to be synchronized. However, in the current version, if `--platform` is not specified, images for all platforms are synchronized; whereas when `-- platform` is specified, only a single platform architecture can be provided. To address this issue, this PR introduces support for specifying multiple platform architectures with `--platform`, separated by commas.

Fixes #1945


wangxiaoxuan273:
@shizhMSFT @FeynmanZhou  Can you take a look?

FeynmanZhou:
Thanks @hellocn9 for the contribution! Are you looking to resolve this issue in this PR? https://github.com/oras-project/oras/issues/1945

cmd/oras/internal/option/platform.go
```diff
@@ -46,12 +47,62 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
+		// Single platform case - existing behavior
+		return opts.parseSinglePlatform(opts.platform)
+	}
+
+	// Multiple platforms case
+	opts.Platforms = make([]*ocispec.Platform, 0, len(platformStrings))
+	for _, platformStr := range platformStrings {
+		platformStr = strings.TrimSpace(platformStr)
+		if platformStr == "" {
+			continue
```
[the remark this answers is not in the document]
hellocn9: We intentionally skip empty entries.

cmd/oras/internal/option/platform.go
```diff
@@ -46,12 +47,62 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
+			return fmt.Errorf("invalid platform: Architecture cannot be empty")
+		}
+		p.OSVersion = osVersion
+		opts.Platforms = append(opts.Platforms, &p)
+	}
+
+	// Set the first platform as the primary one for backward compatibility
+	if len(opts.Platforms) > 0 {
+		opts.Platform = opts.Platforms[0]
+	}
```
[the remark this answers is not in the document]
hellocn9: No error needed here. An empty platform list (len == 0) is treated as “no architecture specified”, which triggers syncing all architectures — this is by design.

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,173 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+			tagNOpts := oras.DefaultTagNOptions
+			tagNOpts.Concurrency = opts.concurrency
+			tagListener := listener.NewTaggedListener(dst, metadataHandler.OnTagged)
+			if _, err = oras.TagN(ctx, tagListener, opts.To.Reference, opts.extraRefs, tagNOpts); err != nil {
+				return err
+			}
+		}
+
+		return metadataHandler.Render()
+	}
```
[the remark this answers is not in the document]
hellocn9: No error needed here.I believe we should maintain backward compatibility here. If the source is not an index/manifest list (i.e., !isIndex), it’s expected behavior to fall back to using only the first platform from opts.Platform.Platforms[0]. We shouldn’t throw an error in this case — silently using the first platform is intentional and avoids breaking existing workflows.

cmd/oras/root/cp.go
```diff
@@ -161,7 +318,41 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+		return false
+	}
+
+	// OSVersion is optional, if one is empty, ignore comparison
+	if a.OSVersion != "" && b.OSVersion != "" && a.OSVersion != b.OSVersion {
+		return false
+	}
+
+	return true
 }
```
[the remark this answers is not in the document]
hellocn9: We’re aware of the OCI spec, but for now, we intentionally skip OSFeatures comparison to avoid overcomplicating platform matching. This keeps behavior predictable for most users. We can consider adding it later if needed.

cmd/oras/internal/option/platform.go
```diff
@@ -46,12 +47,62 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
+
+	return nil
+}
```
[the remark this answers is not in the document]
hellocn9: Agreed, test coverage is needed. We’ll add dedicated tests for multi-platform parsing in a separate PR to avoid bloating this one. Thanks for flagging!

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,173 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+			tagNOpts.Concurrency = opts.concurrency
+			tagListener := listener.NewTaggedListener(dst, metadataHandler.OnTagged)
+			if _, err = oras.TagN(ctx, tagListener, opts.To.Reference, opts.extraRefs, tagNOpts); err != nil {
+				return err
+			}
+		}
+
+		return metadataHandler.Render()
+	}
+}
```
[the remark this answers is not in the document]
hellocn9: This scenario isn’t something we’re addressing right now; we’ll defer it for future consideration.

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,173 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+		}
+
+		if len(opts.extraRefs) != 0 {
+			tagNOpts := oras.DefaultTagNOptions
+			tagNOpts.Concurrency = opts.concurrency
+			tagListener := listener.NewTaggedListener(dst, metadataHandler.OnTagged)
+			if _, err = oras.TagN(ctx, tagListener, opts.To.Reference, opts.extraRefs, tagNOpts); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
hellocn9: Not planning to refactor this right now — keeping it inline for clarity. Will revisit during future refactors.

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,173 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+		MediaType:   index.MediaType,
+		Digest:      digest.FromBytes(newIndexContent),
+		Size:        int64(len(newIndexContent)),
+		Annotations: index.Annotations,
+	}
+
+	// Prepare copy options
+	extendedCopyGraphOptions := oras.DefaultExtendedCopyGraphOptions
+	extendedCopyGraphOptions.Concurrency = opts.concurrency
+	extendedCopyGraphOptions.FindPredecessors = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
```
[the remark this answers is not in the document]
hellocn9: No action needed for now.

hellocn9:
> Thanks @hellocn9 for the contribution! Are you looking to resolve this issue in this PR? #1945

Yes, this PR addresses #1945. Thanks for reviewing!

cmd/oras/internal/option/platform.go
```diff
@@ -46,12 +47,62 @@ func (opts *Platform) Parse(*cobra.Command) error {
...
 	}
 
+	// Split by comma to support multiple platforms
+	platformStrings := strings.Split(opts.platform, ",")
+	if len(platformStrings) == 1 {
+		// Single platform case - existing behavior
+		return opts.parseSinglePlatform(opts.platform)
+	}
+
+	// Multiple platforms case
```
shizhMSFT: Since you have `parseSinglePlatform`, it is better to refactor this part as a function.

cmd/oras/internal/option/platform.go
```diff
@@ -37,7 +38,7 @@ func (opts *Platform) ApplyFlags(fs *pflag.FlagSet) {
 	if opts.FlagDescription == "" {
 		opts.FlagDescription = "request platform"
 	}
-	fs.StringVarP(&opts.platform, "platform", "", "", opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
+	fs.StringVarP(&opts.platform, "platform", "", "", opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]` or comma-separated list for multiple platforms")
```
shizhMSFT: Consider [StringSliceVarP](https://pkg.go.dev/github.com/spf13/pflag#StringSliceVarP) (also see  [StringSlice](https://pkg.go.dev/github.com/spf13/pflag#StringSlice)) and don't parse the comma manually.

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,181 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+	// We don't set TargetPlatform here since we want to get the full index/list
+	root, err := oras.Resolve(ctx, src, opts.From.Reference, resolveOpts)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to resolve %s: %w", opts.From.Reference, err)
+	}
+
+	// Check if the resolved descriptor is an index/manifest list
+	isIndex := root.MediaType == ocispec.MediaTypeImageIndex || root.MediaType == docker.MediaTypeManifestList
+	if !isIndex {
```
shizhMSFT, line 194: If it is not index but multiple platforms are specified, in the case, we should return error as a manifest can only have one platform.

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,181 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+		return fmt.Errorf("failed to parse index: %w", err)
+	}
+
+	// Filter manifests based on the specified platforms
+	var filteredManifests []ocispec.Descriptor
+	for _, manifest := range index.Manifests {
+		if manifest.Platform != nil && matchesAnyPlatform(manifest.Platform, opts.Platform.Platforms) {
+			filteredManifests = append(filteredManifests, manifest)
+		}
+	}
```
shizhMSFT, line 243: @FeynmanZhou Here we have the ambiguity of the semantics. 

Given
```sh
oras cp --platform linux/amd64,linux/arm64,linux/arm/v7 localhost:5000/net-monitor:v1 localhost:6000/net-monitor-copy:v1
```
What if `localhost:5000/net-monitor:v1` does not have `linux/arm/v7`?

By specifying `--platform linux/amd64,linux/arm64,linux/arm/v7`, it gives a strong semantic meaning that the user want to copy the image of 3 platforms over and those 3 platforms must exist, which is an extended semantic meaning of the current single-platform behavior.

cmd/oras/root/cp.go
```diff
@@ -138,20 +142,181 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+		return fmt.Errorf("no manifests match the requested platforms %v; available platforms in index: %s", requestedPlatforms, availableDesc)
+	}
+
+	// Create a new index with only the filtered manifests
+	newIndex := ocispec.Index{
+		Versioned:   index.Versioned,
+		MediaType:   index.MediaType,
+		Manifests:   filteredManifests,
+		Annotations: index.Annotations,
+	}
```
shizhMSFT: It might be safer just do
```
newIndex := index
newIndex.Manifests = filteredManifests
```

shizhMSFT (changes requested):
Besides above comments, please add related unit tests and E2E tests.

cmd/oras/internal/option/platform.go
hellocn9: @shizhMSFT Thanks for the suggestion! I've refactored this part into a dedicated function to improve code clarity and reuse. The change is now included in the latest commit.

cmd/oras/internal/option/platform.go
hellocn9: Thanks for the suggestion! I’ve updated the code to use StringSliceVarP instead of manually parsing comma-separated strings.

cmd/oras/root/cp.go
hellocn9: @shizhMSFT Thanks for pointing this out! I’ve updated the logic to explicitly report which requested platforms were not matched, along with the list of available platforms in the source image. Now users will see a clear error like:

`Error: only 2 of 3 requested platforms were matched: unmatched platforms: [windows/amd64]; available platforms in index: [linux/amd64, linux/arm64]`

This makes the behavior unambiguous and helps users understand exactly what’s missing. @FeynmanZhou

cmd/oras/root/cp.go
hellocn9: Done! Reused the original index — no need to rebuild it when we can just tweak the manifests. Simpler and safer!

cmd/oras/root/cp.go
hellocn9: I’ve added a clear error message for this case. Now, if the source is not an index/manifest list (e.g., a single manifest) but multiple platforms are specified, users will see:
`Error: source xxxx is not an index or manifest list`

hellocn9:
> Besides above comments, please add related unit tests and E2E tests.

Done! Added unit & E2E tests for all platform-related changes.

hellocn9:
@shizhMSFT I’ve addressed all your feedback — please take a look when you have time!

cmd/oras/internal/option/platform.go
```diff
@@ -82,5 +136,5 @@ type ArtifactPlatform struct {
 // ApplyFlags applies flags to a command flag set.
 func (opts *ArtifactPlatform) ApplyFlags(fs *pflag.FlagSet) {
 	opts.FlagDescription = "set artifact platform"
-	fs.StringVarP(&opts.platform, "artifact-platform", "", "", "[Experimental] "+opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
+	fs.StringSliceVarP(&opts.platform, "artifact-platform", "", []string{}, "[Experimental] "+opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
```
shizhMSFT: In go, it is OK to have `nil` for empty slice.
```suggestion
	fs.StringSliceVarP(&opts.platform, "artifact-platform", "", nil, "[Experimental] "+opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
```

cmd/oras/internal/option/platform.go
```diff
@@ -61,7 +115,7 @@ func (opts *Platform) Parse(*cobra.Command) error {
 	case 1:
 		p.Architecture = runtime.GOARCH
 	default:
-		return fmt.Errorf("failed to parse platform %q: expected format os[/arch[/variant]]", opts.platform)
+		return fmt.Errorf("failed to parse platform %q: expected format os[/arch[/variant]]", platformStr)
```
shizhMSFT: In the error message, it should print the original message, not the modified one.

cmd/oras/internal/option/platform.go
```diff
@@ -27,8 +27,9 @@ import (
 
 // Platform option struct.
 type Platform struct {
-	platform        string
+	platform        []string
 	Platform        *ocispec.Platform
+	Platforms       []*ocispec.Platform
```
shizhMSFT, line 32: Now, it becomes interesting in the code. We have `Platform` and `Platforms` and we should just keep `Platforms` and drop `Platform`.

Since `--platform` is shared by multiple commands, it is right time to see if we are breaking other commands (i.e. have regression).
Here's the basic question: Can other commands take multiple platforms? They can? or Only some of them can? We need to double think about this.

/cc @FeynmanZhou

cmd/oras/internal/option/platform.go
```diff
@@ -37,21 +38,74 @@ func (opts *Platform) ApplyFlags(fs *pflag.FlagSet) {
...
+	if len(opts.platform) == 0 {
 		return nil
 	}
 
+	if len(opts.platform) == 1 {
+		// Single platform case - existing behavior
+		return opts.parseSinglePlatform(opts.platform[0])
+	}
+
+	return opts.parseMultiplePlatform(opts.platform)
```
shizhMSFT: By reading `parseSinglePlatform` and `parseMultiplePlatform`, obviously there are duplicated code.

We can implement `parseMultiplePlatform` by re-using most of the code from `parseSinglePlatform`.

You know what? Since `parseMultiplePlatform` is a more generic form of `parseSinglePlatform`, we can just keep the implementation of `parseMultiplePlatform` and drop `parseSinglePlatform`.

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,177 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+		}
+		availablePlatforms = append(availablePlatforms, fmt.Sprintf("%s/%s", manifest.Platform.OS, manifest.Platform.Architecture))
+		if matchesAnyPlatform(manifest.Platform, opts.Platform.Platforms) {
+			filteredManifests = append(filteredManifests, manifest)
+			matchedPlatforms[fmt.Sprintf("%s/%s", manifest.Platform.OS, manifest.Platform.Architecture)] = true
+		}
+	}
+
+	if len(filteredManifests) != len(opts.Platform.Platforms) {
+
```
shizhMSFT: nit: code style: remove empty line

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,177 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+	matchedPlatforms := make(map[string]bool)
+	for _, manifest := range index.Manifests {
+		if manifest.Platform == nil {
+			continue
+		}
+		availablePlatforms = append(availablePlatforms, fmt.Sprintf("%s/%s", manifest.Platform.OS, manifest.Platform.Architecture))
+		if matchesAnyPlatform(manifest.Platform, opts.Platform.Platforms) {
+			filteredManifests = append(filteredManifests, manifest)
+			matchedPlatforms[fmt.Sprintf("%s/%s", manifest.Platform.OS, manifest.Platform.Architecture)] = true
+		}
```
shizhMSFT, line 242: nit: code style: perf: `fmt.Sprintf("%s/%s", manifest.Platform.OS, manifest.Platform.Architecture)` appears twice. We can have a variable for that (like you did at line 229).

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,177 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+
+// matchesAnyPlatform checks if a manifest platform matches any of the specified platforms
+func matchesAnyPlatform(manifestPlatform *ocispec.Platform, platforms []*ocispec.Platform) bool {
+	for _, platform := range platforms {
+		if platformMatches(manifestPlatform, platform) {
+			return true
+		}
+	}
+	return false
+}
```
shizhMSFT, line 360: code style: modernize: Consider [slices.ContainsFunc](https://pkg.go.dev/slices#ContainsFunc)

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,177 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+	if a.Variant != "" && b.Variant != "" && a.Variant != b.Variant {
+		return false
+	}
+
+	// OSVersion is optional; only treat it as a mismatch if both OSVersions are non-empty and different.
+	if a.OSVersion != "" && b.OSVersion != "" && a.OSVersion != b.OSVersion {
+		return false
+	}
+
+	return true
```
shizhMSFT: To be rigorous, A matching B does not imply B matching A in your scenario.

For instance, specifying `--platform linux/arm` matches the `linux/arm/v6` platform in the index. However, specifying `--platform linux/arm/v6` does not match `linux/arm` platform in the index (see also [containerd](https://github.com/containerd/platforms/blob/v1.0.0-rc.2/platforms.go#L101)).

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,177 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+	if from, err := digest.Parse(opts.From.Reference); err == nil && from != newIndexDesc.Digest {
+		opts.From.RawReference = fmt.Sprintf("%s@%s", opts.From.Path, newIndexDesc.Digest.String())
+	}
+
+	if err := metadataHandler.OnCopied(&opts.BinaryTarget, newIndexDesc); err != nil {
+		return err
+	}
+
+	return metadataHandler.Render()
+}
```
shizhMSFT, line 353: Could you refactor this out to a function as the current function is getting to large.

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,177 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+
+	// Create a new index with only the filtered manifests
+	newIndex := index
+	newIndex.Manifests = filteredManifests
+
+	// Marshal the new index
+	newIndexContent, err := json.Marshal(newIndex)
+	if err != nil {
+		return fmt.Errorf("failed to marshal new index: %w", err)
+	}
```
shizhMSFT, line 282: Similarly, I think this part can also be refactored to a new function.

Besides, what if the filtered index is has the same content of the original index?
For example, if the index has `linux/arm64` and `linux/amd64` and the user specifies `--platform linux/arm64,linux/amd64`.
In that case, should we just copy the original index so that we can keep the same digest?

cmd/oras/internal/option/platform.go
hellocn9: Yes, it print the original message. 
`./oras.exe cp --platform linux/amd54/mdn/sss docker.io/library/nginx:1.29.3  myharbor.com/library/nginx:1.29.3`
Get error like:
`Error: failed to parse platform "linux/amd54/mdn/sss": expected format os[/arch[/variant]]`

cmd/oras/internal/option/platform.go
hellocn9: You're absolutely right—the ideal approach would be to keep only `Platforms` and remove `Platform`. However, as you pointed out, removing `Platform` would break other commands since they haven't been adapted yet. To maintain compatibility, we’ve temporarily kept both for now.

In the meantime, we’ve updated the `--platform` option’s help text to include: “or comma-separated list for multiple platforms (supported in oras cp only)”, clearly indicating that multi-platform support currently applies only to the `cp` command.

We’d also greatly appreciate contributions from others to help extend this support to the remaining commands!

cmd/oras/internal/option/platform.go
hellocn9: Yes, I’ve already done that — now there’s only a single, unified conversion logic in place.

cmd/oras/root/cp.go
hellocn9: Done. Thanks

cmd/oras/root/cp.go
hellocn9: Done. Thanks

cmd/oras/root/cp.go
hellocn9: Done. Thanks

cmd/oras/root/cp.go
hellocn9: You're right to point out the importance of precise matching behavior — and I’ve already updated the logic accordingly.

The current behavior is now:

- **If a `Variant` is specified** (e.g., `linux/arm/v5`), it **must match exactly** — both OS/Arch *and* Variant must align.
- **If no `Variant` is specified** (e.g., `linux/arm`), then **Variant is ignored** — only OS and Arch are matched, regardless of whether the index entry has a Variant or not.

So, for example:
- `--platform linux/arm` → matches both `linux/arm` and `linux/arm/v5` in the index
- `--platform linux/arm/v5` → matches **only** `linux/arm/v5`, and will **not** fall back to `linux/arm`

This avoids the incorrect fallback you mentioned and ensures strict, predictable matching semantics. Thanks for raising this — it’s a critical detail for platform selection accuracy!

cmd/oras/root/cp.go
hellocn9: Done. I’ve already refactored that logic into a dedicated function named `doMultipleCopy`.

cmd/oras/root/cp.go
hellocn9: Done. I’ve already refactored that logic into a dedicated function named `filterManifestByPlatform`. 

In the meantime, if the user specifies --platform with a list that exactly matches all the platforms present in the original index (e.g., index has linux/arm64 and linux/amd64, and user passes --platform linux/arm64,linux/amd64), then we treat it as a “no-op” filter — meaning we fall back to the original, unfiltered index, preserving the original digest and avoiding unnecessary re-serialization.

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,211 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+// platformMatches checks if two platforms match
+func platformMatches(manifestPlatform, targetPlatform *ocispec.Platform) bool {
+	if manifestPlatform.OS != targetPlatform.OS || manifestPlatform.Architecture != targetPlatform.Architecture {
+		return false
+	}
+
+	// Variant: optional; if specified, must match exactly, otherwise it is ignored
+	//if targetPlatform.Variant == "" && manifestPlatform.Variant != "" {
+	//	return false
+	//}
```
TerryHowe, line 371:
```suggestion
```

cmd/oras/root/cp.go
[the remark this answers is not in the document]
TerryHowe: Do you want to use CopyGraphOptions here since you are not using extended copy?

cmd/oras/root/cp.go
```diff
@@ -138,6 +142,17 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
 	ctx = registryutil.WithScopeHint(ctx, dst, auth.ActionPull, auth.ActionPush)
 	statusHandler, metadataHandler := display.NewCopyHandler(opts.Printer, opts.TTY, dst)
 
+	// Check if multiple platforms are specified
+	if len(opts.Platform.Platforms) > 1 && !opts.recursive {
```
TerryHowe, line 146: This part is confusing me here, why is this checking for not recursive? If there are multiple platforms specified don't we want to  copy the multiple platforms?

cmd/oras/internal/option/platform.go
```diff
@@ -37,40 +38,54 @@ func (opts *Platform) ApplyFlags(fs *pflag.FlagSet) {
 	if opts.FlagDescription == "" {
 		opts.FlagDescription = "request platform"
 	}
-	fs.StringVarP(&opts.platform, "platform", "", "", opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
+	fs.StringSliceVarP(&opts.platform, "platform", "", nil, opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]` or comma-separated list for multiple platforms (supported in oras cp only)")
```
TerryHowe, line 41: Unrelated to this PR, but that bit about "supported in oras cp only" is a good argument for composition over  inheritance

cmd/oras/internal/option/platform.go
```diff
@@ -82,5 +97,5 @@ type ArtifactPlatform struct {
 // ApplyFlags applies flags to a command flag set.
 func (opts *ArtifactPlatform) ApplyFlags(fs *pflag.FlagSet) {
 	opts.FlagDescription = "set artifact platform"
-	fs.StringVarP(&opts.platform, "artifact-platform", "", "", "[Experimental] "+opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
+	fs.StringSliceVarP(&opts.platform, "artifact-platform", "", nil, "[Experimental] "+opts.FlagDescription+" in the form of `os[/arch][/variant][:os_version]`")
```
TerryHowe, line 100: Why was this changed?

cmd/oras/root/cp.go
```diff
@@ -164,6 +179,211 @@ func runCopy(cmd *cobra.Command, opts *copyOptions) error {
...
+	// Copy all matching manifests and their content
+	for _, manifestDesc := range filteredManifests {
+		// Copy the manifest itself
+		if err = oras.CopyGraph(ctx, src, dst, manifestDesc, extendedCopyGraphOptions.CopyGraphOptions); err != nil {
+			return fmt.Errorf("failed to copy manifest %s: %w", manifestDesc.Digest, err)
+		}
+	}
+
+	// Push the new index to the destination
+	if err = dst.Push(ctx, newIndexDesc, strings.NewReader(string(newIndexContent))); err != nil {
```
TerryHowe, line 324:
```suggestion
	if err = dst.Push(ctx, newIndexDesc, bytes.NewReader(newIndexContent)); err != nil {
```

TerryHowe (changes requested):

TerryHowe:
Ping

polarathene:
@hellocn9 are you going to return to this PR?

TerryHowe:
Stale

TerryHowe:
Follow-up: opened #2163 as a fresh tracking issue with an up-to-date, conflict-free rebase of this work for anyone who'd like to pick it up.

PR #2000, opened 2026-03-09
feat: add --config flag support for oras attach

### LINKED ISSUE #1146: Support to customize the config data

v1.1.0

Support to customize the config descriptor such as mediaType when using ORAS to attach something to a subject artifact.

OCI does not require mandatory content for the config descriptor, so in some scenarios, users may need to define their own artifacts, which may need to customize the type of their config metadata.

ogulcanaydogan:
## Summary

Add `--config <file>[:<type>]` flag support to `oras attach`, matching the existing behavior in `oras push`. This allows users to customize the config descriptor (media type and content) when attaching artifacts to a subject.

**Use case:** OCI does not mandate config content for artifacts, so users may need to define custom config data with specific media types when attaching artifacts.

## Changes

- `cmd/oras/root/attach.go`:
  - Add `manifestConfigRef` field to `attachOptions` struct
  - Register `--config` flag with path format `<file>[:<type>]`
  - Parse config file reference using `fileref.Parse()` and add to file store
  - Set `ConfigDescriptor` and `ConfigAnnotations` in `PackManifestOptions`
  - Add mutual exclusivity check between `--config` and `--platform` flags
  - Add usage example in command help text

### Usage

```bash
# Attach with custom config file and media type
oras attach --artifact-type doc/example \
  --config config.json:application/vnd.me.config \
  localhost:5000/hello:v1 hi.txt
```

## Test plan

- [x] `go build ./cmd/oras/` — builds successfully
- [x] `go vet ./cmd/oras/...` — no issues
- [x] `go test ./...` — full test suite passes (all packages OK)

Fixes #1146

### DIFF
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -193,7 +193,6 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
 
 	packOpts := oras.PackManifestOptions{
 		Subject:             &subject,
-		ConfigAnnotations:   opts.Annotations[option.AnnotationConfig],
 		ManifestAnnotations: opts.Annotations[option.AnnotationManifest],
 		Layers:              descs,
 	}
@@ -206,7 +205,7 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
 		if err != nil {
 			return err
 		}
-		desc.Annotations = packOpts.ConfigAnnotations
+		desc.Annotations = opts.Annotations[option.AnnotationConfig]
 		packOpts.ConfigDescriptor = &desc
 	}
 	pack := func() (ocispec.Descriptor, error) {
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -30,6 +30,7 @@ import (
 	"oras.land/oras/cmd/oras/internal/command"
 	"oras.land/oras/cmd/oras/internal/display"
 	oerrors "oras.land/oras/cmd/oras/internal/errors"
+	"oras.land/oras/cmd/oras/internal/fileref"
 	"oras.land/oras/cmd/oras/internal/option"
 	"oras.land/oras/internal/graph"
 	"oras.land/oras/internal/registryutil"
@@ -43,8 +44,9 @@ type attachOptions struct {
 	option.Platform
 	option.Terminal
 
-	artifactType string
-	concurrency  int
+	artifactType      string
+	manifestConfigRef string
+	concurrency       int
 	// Deprecated: verbose is deprecated and will be removed in the future.
 	verbose bool
 }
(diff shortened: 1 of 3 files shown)


cmd/oras/root/attach.go
```diff
@@ -184,9 +193,22 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
 
 	packOpts := oras.PackManifestOptions{
 		Subject:             &subject,
+		ConfigAnnotations:   opts.Annotations[option.AnnotationConfig],
```
TerryHowe: Seems like this annotation should only be set if `opts.manifestConfigRef != ""`

cmd/oras/root/attach.go
```diff
@@ -121,6 +129,7 @@ Example - Attach file to the manifest tagged 'example.com:v1' in an OCI image la
 	}
 
 	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
+	cmd.Flags().StringVarP(&opts.manifestConfigRef, "config", "", "", "`path` of image config file")
```
TerryHowe, line 132: Is push like this where it doesn't talk about `file[:<type>]`?

TerryHowe (changes requested):
This could use some tests. It may be easier to add to test/e2e/suite/command/attach.go rather than unit tests

cmd/oras/root/attach.go
ogulcanaydogan: Fixed -- moved `ConfigAnnotations` inside the `if opts.manifestConfigRef != ""` block so it's only set when a config file is actually provided.

cmd/oras/root/attach.go
ogulcanaydogan: Yes, push uses the same description: ``path`` of image config file`, without mentioning the `file[:<type>]` form. I've kept it consistent with push.

ogulcanaydogan:
Rebased onto latest main (as of today). All previous review comments have been addressed:
- Added 3 e2e tests in `test/e2e/suite/command/attach.go`
- `ConfigAnnotations` is only set when `manifestConfigRef != ""`
- Flag description is consistent with `oras push`

@TerryHowe ready for re-review when you have a chance, thank you!

TerryHowe (approved):
/lgtm

TerryHowe:
looks like the last commit is missing signoff

ogulcanaydogan:
All commits have DCO sign-off and CI is green. Ready to merge when you have a moment — thanks @TerryHowe!

ogulcanaydogan:
Rebased onto main — all checks green, DCO clean. Ready to merge when you have a moment, @TerryHowe.

ogulcanaydogan:
CI runs need maintainer approval to execute (fork PR restriction). @TerryHowe if you can approve the workflow runs, that should unblock the merge.

ogulcanaydogan:
Friendly bump — CI is green and all feedback has been addressed.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #2037, opened 2026-04-22
cmd/oras/root/attach: pass --annotation $config:* through to PackManifestOptions

### LINKED ISSUE #2022: oras attach ignores --annotation $config:key=val when --config is not provided
## Description

`oras attach` silently drops config annotations set via `--annotation $config:key=val` when the `--config` flag is not also provided.

This is because `runAttach` never sets `packOpts.ConfigAnnotations`, so oras-go generates the default config without applying the requested annotations.

In contrast, `oras push` unconditionally sets:

```go
packOpts := oras.PackManifestOptions{
    ConfigAnnotations: opts.Annotations[option.AnnotationConfig],
    ...
}
```

so config annotations are applied to the auto-generated config even when no `--config` flag is given.

## Steps to reproduce

```bash
oras attach --artifact-type doc/example \
  --annotation '$config:org.example.key=value' \
  localhost:5000/hello:v1 hi.txt
```

Fetch the resulting manifest — the config descriptor will have no annotations, even though `--annotation $config:...` was specified.

## Expected behavior

Config annotations should be applied to the default config descriptor, consistent with `oras push` behavior.

## Fix

In `runAttach` (`cmd/oras/root/attach.go`), set `ConfigAnnotations` on `packOpts` unconditionally:

```go
packOpts := oras.PackManifestOptions{
    Subject:             &subject,
    ConfigAnnotations:   opts.Annotations[option.AnnotationConfig],
    ManifestAnnotations: opts.Annotations[option.AnnotationManifest],
    Layers:              descs,
}
```

SAY-5:
## Summary

Fixes #2022.

`runAttach` built `oras.PackManifestOptions` without a `ConfigAnnotations` entry, so `--annotation $config:key=val` was silently dropped whenever the user did not also pass `--config`: oras-go generated the default config descriptor and never saw the annotations. `runPush` already wires `ConfigAnnotations` unconditionally, so this was strictly an attach-path oversight.

## Fix

Add the same `ConfigAnnotations: opts.Annotations[option.AnnotationConfig]` line to the attach packOpts. When the caller did not pass any `$config:*` annotations the map lookup returns nil and PackManifest keeps its current default-config behaviour.

## Test plan

- [x] `go build ./cmd/oras/...` clean
- [x] `go vet ./cmd/oras/...` clean
- [x] `go test ./cmd/oras/... ./internal/...` green
- [ ] Manual: `oras attach --artifact-type doc/example --annotation '$config:org.example.key=value' localhost:5000/hello:v1 hi.txt` produces a manifest whose config descriptor carries `{"org.example.key": "value"}`.

### DIFF
--- a/cmd/oras/root/attach_test.go
+++ b/cmd/oras/root/attach_test.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"testing"
 
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 
 	"oras.land/oras/cmd/oras/internal/option"
@@ -75,10 +76,9 @@ func Test_runAttach_errType(t *testing.T) {
 	}
 }
 
-func Test_attachCmd_configAnnotationsWiredFromFile(t *testing.T) {
+func Test_attachCmd_configAnnotationsWiredToPackOpts(t *testing.T) {
 	// Regression for #2022: config annotations from an annotation file must
-	// be available as opts.Annotations[option.AnnotationConfig] after parsing
-	// so that runAttach can pass them to oras.PackManifestOptions.ConfigAnnotations.
+	// flow through to oras.PackManifestOptions.ConfigAnnotations.
 	annotationFile := filepath.Join(t.TempDir(), "annot.json")
 	content := `{"$config":{"org.example.key":"val"}}`
 	if err := os.WriteFile(annotationFile, []byte(content), 0600); err != nil {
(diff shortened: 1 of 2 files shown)


TerryHowe (commented):
Is there any way to add a test for this?

cmd/oras/root/attach_test.go
```diff
@@ -72,3 +74,27 @@ func Test_runAttach_errType(t *testing.T) {
 		t.Fatalf("got %v, want %v", got, want)
 	}
 }
+
```
TerryHowe, line 78: This doesn't really test the config annotations. Maybe something like:

```
  In attach.go, pull the struct literal into a helper:

  func buildAttachPackOpts(opts *attachOptions, subject ocispec.Descriptor, descs []ocispec.Descriptor)
  oras.PackManifestOptions {
        return oras.PackManifestOptions{
                Subject:             &subject,
                ConfigAnnotations:   opts.Annotations[option.AnnotationConfig],
                ManifestAnnotations: opts.Annotations[option.AnnotationManifest],
                Layers:              descs,
        }
  }

  and call it from runAttach. Then the test asserts on the field that actually feeds PackManifest:

  func Test_buildAttachPackOpts_configAnnotations(t *testing.T) {
        annotationFile := filepath.Join(t.TempDir(), "annot.json")
        content := `{"$config":{"org.example.key":"val"}}`
        if err := os.WriteFile(annotationFile, []byte(content), 0600); err != nil {
                t.Fatal(err)
        }
        opts := &attachOptions{}
        opts.Packer.AnnotationFilePath = annotationFile
        if err := opts.Packer.Parse(&cobra.Command{}); err != nil {
                t.Fatalf("unexpected error: %v", err)
        }

        packOpts := buildAttachPackOpts(opts, ocispec.Descriptor{}, nil)
        if packOpts.ConfigAnnotations["org.example.key"] != "val" {
                t.Fatalf("ConfigAnnotations not wired: %v", packOpts.ConfigAnnotations)
        }
  }
```

TerryHowe (approved):
/lgtm

TerryHowe:
uggh, lint error. Please rebase too.

TerryHowe (approved):
/lgtm

Issue #2008, opened 2026-03-18
feat: migrate ORAS CLI from oras-go v2 to v3


TerryHowe:
## Overview

This issue tracks the full migration of the ORAS CLI from `oras.land/oras-go/v2` to `github.com/oras-project/oras-go/v3`. The migration is broken into three phases:

1. **Mechanical API migration** — import path and breaking API renames (PR #2002)
2. **Deprecation cleanup** — remove duplicated internal code now first-class in v3
3. **New feature adoption** — unlock user-facing capabilities provided by v3

---

## Motivation

oras-go v3 introduces a new module path, several breaking API renames, and new packages that provide capabilities the ORAS CLI currently either lacks entirely or implements through duplicated custom code. The primary user-facing motivation is support for the standard [containers configuration file stack](https://github.com/containers/image/tree/main/docs):

### containers/image configuration files supported by oras-go v3

| Config File | Location | Benefit to ORAS users |
|---|---|---|
| [`containers-auth.json`](https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md) | `~/.config/containers/auth.json` | Podman/Buildah/Skopeo credentials are read automatically alongside Docker `config.json` — no `--registry-config` flag needed |
| [`containers-registries.conf`](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md) | `/etc/containers/registries.conf` | Registry mirrors, blocked registries, and location remapping are applied automatically on every pull — critical for air-gapped and enterprise deployments |
| [`containers-certs.d`](https://github.com/containers/image/blob/main/docs/containers-certs.d.5.md) | `/etc/containers/certs.d/<host>/` | Per-registry CA certificates and client TLS certs are loaded without `--ca-file` flags |

Beyond the config-file ecosystem, full v3 adoption also:

- Fixes a **process-safety bug** in the `ORAS_CACHE` path (`oci.New()` is not safe for concurrent process writes)
- Removes ~300 lines of duplicated internal code (`internal/cache/`, `internal/trace/transport.go`)
- Adds **structured error messages** for `oras cp` that distinguish source from destination failures

---

### Phase 1 — Mechanical API Migration (PR #2002)

> Status: **open (draft)**

Replace all `oras.land/oras-go/v2` imports with `github.com/oras-project/oras-go/v3` and update all breaking call sites.

| Change | Status |
|---|---|
[...]

PR #2002, opened 2026-03-11
feat: migrate from oras-go v2 to v3

TerryHowe:
## Summary

- Replace all `oras.land/oras-go/v2` imports with `github.com/oras-project/oras-go/v3`
- Update API call sites for breaking changes in v3 (credential types, repository struct fields, reference access, login/logout functions)
- Add blank import of `registry/remote/config` package to register the config loader required by v3

## Key API Changes

| v2 | v3 |
|---|---|
| `auth.Credential{}` | `credentials.Credential{}` |
| `auth.EmptyCredential` | `credentials.EmptyCredential` |
| `client.Credential =` | `client.CredentialFunc =` |
| `credentials.Credential(store)` | `store.Get` |
| `credentials.Login/Logout` | `remote.Login/Logout` |
| `repo.Client/PlainHTTP/HandleWarning` | `repo.Registry.Client/PlainHTTP/HandleWarning` |
| `repo.Reference.Registry` | `repo.Registry.Reference.Registry` |
| `repo.Reference.Repository` | `repo.RepositoryName` |
| `repo.Reference` (field) | `repo.Reference()` (method) |
| `ref.Reference` (field) | `ref.GetReference()` |
| `SetReferrersCapability()` returns error | no return value |

## Test plan

- [x] `go build ./...` succeeds
- [x] `go test ./...` all pass
- [x] `go vet ./...` clean

Closes: https://github.com/oras-project/oras/issues/2118

PR #2039, opened 2026-04-25
Add flag to make manifest and tar reproducible

### LINKED ISSUE #1464: make `oras push` and `oras attach` deterministic

v1.2.0

Deterministically generate manifests for oras push and oras attach if the same content (e.g. blobs, annotations) are packed.

Related issue: https://github.com/oras-project/oras-go/issues/748, https://github.com/oras-project/oras-www/issues/366

If the to-be uploaded file is a folder, ORAS will pack the folder as a tarball archive. The last modified time(mtime) is include in the archive so the digest of the packed tarball changes even when file content are identical. `oras` CLI should provide a flag to strip out the time info so the packing is deterministic.

Related PR: https://github.com/oras-project/oras/pull/126

With deterministic builds (a.k.a. [reproducible builds](https://en.wikipedia.org/wiki/Reproducible_builds)), the oras push command will not push two different manifests. Deterministic builds also play an important role in CSSC (see [blog](https://blogs.vmware.com/opensource/2022/07/12/what-makes-a-build-reproducible-part-1/)).

afbjorklund:
This will set the timestamps to the zero time.Time timestamp, to avoid it being set to the current/modification timestamp.

For the manifest this translates to 0001-01-01, but for the tarball it translates to 1970-01-01 (since it uses seconds).

Adds `--reproducible`  flag, to push and attach.
This will make their resulting digests deterministic.

Fixes #1464

### DIFF
--- a/cmd/oras/root/attach.go
+++ b/cmd/oras/root/attach.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
@@ -47,6 +48,7 @@ type attachOptions struct {
 	artifactType      string
 	manifestConfigRef string
 	concurrency       int
+	reproducible      bool
 	// Deprecated: verbose is deprecated and will be removed in the future.
 	verbose bool
 }
@@ -134,6 +136,7 @@ Example - Attach file 'hi.txt' with the custom manifest config "config.json" of 
 	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", true, "print status output for unnamed blobs")
 	opts.FlagDescription = "attach to an arch-specific subject"
 	_ = cmd.MarkFlagRequired("artifact-type")
+	cmd.Flags().BoolVarP(&opts.reproducible, "reproducible", "", false, "make it reproducible")
 	_ = cmd.Flags().MarkDeprecated("verbose", "and will be removed in a future release.")
 	opts.EnableDistributionSpecFlag()
 	opts.SetTypes(option.FormatTypeText, option.FormatTypeJSON, option.FormatTypeGoTemplate)
@@ -217,6 +220,19 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
 	}
 	pack := func() (ocispec.Descriptor, error) {
 		return oras.PackManifest(ctx, store, oras.PackManifestVersion1_1, opts.artifactType, packOpts)
+	}
+
+	if opts.reproducible {
+		// make manifest reproducible
+		annotations := packOpts.ManifestAnnotations
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		zero := time.Time{}
+		annotations[ocispec.AnnotationCreated] = zero.Format(time.RFC3339)
+		packOpts.ManifestAnnotations = annotations
+
+		store.TarReproducible = true
 	}
 
 	copyFunc := func(root ocispec.Descriptor) error {
--- a/cmd/oras/root/push.go
+++ b/cmd/oras/root/push.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"strings"
+	"time"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
@@ -53,6 +54,7 @@ type pushOptions struct {
 	manifestConfigRef string
 	artifactType      string
 	concurrency       int
+	reproducible      bool
 	// Deprecated: verbose is deprecated and will be removed in the future.
 	verbose bool
 }
@@ -170,6 +172,7 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 	cmd.Flags().StringVarP(&opts.manifestConfigRef, "config", "", "", "`path` of image config file")
 	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
 	cmd.Flags().IntVarP(&opts.concurrency, "concurrency", "", 5, "concurrency level")
+	cmd.Flags().BoolVarP(&opts.reproducible, "reproducible", "", false, "make it reproducible")
 	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", true, "print status output for unnamed blobs")
 	_ = cmd.Flags().MarkDeprecated("verbose", "and will be removed in a future release.")
 	opts.SetTypes(option.FormatTypeText, option.FormatTypeJSON, option.FormatTypeGoTemplate)
@@ -190,6 +193,20 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
 		return err
 	}
 	defer func() { _ = store.Close() }()
+
+	if opts.reproducible {
+		// make manifest reproducible
+		annotations := packOpts.ManifestAnnotations
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		zero := time.Time{}
+		annotations[ocispec.AnnotationCreated] = zero.Format(time.RFC3339)
+		packOpts.ManifestAnnotations = annotations
+
+		store.TarReproducible = true
+	}
+
 	if opts.manifestConfigRef != "" {
 		path, cfgMediaType, err := fileref.Parse(opts.manifestConfigRef, oras.MediaTypeUnknownConfig)
 		if err != nil {


cmd/oras/root/attach.go
```diff
@@ -158,6 +161,19 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
 	}
 	defer func() { _ = store.Close() }()
 
+	if opts.reproducible {
+		// make manifest reproducible
+		annotations := packOpts.ManifestAnnotations
```
TerryHowe: packOpts doesn't seem to exist here

cmd/oras/root/push.go
```diff
@@ -190,6 +193,20 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
...
 	defer func() { _ = store.Close() }()
+
+	if opts.reproducible {
+		// make manifest reproducible
+		annotations := packOpts.ManifestAnnotations
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		zero := time.Time{}
+		annotations[ocispec.AnnotationCreated] = zero.Format(time.RFC3339)
```
TerryHowe, line 204: It is standard to use environment variable SOURCE_DATE_EPOCH for date. If the Reproducible Build standard says it should be used if set, we could get rid of the flag. I think that is the case, but that would make this a breaking change (may not be a bad thing).

TerryHowe (changes requested):

cmd/oras/root/push.go
afbjorklund: Unfortunately the store is not that flexible, it only takes a boolean `TarReproducible`

afbjorklund:
```go
        // TarReproducible controls if the tarballs generated
        // for the added directories are reproducible.
        // When specified, some metadata such as change time
        // will be removed from the files in the tarballs. Default value: false.
        TarReproducible bool
```

```go
                if removeTimes {
                        header.ModTime = time.Time{}
                        header.AccessTime = time.Time{}
                        header.ChangeTime = time.Time{}
                }
```

afbjorklund: So it would need a change in oras-go (for tar), if you wanted to make a proper fix.

For instance when using squashfs-tools, it will "clamp" the timestamp to the SDE.
If the timestamp is newer, it will be set back to the SourceDateEpoch (not to zero)
If the timestamp is older, it is kept as-is (not zero) since it will be reproducible.

cmd/oras/root/attach.go
afbjorklund: Right, it was defined further down

cmd/oras/root/push.go
afbjorklund: Another option would be to fix oras-go, to not add timestamps when not provided.

That depends on the definition of an "artifact", and of the "created" of an artifact.
It is the date when the JSON was created, like today? Or is it the about the aritfact?
And how do you find out when a generic artifact (file) was created, not just modified

TerryHowe: We'd need to fix oras-go then although I thought oras-go was set up correctly for this. I might be thinking of what is on main in oras-go though instead of what is on v2

TerryHowe: I creeated https://github.com/oras-project/oras-go/issues/1252 to track this

afbjorklund: The easiest fix would be to revert the addition of a bogus timestamp to each artifact:

```go
        now := time.Now().UTC()
        copied[annotationCreatedKey] = now.Format(time.RFC3339)
```

Then it would be reproducible, without using `--reproducible` or SourceDateEpoch

https://github.com/oras-project/oras/pull/2039#discussion_r3604798222

TerryHowe:
Following up on the timestamp discussion in https://github.com/oras-project/oras/pull/2039#discussion_r3564468183 — I went and audited the library side, and I think the picture has changed since that thread started.

**The oras-go work is done, on `main` (v3).** oras-project/oras-go#1252 was closed by oras-project/oras-go#1289, which added timestamp clamping to `content/file`:

```go
// TarModTime, when not nil, sets the upper bound for the timestamps in the
// tarballs generated for the added directories. Any entry with a modification
// time later than TarModTime is clamped down to TarModTime, while entries
// older than it are left unchanged.
TarModTime *time.Time
```

That is exactly the `SOURCE_DATE_EPOCH` clamp semantics @afbjorklund described (newer entries pulled back, older entries kept as-is, rather than everything zeroed), and `tarDirectory` also normalizes `Uid`/`Gid`/`Uname`/`Gname`. It shipped in `v3.0.0-rc.1`, so no pseudo-version is needed to consume it.

**Nothing else in the library needs changing.** There is exactly one `time.Now()` left in non-test code — `pack.go` in `ensureAnnotationCreated` — and it already yields to a caller-supplied value:

```go
if createdTime, ok := annotations[annotationCreatedKey]; ok {
    // validate RFC 3339, return as-is
}
```

So the manifest half needs no library change at all, and we should *not* revert that `time.Now()` line: it would be a breaking behavior change for every oras-go consumer, and `org.opencontainers.image.created` is legitimate OCI metadata. Between those two knobs, plus lexical `filepath.Walk` ordering and a zero gzip MTIME, a push is fully reproducible.

**The blocker is the CLI, not the library.** This repo is still on `oras.land/oras-go/v2 v2.6.2`, and the v2 line only has the old `TarReproducible bool` (zero everything, no clamping). I would rather not backport this to v2, so I think this PR is best sequenced behind #2008 (migrate the CLI from oras-go v2 to v3). Note that is a module-path move (`oras.land/oras-go/v2` -> `github.com/oras-project/oras-go/v3`), not a version bump.

**What the CLI code then looks like.** Resolve a single timestamp and use it in both places:

```go
// SOURCE_DATE_EPOCH if set, else the zero time
if _, ok := packOpts.ManifestAnnotations[ocispec.AnnotationCreated]; !ok {
    packOpts.ManifestAnnotations[ocispec.AnnotationCreated] = t.Format(time.RFC3339)
}
store.TarModTime = &t
```

A few suggestions for when this is rebased onto v3:

- Keep the `SOURCE_DATE_EPOCH` parsing here in the CLI rather than in oras-go — the library taking a `*time.Time` is the right boundary.
- Only set the created annotation if the user has not already supplied one via `--annotation` / `--annotation-file`; an explicit user value should win. (Worth noting `oras push --annotation org.opencontainers.image.created=...` already produces a reproducible manifest today; the flag's real value-add is the tarball side plus the ergonomics.)
- Move the flag and the logic into `option.Packer` — both `push` and `attach` embed it, which removes the duplicated block in the current diff and gives one testable place for the env parsing.
- Nits: `BoolVarP` with an empty shorthand should be `BoolVar`, and `"make it reproducible"` needs a fuller `--help` description.

Finally, on my original question of whether the flag should exist at all, or whether the presence of `SOURCE_DATE_EPOCH` should just drive this on its own: against v2 that was a breaking change to existing digests, which is why I hesitated. If this lands in the same release as the v3 migration, we get to pick the spec-compliant behavior without a separate break — so it is worth deciding deliberately rather than defaulting to the flag.

TerryHowe:
To make the prerequisite concrete: **depends on #2002** (`feat: migrate from oras-go v2 to v3`), which is the PR implementing #2008.

#2002 pins exactly the version this feature needs:

```
github.com/oras-project/oras-go/v3 v3.0.0-rc.1
```

and `v3.0.0-rc.1` is the tag that contains `file.Store.TarModTime` (oras-project/oras-go#1289), so once #2002 merges everything required here is available with no further library or dependency work.

Suggested sequencing: hold this PR until #2002 lands, then rebase onto it and switch `store.TarReproducible = true` over to `store.TarModTime = &t`.

cmd/oras/root/attach.go
```diff
@@ -219,6 +222,19 @@ func runAttach(cmd *cobra.Command, opts *attachOptions) error {
...
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		zero := time.Time{}
+		annotations[ocispec.AnnotationCreated] = zero.Format(time.RFC3339)
+		packOpts.ManifestAnnotations = annotations
+
+		store.TarReproducible = true
+	}
```
TerryHowe, line 236: **This is a silent no-op for the tar half.** `store.TarReproducible` is set here, but `loadFiles(ctx, store, ...)` already ran at line 194. Tarring is eager: `store.Add` -> `descriptorFromDir` -> `tarDirectory(ctx, dir, name, tw, s.TarReproducible, buf)` (oras-go v2.6.2 `content/file/file.go:433,539`) computes the digest at `Add` time. By the time the flag flips, the tar and its digest already exist with real timestamps, so `oras attach --reproducible <dir>` still produces a non-deterministic digest -- with no error and no warning. `push.go` gets this right (its block is before `loadFiles`), which makes the asymmetry easy to miss.

`packOpts` isn't in scope up there, so this needs splitting rather than moving: put `store.TarReproducible` right after `file.New`/`defer store.Close()` (~line 173), and keep the annotation half here.

```go
	// right after defer func() { _ = store.Close() }()
	if opts.reproducible {
		store.TarReproducible = true
	}
```

Then here (also fixes the clobber + in-place mutation noted on `push.go`; needs a `maps` import):

```suggestion
	if opts.reproducible {
		// make manifest reproducible; store.TarReproducible is set earlier,
		// before loadFiles, because tarring happens eagerly at Add time
		annotations := maps.Clone(packOpts.ManifestAnnotations)
		if annotations == nil {
			annotations = map[string]string{}
		}
		if _, ok := annotations[ocispec.AnnotationCreated]; !ok {
			annotations[ocispec.AnnotationCreated] = time.Unix(0, 0).UTC().Format(time.RFC3339)
		}
		packOpts.ManifestAnnotations = annotations
	}
```

cmd/oras/root/attach.go
```diff
@@ -134,6 +136,7 @@ Example - Attach file 'hi.txt' with the custom manifest config "config.json" of
 	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", true, "print status output for unnamed blobs")
 	opts.FlagDescription = "attach to an arch-specific subject"
 	_ = cmd.MarkFlagRequired("artifact-type")
+	cmd.Flags().BoolVarP(&opts.reproducible, "reproducible", "", false, "make it reproducible")
```
TerryHowe, line 139: Same nit as in `push.go`: `BoolVar` rather than `BoolVarP` with an empty shorthand, and a description that says what the flag does.

```suggestion
	cmd.Flags().BoolVar(&opts.reproducible, "reproducible", false, "use a fixed creation timestamp so the attached manifest and tarballs are byte-for-byte reproducible")
```

cmd/oras/root/push.go
```diff
@@ -190,6 +193,20 @@ func runPush(cmd *cobra.Command, opts *pushOptions) error {
...
+		packOpts.ManifestAnnotations = annotations
+
+		store.TarReproducible = true
+	}
```
TerryHowe, line 208: Three things in this block:

1. **Clobbers a user-supplied `created`.** `ensureAnnotationCreated` (oras-go v2.6.2 `pack.go`) already honors a caller-provided `org.opencontainers.image.created` and only falls back to `time.Now()` when absent -- so `oras push --annotation "org.opencontainers.image.created=2020-01-01T00:00:00Z"` is deterministic today, and this overwrites that explicit value with year 1. Only set it when the key is absent.
2. **Mutates the caller's map in place.** `packOpts.ManifestAnnotations` aliases `opts.Annotations[option.AnnotationManifest]`; when non-nil this writes into that shared map rather than a copy. Benign in the current single-pass flow, but it's an action-at-a-distance edit of parsed user input. `maps.Clone` first, matching what `ensureAnnotationCreated` does.
3. **Zero time vs. the epoch.** `time.Time{}` gives an internally inconsistent artifact: the manifest reads `0001-01-01T00:00:00Z` while the tar headers read `1970-01-01T00:00:00Z`, because Go's tar writer encodes a zero `ModTime` as epoch 0 (verified: round-trips as `unix=0`, USTAR). Consumers reading a build date get two different answers depending on where they look. Epoch 0 is also the conventional `SOURCE_DATE_EPOCH` zero value -- and once this is on v3, `t` should come from `$SOURCE_DATE_EPOCH` when set and feed both the annotation and `store.TarModTime`.

```suggestion
	if opts.reproducible {
		// make manifest reproducible
		annotations := maps.Clone(packOpts.ManifestAnnotations)
		if annotations == nil {
			annotations = map[string]string{}
		}
		if _, ok := annotations[ocispec.AnnotationCreated]; !ok {
			annotations[ocispec.AnnotationCreated] = time.Unix(0, 0).UTC().Format(time.RFC3339)
		}
		packOpts.ManifestAnnotations = annotations

		store.TarReproducible = true
	}
```

(needs a `maps` import; `time` is already there)

cmd/oras/root/push.go
```diff
@@ -170,6 +172,7 @@ Example - Push file "hi.txt" into an OCI image layout folder 'layout-dir' with t
 	cmd.Flags().StringVarP(&opts.manifestConfigRef, "config", "", "", "`path` of image config file")
 	cmd.Flags().StringVarP(&opts.artifactType, "artifact-type", "", "", "artifact type")
 	cmd.Flags().IntVarP(&opts.concurrency, "concurrency", "", 5, "concurrency level")
+	cmd.Flags().BoolVarP(&opts.reproducible, "reproducible", "", false, "make it reproducible")
```
TerryHowe, line 175: Nit: `BoolVarP` with an empty shorthand should be `BoolVar`, and `"make it reproducible"` doesn't tell the user what it does or what it costs.

```suggestion
	cmd.Flags().BoolVar(&opts.reproducible, "reproducible", false, "use a fixed creation timestamp so the pushed manifest and tarballs are byte-for-byte reproducible")
```

TerryHowe (commented):
Re-reviewed against the current head (`d9eeeed`). Two structural notes that don't belong on any single line:

**Duplicated logic.** The blocks in `push.go` and `attach.go` are byte-identical apart from the receiver, and both `pushOptions` and `attachOptions` embed `option.Packer`. That is where a `Reproducible bool`, its `ApplyFlags` registration, and a small `applyTo(packOpts *oras.PackManifestOptions, store *file.Store)` helper belong (`cmd/oras/internal/option/packer.go`). It also gives the flag a unit-test home next to `packer_test.go`, and makes the ordering bug below impossible to reintroduce.

**No tests.** The diff is two files, neither a test. The natural e2e assertion is the one the feature promises: push the same fixture twice with `--reproducible` and assert identical digests, and the same for `attach` -- which is exactly where the bug is. The checklist's test box is still unchecked.

Sequencing is covered in my two comments above: this is best rebased onto #2002 (oras-go v3.0.0-rc.1), at which point `store.TarReproducible = true` becomes `store.TarModTime = &t` and the `SOURCE_DATE_EPOCH` point below can be implemented properly with clamping.

afbjorklund:
I still think the best would be if oras-go stopped inserting bogus timestamps in the created artifacts, then they wouldn't _have_ to be overwritten with values like 0 (but it is a nice feature to be able to provide created time).

```go
        now := time.Now().UTC()
        copied[annotationCreatedKey] = now.Format(time.RFC3339)
```

TerryHowe:
Beyond what https://github.com/oras-project/oras-go/pull/1289/changes does?

afbjorklund:
That is the nice new feature, but still requires work from the user to work around what is essentially a bug in the old code... Now you need to set SOURCE_DATE_EPOCH, instead of the old workaround to add a zero annotation.

But there is no way to just _remove_ the `org.opencontainers.artifact.created` annotation, when the information  is not available? The common artifact does **not** store a created date*, we fixed it for SIF by adding it to metadata.

\* Theoretically one could use the "birth" date.

But it is not very well supported, in file systems:

https://github.com/djherbis/times

And it doesn't "follow" when you copy it, etc.

PR #2062, opened 2026-05-27
chore: replace containerd/console with golang.org/x/term

### LINKED ISSUE #1449: containerd/console is broken for mac consider creack/pty

TerryHowe:
## Summary
- Replaces the effectively-unmaintained `github.com/containerd/console` (notably a long-standing macOS bug) with `golang.org/x/term`, maintained by the Go team and working on linux, mac and windows.
- The `console` package now renders to an internal `terminal` seam backed by `x/term`, keeping the terminal library isolated behind an interface so the backend can be swapped or extended later without touching callers. Cursor/escape-code logic is unchanged.
- Test helpers no longer create a real pty (`containerd.NewPty`); they use `os.Pipe` via `testutils.NewPipe`/`MatchPipe`, so the previously linux-only TTY tests now run on **all platforms** (no creack/pty, no ConPTY, no x/sys).

## Note for reviewers
Terminal detection already lives in `option.Terminal.Parse` (gating on `term.IsTerminal(os.Stderr)` before any TTY handler is constructed). The previous containerd-based `NewConsole` also rejected non-terminal files as defense-in-depth; that check is now redundant, so `NewConsole` rejects only a `nil` file. The three `*_invalidTTY` handler tests were updated to pass `nil` accordingly. Happy to thread an injectable console through the handler chain instead if you'd prefer to keep the non-terminal rejection.

Closes: #1449

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` for affected packages, now passing on macOS (previously skipped)

### DIFF
--- a/cmd/oras/internal/display/status/console/console.go
+++ b/cmd/oras/internal/display/status/console/console.go
@@ -16,10 +16,12 @@ limitations under the License.
 package console
 
 import (
+	"errors"
+	"io"
 	"os"
 
-	containerd "github.com/containerd/console"
 	"github.com/morikuni/aec"
+	"golang.org/x/term"
 )
 
 const (
@@ -33,9 +35,8 @@ const (
 	Restore = "\0338"
 )
 
-// Console is a wrapper around containerd's Console and ANSI escape codes.
+// Console is a wrapper around a terminal and ANSI escape codes.
 type Console interface {
-	containerd.Console
 	GetHeightWidth() (height, width int)
 	Save()
 	NewRow()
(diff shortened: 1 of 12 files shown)


TerryHowe:
@renshao if you get a chance, maybe you could functionally test this on Windows

renshao:
> @renshao if you get a chance, maybe you could functionally test this on Windows

Yes, I can. I will report back my test result.

renshao:
I tested 3 scenarios (push, pull, cp) in Windows 11 PowerShell Terminal. I first compiled the binary `oras.exe` using:
```
go build -o oras.exe .\cmd\oras
```

## `oras push`
image

## `oras pull`
image

## `oras cp`
image

The terminal UI looks fine, same as previous oras. The change looks good.

renshao (approved):

TerryHowe:
Thanks for testing @renshao

shizhMSFT (approved):
LGTM. The change looks clean.

PR #2079, opened 2026-06-10
docs: add examples to logout command help text

ogulcanaydogan:
The `logout` command's `Long` description had a single bare example while its sibling `login` command has six richly-labelled examples. Add a second example showing the `--registry-config` flag, and update the existing example label to match the `login` style (`Example - Log out from a registry:`).


TerryHowe (approved):
/lgtm

PR #2083, opened 2026-06-14
option: support layer-level annotations via --annotation flag

### LINKED ISSUE #2031: Support layer-level annotations via --annotation flag in oras push

1.3.0

Extend --annotation to support an optional file prefix that binds the annotation to a specific layer descriptor, using a clearly documented syntax. For example:

```
# manifest-level (existing behavior)
--annotation "key=value"

# layer-level (proposed)
--annotation "filename.ex5:key=value"
```

The --annotation flag only supports manifest-level annotations:

```bash
Flags:
  -a, --annotation stringArray                               manifest annotations
```

There is no inline CLI syntax to bind an annotation to a specific layer. The [official docs](https://oras.land/docs/commands/oras_push/) show only this usage for manifest annotations:

```bash
oras push --annotation "key=val" localhost:5000/hello:v1
```

Layer-level annotations are exclusively supported via --annotation-file, which requires generating a temporary JSON file

Problem

In scripting and CI/CD scenarios where layer annotations need to be set dynamically per file, users are forced to generate, manage, and clean up temporary annotation files — just to bind a single annotation to a layer. This is cumbersome and error-prone, especially in batch push loops.
Additionally, the lack of a CLI-native syntax invites users to attempt --annotation "<filename>:<key>=<value>" intuitively, which silently produces incorrect results: instead of binding to the layer it dumps a literal key <filename>:<key> into the manifest annotations, making it very hard to diagnose.

BRGOVIND:
﻿Closes #2031 — extends `--annotation` to accept layer-level annotations directly, without needing a temporary `--annotation-file`.

## Problem

`oras push --annotation` currently only supports manifest-level annotations.
To set layer-level annotations users must create a temporary JSON file and pass it with `--annotation-file`:

```json
{
  "": {"org.opencontainers.image.title": "demo"},
  "myfile.tar": {"com.example.layer": "data"}
}
```

This is cumbersome in CI/CD pipelines and scripting.

## Solution

Extend the `--annotation` flag to accept an optional file prefix:

| Syntax | Effect |
|--------|--------|
| `--annotation "key=value"` | Manifest-level (existing behavior, unchanged) |
| `--annotation "filename.tar:key=value"` | Layer-level, bound to that filename |

When the key part (before `=`) contains a colon, everything before the colon is treated as the file reference and everything after as the annotation key. OCI annotation keys use reverse-DNS with dots (not colons), so `:`) is safe as a separator.

Duplicate key detection is applied per target (manifest or per-file). The resulting `Annotations` map mirrors the `--annotation-file` JSON format exactly, so no changes are needed downstream in the push path.

## Example

```sh
oras push localhost:5000/demo \
  --annotation "org.opencontainers.image.title=demo" \
  --annotation "myfile.tar:com.example.layer=data"
```

### DIFF
--- a/cmd/oras/internal/option/annotation.go
+++ b/cmd/oras/internal/option/annotation.go
@@ -28,40 +28,80 @@ import (
 var (
 	errAnnotationFormat      = errors.New("annotation value doesn't match the required format")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
+	errAnnotationLayerKey    = errors.New("annotation key must not be empty after the file prefix")
 )
 
 // Annotation option struct.
 type Annotation struct {
-	// ManifestAnnotations contains raw input of manifest annotation "key=value" pairs
+	// ManifestAnnotations contains raw input annotation flags.
+	// Two formats are accepted:
+	//   "key=value"               → manifest-level annotation
+	//   "filename:key=value"      → layer-level annotation bound to filename
 	ManifestAnnotations []string
 
-	// Annotations contains parsed manifest and config annotations
+	// Annotations contains parsed annotations keyed by target:
+	// "$manifest" → manifest-level, or a filename → layer-level.
 	Annotations map[string]map[string]string
 }
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Annotation) ApplyFlags(fs *pflag.FlagSet) {
-	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil, "manifest annotations")
+	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil,
+		`manifest annotations (e.g. "key=value") or layer annotations (e.g. "filename:key=value")`)
 }
 
 // Parse parses the input annotation flags.
+//
+// Accepted formats:
+//   - "key=value"           → manifest-level annotation
+//   - "filename:key=value"  → layer-level annotation bound to filename
+//
+// A colon before the first equals sign is treated as the file/key separator.
+// This matches the annotation-file JSON format where filenames are top-level keys.
 func (opts *Annotation) Parse(*cobra.Command) error {
 	manifestAnnotations := make(map[string]string)
+	fileAnnotations := make(map[string]map[string]string)
+
 	for _, anno := range opts.ManifestAnnotations {
-		key, val, success := strings.Cut(anno, "=")
+		// Split on the first "=" to separate key-part from value.
+		keyPart, val, success := strings.Cut(anno, "=")
 		if !success {
 			return &oerrors.Error{
 				Err:            errAnnotationFormat,
-				Recommendation: `Please use the correct format in the flag: --annotation "key=value"`,
+				Recommendation: `Please use the correct format: --annotation "key=value" or --annotation "filename:key=value"`,
 			}
 		}
-		if _, ok := manifestAnnotations[key]; ok {
-			return fmt.Errorf("%w: %v, ", errAnnotationDuplication, key)
+
+		// A colon in keyPart (before the "=") signals a layer annotation.
+		if fileRef, key, isLayer := strings.Cut(keyPart, ":"); isLayer {
+			if key == "" {
+				return &oerrors.Error{
+					Err:            errAnnotationLayerKey,
+					Recommendation: fmt.Sprintf(`Provide an annotation key after the file prefix: --annotation "%s:key=value"`, fileRef),
+				}
+			}
+			if fileAnnotations[fileRef] == nil {
+				fileAnnotations[fileRef] = make(map[string]string)
+			}
+			if _, ok := fileAnnotations[fileRef][key]; ok {
+				return fmt.Errorf("%w: %v (layer: %v)", errAnnotationDuplication, key, fileRef)
+			}
+			fileAnnotations[fileRef][key] = val
+		} else {
+			// Manifest-level annotation.
+			key := keyPart
+			if _, ok := manifestAnnotations[key]; ok {
+				return fmt.Errorf("%w: %v, ", errAnnotationDuplication, key)
+			}
+			manifestAnnotations[key] = val
 		}
-		manifestAnnotations[key] = val
 	}
+
 	opts.Annotations = map[string]map[string]string{
 		AnnotationManifest: manifestAnnotations,
+	}
+	for fileRef, annos := range fileAnnotations {
+		opts.Annotations[fileRef] = annos
 	}
 	return nil
 }
--- a/cmd/oras/internal/option/annotation.go
+++ b/cmd/oras/internal/option/annotation.go
@@ -28,80 +28,88 @@ import (
 var (
 	errAnnotationFormat      = errors.New("annotation value doesn't match the required format")
 	errAnnotationDuplication = errors.New("duplicate annotation key")
-	errAnnotationLayerKey    = errors.New("annotation key must not be empty after the file prefix")
+	errAnnotationTarget      = errors.New("annotation target must not be empty before the colon")
+	errAnnotationLayerKey    = errors.New("annotation key must not be empty after the target")
 )
 
 // Annotation option struct.
 type Annotation struct {
 	// ManifestAnnotations contains raw input annotation flags.
 	// Two formats are accepted:
-	//   "key=value"               → manifest-level annotation
-	//   "filename:key=value"      → layer-level annotation bound to filename
+	//   "key=value"             → manifest-level annotation
+	//   "target:key=value"      → annotation scoped to an explicit target,
+	//                             which is a filename (layer), "$config", or
+	//                             "$manifest"
 	ManifestAnnotations []string
 
 	// Annotations contains parsed annotations keyed by target:
-	// "$manifest" → manifest-level, or a filename → layer-level.
+	// "$manifest" → manifest-level, "$config" → config-level, or a filename
+	// → layer-level.
 	Annotations map[string]map[string]string
 }
 
 // ApplyFlags applies flags to a command flag set.
 func (opts *Annotation) ApplyFlags(fs *pflag.FlagSet) {
 	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil,
-		`manifest annotations (e.g. "key=value") or layer annotations (e.g. "filename:key=value")`)
+		`annotation in the format of "key=value"; prefix a target to scope it, e.g. "filename:key=value" (layer), "$config:key=value" (config), or "$manifest:key=value" (manifest)`)
 }
 
 // Parse parses the input annotation flags.
 //
 // Accepted formats:
 //   - "key=value"           → manifest-level annotation
-//   - "filename:key=value"  → layer-level annotation bound to filename
+//   - "target:key=value"    → annotation scoped to an explicit target:
+//     a filename (layer), "$config", or "$manifest"
 //
-// A colon before the first equals sign is treated as the file/key separator.
-// This matches the annotation-file JSON format where filenames are top-level keys.
+// A colon before the first equals sign separates the target from the key.
+// This matches the annotation-file JSON format where targets ("$manifest",
+// "$config", or filenames) are top-level keys. OCI annotation keys use
+// reverse-DNS with dots (not colons), so the colon is safe as a separator.
 func (opts *Annotation) Parse(*cobra.Command) error {
-	manifestAnnotations := make(map[string]string)
-	fileAnnotations := make(map[string]map[string]string)
+	// The manifest target always has an entry so downstream packing can rely
+	// on its presence even when only layer/config annotations are set.
+	annotations := map[string]map[string]string{
+		AnnotationManifest: {},
+	}
 
 	for _, anno := range opts.ManifestAnnotations {
-		// Split on the first "=" to separate key-part from value.
-		keyPart, val, success := strings.Cut(anno, "=")
-		if !success {
+		// Split on the first "=" to separate the key-part from the value.
+		keyPart, value, ok := strings.Cut(anno, "=")
+		if !ok {
 			return &oerrors.Error{
 				Err:            errAnnotationFormat,
-				Recommendation: `Please use the correct format: --annotation "key=value" or --annotation "filename:key=value"`,
+				Recommendation: `Please use the correct format: --annotation "key=value" or --annotation "target:key=value"`,
 			}
 		}
 
-		// A colon in keyPart (before the "=") signals a layer annotation.
-		if fileRef, key, isLayer := strings.Cut(keyPart, ":"); isLayer {
-			if key == "" {
+		// A colon in the key-part (before the "=") scopes the annotation to
+		// an explicit target. Without a colon it applies to the manifest.
+		target, key := AnnotationManifest, keyPart
+		if ref, refKey, scoped := strings.Cut(keyPart, ":"); scoped {
+			if ref == "" {
 				return &oerrors.Error{
-					Err:            errAnnotationLayerKey,
-					Recommendation: fmt.Sprintf(`Provide an annotation key after the file prefix: --annotation "%s:key=value"`, fileRef),
+					Err:            errAnnotationTarget,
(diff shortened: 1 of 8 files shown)


TerryHowe (changes requested):
Seems reasonable to me although I'm curious what other maintainers think.

It will need some tests though unit and ideally an e2e

BRGOVIND:
Thanks for the review @TerryHowe! I've pushed unit tests for the layer-level annotation parsing (`cmd/oras/internal/option/annotation_test.go`) covering manifest-level, layer-level, mixed, multiple annotations per layer, empty input, values containing `=`, format errors, and duplicate-key cases.

Happy to add an e2e case as well — since you mentioned wanting other maintainers' input on the approach, I held off on the heavier e2e suite until the design direction is confirmed. Just let me know and I'll add it. Re-requesting review when you get a chance.

cmd/oras/internal/option/annotation.go
```diff
@@ -28,40 +28,80 @@ import (
...
-				Recommendation: `Please use the correct format in the flag: --annotation "key=value"`,
+				Recommendation: `Please use the correct format: --annotation "key=value" or --annotation "filename:key=value"`,
 			}
 		}
-		if _, ok := manifestAnnotations[key]; ok {
-			return fmt.Errorf("%w: %v, ", errAnnotationDuplication, key)
+
+		// A colon in keyPart (before the "=") signals a layer annotation.
+		if fileRef, key, isLayer := strings.Cut(keyPart, ":"); isLayer {
+			if key == "" {
```
TerryHowe: Can you add a check for fileRef=="" here and a test case?

cmd/oras/internal/option/annotation.go
```diff
@@ -28,40 +28,80 @@ import (
...
+			manifestAnnotations[key] = val
 		}
-		manifestAnnotations[key] = val
 	}
+
 	opts.Annotations = map[string]map[string]string{
 		AnnotationManifest: manifestAnnotations,
 	}
+	for fileRef, annos := range fileAnnotations {
+		opts.Annotations[fileRef] = annos
```
TerryHowe: Doesn't there need some check somewhere that fileRef is an file in this operation?

cmd/oras/internal/option/annotation_test.go
```diff
@@ -0,0 +1,147 @@
...
+*/
+
+package option
+
+import (
+	"errors"
+	"reflect"
+	"testing"
+)
+
```
TerryHowe, line 23: Can you include some tests for "$manifest" and "$config" mentioned in the issue?

TerryHowe (changes requested):
Good base here, as well as the other comments, can you update the push and attach help text for this? Thanks!

cmd/oras/internal/option/annotation.go
BRGOVIND: Done. An empty target before the colon (e.g. `":key=value"`) now returns a dedicated `errAnnotationTarget` with guidance to either use `"key=value"` for a manifest annotation or `"target:key=value"`. Covered by `TestAnnotation_Parse_errorEmptyTarget`.

cmd/oras/internal/option/annotation.go
BRGOVIND: Good catch — yes. I added `Packer.validateAnnotationTargets()` (in `packer.go`), which runs after parsing and errors with `errAnnotationNoMatchingFile` if a layer target does not match any file in the command. `$manifest` and `$config` are exempt.

I scoped this to the `--annotation` flag path only (it is mutually exclusive with `--annotation-file` via `errAnnotationConflict`), so existing `--annotation-file` behavior is untouched — erroring there could break users who currently rely on extra keys being ignored. Happy to broaden it to the annotation-file path too if you would prefer the stricter behavior everywhere. Covered by `TestPacker_parseAnnotations_layerTargetValidation`.

cmd/oras/internal/option/annotation_test.go
BRGOVIND: Added. The colon prefix now accepts the special `$config` and `$manifest` targets in addition to filenames, with tests: `TestAnnotation_Parse_configTarget`, `TestAnnotation_Parse_manifestTargetMergesWithBareKeys`, and `TestAnnotation_Parse_specialAndLayerTargets`. This also fixed a bug where `"$manifest:key=value"` overwrote bare `"key=value"` manifest annotations instead of merging; duplicate detection now applies across both forms (`TestAnnotation_Parse_errorDuplicateAcrossManifestForms`).

cmd/oras/root/push.go
```diff
@@ -103,6 +103,12 @@ Example - Push file to the HTTP registry:
 Example - Push repository with manifest annotations:
   oras push --annotation "key=val" localhost:5000/hello:v1
 
+Example - Push file "hi.txt" with a layer annotation scoped to that file:
+  oras push --annotation "hi.txt:key=val" localhost:5000/hello:v1 hi.txt
+
+Example - Push repository with a config annotation:
+  oras push --annotation "$config:key=val" localhost:5000/hello:v1
```
TerryHowe: Shell variable expansion would break this with double quotes
```suggestion
  oras push --annotation '$config:key=val' localhost:5000/hello:v1
```

cmd/oras/root/push.go
```diff
@@ -103,6 +103,12 @@ Example - Push file to the HTTP registry:
...
+
```
TerryHowe, line 114: I wonder about a $manifest example here just to bring awareness to it.

cmd/oras/internal/option/annotation.go
```diff
@@ -28,40 +28,88 @@ import (
...
-	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil, "manifest annotations")
+	fs.StringArrayVarP(&opts.ManifestAnnotations, "annotation", "a", nil,
+		`annotation in the format of "key=value"; prefix a target to scope it, e.g. "filename:key=value" (layer), "$config:key=value" (config), or "$manifest:key=value" (manifest)`)
 }
 
 // Parse parses the input annotation flags.
+//
+// Accepted formats:
+//   - "key=value"           → manifest-level annotation
+//   - "target:key=value"    → annotation scoped to an explicit target:
```
TerryHowe, line 61: This feature would technically be a breaking change for keys that had a : in them previously it should be noted.

cmd/oras/internal/option/annotation_test.go
```diff
@@ -0,0 +1,218 @@
...
```
TerryHowe, line 23: Needs e2e tests I think still. Ideally normal file, $config and $manifest for attach and push.

TerryHowe (changes requested):
Thanks, few more things

jmorla:
Hi @BRGOVIND,
If you're busy or stuck on any part of this, I'd be happy to help out or push a commit directly to get this ready for review. Just let me know how I can best support you!

BRGOVIND:
> Hi @BRGOVIND, If you're busy or stuck on any part of this, I'd be happy to help out or push a commit directly to get this ready for review. Just let me know how I can best support you!

that'd be great 
I'm in the middle of my exams 😭

jmorla:
@BRGOVIND I created this [Pull Request](https://github.com/BRGOVIND/oras/pull/2) which is pointing to your PR to resolve @TerryHowe's comments. please let me know what you think

BRGOVIND:
> @BRGOVIND I created this [Pull Request](https://github.com/BRGOVIND/oras/pull/2) which is pointing to your PR to resolve @TerryHowe's comments. please let me know what you think

Awesome, thanks Morla! I’ve merged the PR into my branch. appreciate you jumping in and handling the push updates and E2E coverage

cmd/oras/internal/option/annotation.go
TerryHowe: This breaking change here is still an issue

cmd/oras/internal/option/packer.go
```diff
@@ -109,7 +110,43 @@ func (opts *Packer) parseAnnotations(cmd *cobra.Command) error {
...
+// validateAnnotationTargets ensures every layer annotation set via the
+// `--annotation "filename:key=value"` flag refers to a file that is part of
+// this command. The special "$manifest" and "$config" targets are always
+// allowed. This surfaces typos as errors instead of silently dropping them,
+// since loadFiles only applies annotations to files it actually loads.
+//
+// It only runs for the `--annotation` flag path; the `--annotation-file` path
+// is mutually exclusive with it (see parseAnnotations) and is left untouched
+// to preserve backward compatibility.
+func (opts *Packer) validateAnnotationTargets() error {
```
TerryHowe, line 130: This validate method, doesn't it seem like it is a natural fit in annotation.go maybe called in Parse?

cmd/oras/internal/option/annotation.go
TerryHowe: Any innovative ways to get around the breaking nature of this change?

cmd/oras/internal/option/packer.go
BRGOVIND: Good point.
I considered moving this into annotation.go/Parse, but this validation is specifically concerned with whether the parsed target is valid for the packer’s current context, rather than parsing the annotation syntax itself. Keeping it here also leaves the existing ones behavior unchanged for backward compatibility. I’m happy to move it if you think that separation would be clearer.

cmd/oras/internal/option/annotation.go
jmorla: Two solutions come to mind to get around the breaking change:
• Introduce a dedicated flag like --layer-annotation.
• Add syntactic sugar like --annotation "@myfile.tar:key=value" (using @ or another prefix) so the user's intent to scope to a file is explicit and won't collide with existing manifest keys that contain colons.

let me know what do you think @TerryHowe

jmorla-bu: @TerryHowe we could either create a separated argument (maybe `layer-annotation`) or include a custom prefix '@filename:key=value'

cmd/oras/internal/option/annotation.go
```diff
@@ -28,40 +28,88 @@ import (
...
+	//   "target:key=value"      → annotation scoped to an explicit target,
+	//                             which is a filename (layer), "$config", or
+	//                             "$manifest"
 	ManifestAnnotations []string
 
-	// Annotations contains parsed manifest and config annotations
+	// Annotations contains parsed annotations keyed by target:
+	// "$manifest" → manifest-level, "$config" → config-level, or a filename
+	// → layer-level.
 	Annotations map[string]map[string]string
```
TerryHowe, line 48: **Critical:** `manifest index create` also embeds `option.Annotation` (create.go:55) but not `option.Packer`, so it gets this colon splitting with none of `validateAnnotationTargets`. It only reads `Annotations["$manifest"]`, so any other target is silently discarded:

```
# this branch
$ oras manifest index create --oci-layout idx:v1 child --annotation "com.example:key=val"
→ index has no annotations field at all

# main, same command
→ "annotations": {"com.example:key": "val"}
```

Silent data loss, no error. Suggest making the target syntax opt-in so only `Packer` consumers get it:

```suggestion
	ManifestAnnotations []string

	// AllowTargets enables the "target:key=value" form. Consumers that only
	// support manifest-level annotations (e.g. `manifest index create`) leave
	// this false so that a ":" in a key stays part of the key.
	AllowTargets bool

	// Annotations contains parsed annotations keyed by target:
	// "$manifest" → manifest-level, "$config" → config-level, or a filename
	// → layer-level.
	Annotations map[string]map[string]string
```

cmd/oras/internal/option/annotation.go
```diff
@@ -28,40 +28,88 @@ import (
...
 				Err:            errAnnotationFormat,
-				Recommendation: `Please use the correct format in the flag: --annotation "key=value"`,
+				Recommendation: `Please use the correct format: --annotation "key=value" or --annotation "target:key=value"`,
+			}
+		}
+
+		// A colon in the key-part (before the "=") scopes the annotation to
+		// an explicit target. Without a colon it applies to the manifest.
+		target, key := AnnotationManifest, keyPart
+		if ref, refKey, scoped := strings.Cut(keyPart, ":"); scoped {
```
TerryHowe, line 88: Gate the split on the opt-in flag (see the struct comment):

```suggestion
		if ref, refKey, scoped := strings.Cut(keyPart, ":"); scoped && opts.AllowTargets {
```

Separately, `Cut` takes the *first* colon while `fileref.Parse` deliberately special-cases the Windows drive prefix (`fileref/windows.go` `doParse`). So `--annotation "C:\dir\file.txt:key=val"` parses target `C` and fails validation, with no way to express it. Matching `keyPart` against the known `FileRefs` set (longest match) instead of a blind first-colon cut would handle both this and the case where someone mirrors the file-ref syntax and writes `hi.txt:application/vnd.foo:key=val`.

cmd/oras/internal/option/packer.go
```diff
@@ -109,7 +110,43 @@ func (opts *Packer) parseAnnotations(cmd *cobra.Command) error {
 		}
 	}
 	if len(opts.ManifestAnnotations) != 0 {
-		return opts.Annotation.Parse(cmd)
+		if err := opts.Annotation.Parse(cmd); err != nil {
+			return err
+		}
+		return opts.validateAnnotationTargets()
+	}
```
TerryHowe, line 117:
```suggestion
	if len(opts.ManifestAnnotations) != 0 {
		opts.Annotation.AllowTargets = true
		if err := opts.Annotation.Parse(cmd); err != nil {
			return err
		}
		return opts.validateAnnotationTargets()
	}
```

Also: `validateAnnotationTargets` ranges over `opts.Annotations`, which is a map — with two bad targets the reported one is random. Iterating `opts.ManifestAnnotations` in flag order would make the error reproducible.

cmd/oras/internal/option/annotation.go
```diff
@@ -28,40 +28,88 @@ import (
...
+//
+// Accepted formats:
+//   - "key=value"           → manifest-level annotation
+//   - "target:key=value"    → annotation scoped to an explicit target:
+//     a filename (layer), "$config", or "$manifest"
+//
+// A colon before the first equals sign separates the target from the key.
+// This matches the annotation-file JSON format where targets ("$manifest",
+// "$config", or filenames) are top-level keys. OCI annotation keys use
+// reverse-DNS with dots (not colons), so the colon is safe as a separator.
```
TerryHowe, line 67: The reverse-DNS convention isn't a spec rule — the image-spec places no character restriction on annotation keys, so `:` is only *discouraged*, not invalid. This is a real break for anyone using one today:

```
$ oras push --annotation "com.example:key=val" localhost:5000/hello:v1
Error: annotation target does not match any file in this command: "com.example"
```

At least it's loud on push/attach (unlike index create). But it needs a release note calling out the break and the `$manifest:` escape hatch, and I'd like another maintainer to weigh in on whether that's acceptable or whether the syntax should be opt-in behind a flag. Worth softening this comment to say convention rather than fact.

TerryHowe (changes requested):
Thanks for pushing this forward, and thanks @jmorla for jumping in on the help text and e2e.

I built the branch and exercised it against `main`. The feature itself works — verified layer/config/manifest annotations all land correctly via `--oci-layout`. Two things need fixing before merge, inline below.

The short version: `manifest index create` embeds `option.Annotation` directly (not `option.Packer`), so it gets the new colon splitting but none of the target validation — annotations with a `:` in the key are now silently dropped there. And on `push`/`attach`, a manifest key containing `:` that worked on `main` is now an error. `:` in an annotation key is discouraged by convention, but nothing in the image-spec forbids it, so this is a behavior break we should be deliberate about.

Local runs: `go build ./...`, `go vet ./cmd/...` and `go test ./cmd/oras/internal/option/...` all pass. Note CI hasn't run past DCO on this PR, so the new tests are unverified upstream.

cmd/oras/root/attach.go
TerryHowe, line 1: **Important**, on the guard at attach.go:168: it checks `len(opts.Annotations[option.AnnotationManifest]) == 0`, but `Parse` now always seeds a `$manifest` entry that is often empty. So a config-only attach is rejected with a message that contradicts what the user typed:

```
$ oras attach --artifact-type doc/e --annotation '$config:k=v' localhost:5000/hello:v1
```

The guard should consider the whole `Annotations` map, not just the manifest bucket.

(Commenting at file level since line 168 isn't in the diff.)

Issue #2111, opened 2026-07-13
checksum signature file is missing from release 1.3.3


bb-Ricardo:
Hi,

I noticed that the file `oras_1.3.3_checksums.txt.asc` is missing from the release assets of [1.3.3](https://github.com/oras-project/oras/releases/tag/v1.3.3). It was still present in 1.3.2:  [oras_1.3.2_checksums.txt.asc](https://github.com/oras-project/oras/releases/download/v1.3.2/oras_1.3.2_checksums.txt.asc)

I had a look at the version diff and the only think which I could see were the updated and pinned versions in the release workflow file: https://github.com/oras-project/oras/compare/v1.3.2...v1.3.3

Would it be possible to add this file additionally to the release? currently our validation pipeline is failing as it is not possible to verify the checksum file vie GPG. I know there is this FR https://github.com/oras-project/oras/issues/2006 but I assume this is not implemented at the moment.

TerryHowe:
The `oras_1.3.3_checksums.txt.asc` signature file has been added to the [v1.3.3 release](https://github.com/oras-project/oras/releases/tag/v1.3.3). It's signed with the same GPG key used for v1.3.2, so your validation pipeline should verify it as before:

```
curl -sSLO https://github.com/oras-project/oras/releases/download/v1.3.3/oras_1.3.3_checksums.txt
curl -sSLO https://github.com/oras-project/oras/releases/download/v1.3.3/oras_1.3.3_checksums.txt.asc
gpg --verify oras_1.3.3_checksums.txt.asc oras_1.3.3_checksums.txt
```

The signature file is created as a manual step during release, which is why it was missed this time. Thanks for the report! Longer term, #2006 tracks automating artifact signing so this doesn't get dropped again.

TerryHowe:
Should be fixed, reopen if there is a problem. Created PR so it doesn't have to be a manual process.

Issue #2006, opened 2026-03-12
feat: adopt GitHub attestations for release artifact signing


TerryHowe:
### Overview

This issue tracks adopting a modern signing mechanism for ORAS release artifacts, such as [Notary Project (notation)](https://notaryproject.dev/), [sigstore/cosign](https://github.com/sigstore/cosign) attestations, or [GitHub artifact attestations](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations).

This was raised as a side note in #2005, where a user noted that GPG-based `.asc` signature files were missing from the release. The user also pointed out that a modern signing alternative would be acceptable, as long as binary integrity can be verified in a pipeline.

### Motivation

- GitHub attestations are natively integrated with GitHub Actions and require no external key management.
- [Notary Project](https://notaryproject.dev/) (`notation`) is a CNCF project purpose-built for signing software supply chain artifacts and aligns naturally with the OCI/ORAS ecosystem.
- Sigstore/cosign provides keyless signing via OIDC, improving the security posture compared to long-lived GPG keys.
- Any of these approaches would allow users to verify release artifacts programmatically, e.g.:
  ```
  # GitHub attestations
  gh attestation verify oras_linux_amd64.tar.gz --repo oras-project/oras

  # cosign
  cosign verify-blob --certificate-identity ... oras_linux_amd64.tar.gz
  ```

### Proposed approach

- **Option A (preferred):** Integrate `actions/attest-build-provenance` (GitHub attestations) into the release workflow to generate SLSA provenance attestations for all release artifacts.
- **Option B:** Use [notation](https://github.com/notaryproject/notation) to sign release artifacts with a Notary-compatible signature, aligning with the broader CNCF supply chain security ecosystem.
- **Option C:** Use `sigstore/cosign-installer` + `cosign sign-blob` for keyless signing.
- Document the verification steps in the release notes and/or README.

### References

- #2005 — Signature files missing from release artifacts (side note)
- [Notary Project](https://notaryproject.dev/) / [notation CLI](https://github.com/notaryproject/notation)
- [sigstore/cosign](https://github.com/sigstore/cosign)
- [GitHub artifact attestations docs](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations)
- [actions/attest-build-provenance](https://github.com/actions/attest-build-provenance)

TerryHowe:
[...]

PR #2112, opened 2026-07-13
feat: sign release checksums in the release workflow

TerryHowe:
## What

Sign the release `checksums.txt` automatically in the release workflow so the `.asc` signature file is always published.

## Why

The `oras_<version>_checksums.txt.asc` signature has been produced by a manual, out-of-band maintainer step each release. That step was missed for v1.3.3, so the signature file was absent from the release assets and consumers' GPG-based verification pipelines broke (#2111). The `.asc` for v1.3.3 has since been added by hand; this PR prevents a recurrence by automating it.

## Changes

- **`.goreleaser.yml`** — enable a scoped `signs` block (`artifacts: checksum`) that produces `${artifact}.asc` using a dedicated **ORAS Project Release Signing** key via loopback pinentry.
- **`.github/workflows/release-github.yml`** — import the signing key from the `GPG_PRIVATE_KEY` secret and pass `GPG_PASSPHRASE` to goreleaser.
- **`KEYS`** — add the new project signing public key (`C6FC42F0DE84A345FAADB742F44418110791EF37`) and document how to verify a release.

## New signing key

This introduces a **dedicated project signing key** rather than continuing to sign with an individual maintainer's personal key. The new public key is added to `KEYS`. Both the old and new keys are listed, so this does not invalidate signatures on prior releases — but anyone pinning a specific maintainer key for verification should add the new one.

## Required before this can be used

Two repository secrets must be configured on `oras-project/oras` before the next tagged release:

- `GPG_PRIVATE_KEY` — ASCII-armored private key for the signing key above
- `GPG_PASSPHRASE` — its passphrase

## Verifying a release (after this lands)

```
gpg --import KEYS
gpg --verify oras_<version>_checksums.txt.asc oras_<version>_checksums.txt
```

Relates to #2111, #2006.

### DIFF
--- a/.github/workflows/release-github.yml
+++ b/.github/workflows/release-github.yml
@@ -34,6 +34,11 @@ jobs:
         uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
         with:
           syft-version: v1.32.0
+      - name: Import GPG signing key
+        run: |
+          echo "${GPG_PRIVATE_KEY}" | gpg --batch --import
+        env:
+          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
       - name: run goreleaser
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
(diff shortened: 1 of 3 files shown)


sabre1041 (approved):
LGTM

PR #2114, opened 2026-07-18
Fall back to config media type for old-style artifact types

### LINKED ISSUE #2015: Specification Deviation in Multi Arch Artifacts

A user was mentioning that we may not be properly following the specification for multi arch artifacts. Specifically around https://github.com/opencontainers/image-spec/blob/main/artifacts-guidance.md

I don't completely understand all the details, but wanted to record that though for investigation.

The spec to be followed

### How can we reproduce it?

Create a multiarch artifact

1.3.0

Linux

Kunalbehbud:

Addresses Issue 1 of #2015 (the `config.mediaType` fallback for descriptor `artifactType`), following @TerryHowe's assessment on the issue.

When `oras manifest index create`/`update` enriches a child image manifest descriptor, `enrichDescriptor` set `desc.ArtifactType = manifest.ArtifactType` and left it empty for **old-style artifacts** — those created before the manifest `artifactType` field existed, which convey their type through `config.mediaType` instead (e.g. historically-created artifacts like OpenTofu providers).

The image-spec [descriptor guidance](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties) says `artifactType` "is the value of the config descriptor `mediaType` when the descriptor references an image manifest", and the [manifest guidelines](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage) explicitly call for tooling to "fallback to the `config.mediaType` value". This change does that: when `artifactType` is empty and the config is not a standard OCI image config, the enriched descriptor falls back to `config.mediaType`.

`enrichDescriptor` is shared by both the index `create` and `update` paths, so a single change fixes both (the issue assessment expected two edits).

This PR intentionally scopes to Issue 1 only. Issue 2 (a mechanism/flag to attach platform to artifact descriptors) is a larger, separate change and is left for a follow-up.

Fixes #2015

- [x]  Does the affected code have corresponding tests, e.g. unit test, E2E test? — Added two `Test_enrichDescriptor` cases: the fallback path (empty `artifactType` + custom `config.mediaType`), and the guard that keeps `artifactType` empty when the config is the standard OCI image config type. This closes the "no test for the old-style fallback path" gap noted in the issue.
- [ ]  Does this change require a documentation update? — No; behavior now matches the documented spec expectation.
- [ ]  Does this introduce breaking changes that would require an announcement or bumping the major version? — No; it only populates a descriptor field that was previously left empty for these artifacts.
- [x]  Do all new files have an appropriate license header? — No new files.

### DIFF
--- a/cmd/oras/root/manifest/index/create.go
+++ b/cmd/oras/root/manifest/index/create.go
@@ -241,6 +241,13 @@ func enrichDescriptor(ctx context.Context, target oras.ReadOnlyTarget, desc ocis
 			return ocispec.Descriptor{}, err
 		}
 		desc.ArtifactType = manifest.ArtifactType
+		if desc.ArtifactType == "" && manifest.Config.MediaType != ocispec.MediaTypeImageConfig {
+			// Old-style artifacts predate the manifest artifactType field and
+			// convey their type through config.mediaType instead. Per the
+			// image-spec descriptor guidance, fall back to config.mediaType so
+			// the enriched descriptor still advertises an artifactType.
+			desc.ArtifactType = manifest.Config.MediaType
+		}
 	} else if descriptor.IsIndex(desc) {
 		var index ocispec.Index
 		if err := json.Unmarshal(manifestBytes, &index); err != nil {
--- a/cmd/oras/root/manifest/index/create.go
+++ b/cmd/oras/root/manifest/index/create.go
@@ -38,6 +38,7 @@ import (
 	"oras.land/oras/cmd/oras/internal/option"
 	"oras.land/oras/internal/contentutil"
 	"oras.land/oras/internal/descriptor"
+	"oras.land/oras/internal/docker"
 	"oras.land/oras/internal/listener"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/oras/root/manifest/index/create.go
```diff
@@ -241,6 +241,13 @@ func enrichDescriptor(ctx context.Context, target oras.ReadOnlyTarget, desc ocis
 			return ocispec.Descriptor{}, err
 		}
 		desc.ArtifactType = manifest.ArtifactType
+		if desc.ArtifactType == "" && manifest.Config.MediaType != ocispec.MediaTypeImageConfig {
+			// Old-style artifacts predate the manifest artifactType field and
+			// convey their type through config.mediaType instead. Per the
+			// image-spec descriptor guidance, fall back to config.mediaType so
+			// the enriched descriptor still advertises an artifactType.
+			desc.ArtifactType = manifest.Config.MediaType
+		}
```
TerryHowe: Not sure on whitespace, but I think this needs to be something like
```suggestion
		if desc.ArtifactType == "" {
    		switch manifest.Config.MediaType {
    		case ocispec.MediaTypeImageConfig, ocispec.MediaTypeEmptyJSON, docker.MediaTypeConfig:
        		// standard image / empty config: not an artifact type, leave empty
    		default:
        		desc.ArtifactType = manifest.Config.MediaType
    		}
		}
```

TerryHowe (changes requested):

Kunalbehbud:
Good catch — applied. Switched to the switch-based guard so the config media type is only used when it isn't a standard image config: `ocispec.MediaTypeImageConfig`, `ocispec.MediaTypeEmptyJSON`, and the Docker image config are all excluded.

`docker.MediaTypeConfig` didn't exist yet, so I added it to `internal/docker` alongside the existing manifest media types (`application/vnd.docker.container.image.v1+json`).

Also extended `Test_enrichDescriptor` to cover all three excluded config types staying empty, on top of the fallback case. PTAL.

TerryHowe (approved):
/lgtm

TerryHowe:
In test/e2e/internal/testdata/nonjson_config/const.go:
```
var (
      Descriptor = ocispec.Descriptor{
              MediaType:    "application/vnd.oci.image.manifest.v1+json",
              Digest:       "sha256:9d16f5505246424aed7116cb21216704ba8c919997d0f1f37e154c11d509e1d2",
              Size:         529,
              ArtifactType: "application/vnd.unknown.config.v1+json", // ← add this
      }
)
```

TerryHowe (changes requested):
Test is broken

PR #2126, opened 2026-08-05
fix: avoid send on closed channel in progress messenger

### LINKED ISSUE #2125: panic: send on closed channel in progress messenger during oras copy

`oras copy` can crash with `panic: send on closed channel` in the progress/display code. Reported in #2101 (see [this comment](https://github.com/oras-project/oras/issues/2101#issuecomment-5181808138)) while running a plain `oras copy docker.io/library/nginx:1.31.2 localhost:5000/nginx:1.31.2`.

```text
panic: send on closed channel

goroutine 69 [running]:
oras.land/oras/cmd/oras/internal/display/status/progress.(*messenger).Fail(...)
	oras.land/oras/cmd/oras/internal/display/status/progress/messenger.go:46 +0x74
oras.land/oras/internal/progress.(*readTracker).Read(...)
	oras.land/oras/internal/progress/tracker.go:105 +0xf3
io.discard.ReadFrom(...)
io.copyBuffer(...)
net/http.(*transferWriter).doBodyCopy(...)
net/http.(*transferWriter).writeBody(...)
net/http.(*Request).write(...)
net/http.(*persistConn).writeLoop(...)
	net/http/transport.go:2600 +0x174
created by net/http.(*Transport).dialConn
```

### Root cause

Use-after-close race on the progress channel:

- Each tracked blob wraps its reader via `progress.TrackReader` (`internal/progress/tracker.go`). Reads drive `readTracker.Read`, which on a non-EOF error calls `tracker.Fail(err)` (`internal/progress/tracker.go:107-111`).
- `tracker.Fail` → `messenger.Fail` does an **unguarded send**: `m.update <- updateStatusError(err)` (`cmd/oras/internal/display/status/progress/messenger.go:46`). `messenger.Update` (lines 31/39) has the same exposure.
- The channel is closed by `messenger.Close()` (`messenger.go:56`), invoked from `Reader.StopTracker()` (`track/reader.go:69`) via `defer r.StopTracker()` in `referenceGraphTarget.PushReference`/`Push` (`track/target.go:102`) — this fires as soon as the push call returns.
- **The race:** `net/http`'s `persistConn.writeLoop` goroutine keeps reading the request-body reader (the tracked reader) *after* the round-trip returned and `Close()` already closed the channel. When that late read hits `EOF`/error, `Fail` sends on the closed channel → panic. The stack trace confirms exactly this path (`writeLoop → doBodyCopy → readTracker.Read → messenger.Fail`).

The `messenger` struct already has a `closed bool` field (`messenger.go:23`), but it only guards `Close()` itself — not `Update`/`Fail`, and it is not synchronized.

### Affected code

- `cmd/oras/internal/display/status/progress/messenger.go:44-48` — `Fail`: the unguarded `m.update <-` that panics. **Primary fix site.**
- `cmd/oras/internal/display/status/progress/messenger.go:28-42` — `Update`: same exposure (blocking sends at lines 31, 39).
- `cmd/oras/internal/display/status/progress/messenger.go:50-59` — `Close`: closes the channel; `closed` flag is unsynchronized.
- `internal/progress/tracker.go:96-113` — `readTracker.Read`: calls `tracker.Fail(err)` from the `net/http` writeLoop goroutine — the trigger.
- `internal/progress/tracker.go:150-168` — `writeTracker.Write`: analogous `Fail` path; a fix should cover it too.
- `cmd/oras/internal/display/status/track/reader.go:67-76` — `StopTracker`/`StopManager`: where `tracker.Close()` ends the channel's lifetime.
- `cmd/oras/internal/display/status/track/target.go:97-111` — `PushReference` with `defer r.StopTracker()`.
- `cmd/oras/internal/display/status/progress/manager.go:125-136` — `newTracker`: single drain goroutine + `bufferSize=1`; the concurrency context a fix must respect.

### Suggested approach

1. **Make late sends no-ops instead of panics.** Add a `done chan struct{}` to `messenger`; `Close` closes `done`. In `Update` and `Fail`, wrap every send in:
   ```go
   select {
   case m.update <- upd:
   case <-m.done:
       // tracker already closed; drop the update
   }
   ```
   This avoids both the panic and a lock-across-blocking-send deadlock. (The `StateTransmitting` case already uses a non-blocking `select`; extend the same discipline to the blocking cases and to `Fail`.)
2. Ensure `Close` ordering/idempotency so a concurrent `Fail`/`Update` cannot send after `close(m.update)` — close `done` first, then drain.
3. Add a `-race` regression test in `messenger_test.go`: hammer `Fail`/`Update` from a goroutine while the main goroutine calls `Close`, assert no panic.

### Notes / risk

- All internal packages — **no public API impact**, no backward-compat concerns.
- Real risk is concurrency correctness: `bufferSize = 1` with a single drain goroutine (`manager.go`), so a naive mutex held across a blocking send can deadlock against `Close`. Prefer the `done`-channel `select` over a mutex.
- Coverage gap: the progress package has tests but none exercise the post-close send race.

### Context

Split out from #2101. That issue also reported a separate hang (`oras copy --from-oci-layout` stalling after 100% with an `EOF` on the blob `PUT`), which appears to be a registry/transport-timeout or `oras-go` retry behavior rather than a CLI bug. This issue tracks only the `send on closed channel` panic (part B of the analysis in #2101).

Reproduced against 1.3.2 / `main`. Panic path present in current `main`.

harshasiddartha:

`oras copy` can panic with `send on closed channel` in the progress display.

The race, precisely:

- `messenger.Close()` did `m.update <- updateStatusEndTime(); close(m.update)`. It is called from `Reader.StopTracker()`, which runs via `defer r.StopTracker()` in `referenceGraphTarget.PushReference` / `Push` — i.e. as soon as the push call returns.
- The tracked reader handed to the push is the HTTP request body. `net/http`'s `persistConn.writeLoop` goroutine keeps reading it after the round trip has returned. When one of those late reads returns a non-EOF error, `readTracker.Read` calls `tracker.Fail(err)`, which was an unguarded `m.update <- updateStatusError(err)`.
- So the closing goroutine (the caller of `PushReference`) and the sending goroutine (`net/http` `writeLoop`) are unsynchronized, and a send can land after the close. That is exactly the reported stack: `writeLoop → doBodyCopy → readTracker.Read → messenger.Fail`. `writeTracker.Write` has the same `Fail` path.

The existing `closed bool` only short-circuited a second `Close()`; it was unsynchronized and did not guard `Update`/`Fail`.

The fix: the messenger no longer closes the update channel at all. It owns a `done` channel that `Close` closes (once, via `sync.Once`), every blocking send selects on `done`, and the drain goroutine in `manager.newTracker` now stops on `done` after applying whatever is still buffered. A late `Update`/`Fail` therefore takes the `done` branch and drops the update instead of panicking. `Close` sends the end-time update *before* closing `done`, so the final status is never lost.

Note on the approach: adding a `select` on `done` while still closing the update channel would not have fixed it. `select` commits to a ready send case and then performs the send, so a concurrent `close` in that window still panics. Not closing the channel is what actually serializes the two goroutines — the only cross-goroutine transition left is `close(done)`, and a closed `done` is permanently ready, so senders can neither panic nor block. A mutex held across the blocking send was rejected as it couples sender liveness to the drain goroutine.

Reproduced before the fix with `go test -race` (drain goroutine + a goroutine calling `Fail` while `Close` runs):

```
==================
WARNING: DATA RACE
Read at 0x00c00013b6d0 by goroutine 9:
  runtime.chansend1()
  oras.land/oras/cmd/oras/internal/display/status/progress.(*messenger).Fail()
      cmd/oras/internal/display/status/progress/messenger.go:46 +0xa4
...
panic: send on closed channel
```

`Test_messenger_Close_concurrentUpdate` is the regression test for it; it fails if the channel close is put back.

Verified: `go build ./...`, `go vet ./...`, `go test -race ./...` (all packages pass), `golangci-lint run ./...` (0 issues).

Fixes #2125

- [x]  Does this change require a documentation update? — no, all internal packages, no user-facing or API change
- [x]  Does this introduce breaking changes that would require an announcement or bumping the major version? — no
- [x]  Do all new files have an appropriate license header? — no new files

### DIFF
--- a/cmd/oras/internal/display/status/progress/manager.go
+++ b/cmd/oras/internal/display/status/progress/manager.go
@@ -123,16 +123,11 @@ func (m *manager) Track(desc ocispec.Descriptor) (progress.Tracker, error) {
 }
 
 func (m *manager) newTracker(s *status) progress.Tracker {
-	ch := make(chan statusUpdate, bufferSize)
+	msg := newMessenger(m.prompts)
 	m.updating.Go(func() {
-		for update := range ch {
-			update(s)
-		}
+		msg.drain(s)
 	})
-	return &messenger{
-		update:  ch,
-		prompts: m.prompts,
-	}
+	return msg
 }
 
 // Close stops all status and waits for updating and rendering.
--- a/cmd/oras/internal/display/status/progress/messenger.go
+++ b/cmd/oras/internal/display/status/progress/messenger.go
@@ -15,45 +15,96 @@ limitations under the License.
 
 package progress
 
-import "oras.land/oras/internal/progress"
+import (
+	"sync"
+
+	"oras.land/oras/internal/progress"
+)
 
 // messenger is progress message channel.
+//
+// The update channel is deliberately never closed. Trackers are handed to
+// io.Reader / io.Writer wrappers that can outlive the operation they track,
+// most notably the net/http request body reader, which keeps being read by
+// the transport write loop after the round trip returns. Closing the channel
+// would let such a late Update or Fail send on a closed channel and panic.
+// Instead, Close closes the done channel and every send races the done signal,
+// so late updates are dropped instead of panicking.
 type messenger struct {
-	update  chan statusUpdate
-	closed  bool
-	prompts map[progress.State]string
+	update    chan statusUpdate
+	done      chan struct{}
+	closeOnce sync.Once
+	prompts   map[progress.State]string
+}
+
+// newMessenger creates a new messenger.
+func newMessenger(prompts map[progress.State]string) *messenger {
+	return &messenger{
+		update:  make(chan statusUpdate, bufferSize),
+		done:    make(chan struct{}),
+		prompts: prompts,
+	}
+}
+
+// send delivers the update unless the messenger is already closed.
+func (m *messenger) send(update statusUpdate) {
+	select {
+	case m.update <- update:
+	case <-m.done:
+		// the messenger is closed, drop the update
+	}
 }
 
 // Update sends the status to the message channel.
 func (m *messenger) Update(status progress.Status) error {
 	switch status.State {
 	case progress.StateInitialized:
-		m.update <- updateStatusStartTime()
+		m.send(updateStatusStartTime())
 	case progress.StateTransmitting:
 		select {
 		case m.update <- updateStatusMessage(m.prompts[progress.StateTransmitting], status.Offset):
 		default:
-			// drop message if channel is full
+			// drop message if channel is full or the messenger is closed
 		}
 	default:
-		m.update <- updateStatusMessage(m.prompts[status.State], status.Offset)
+		m.send(updateStatusMessage(m.prompts[status.State], status.Offset))
 	}
 	return nil
 }
 
 // Fail sends the error to the message channel.
 func (m *messenger) Fail(err error) error {
-	m.update <- updateStatusError(err)
+	m.send(updateStatusError(err))
 	return nil
 }
 
-// Close marks the progress as completed and closes the message channel.
+// Close marks the progress as completed and stops the message channel.
 func (m *messenger) Close() error {
-	if m.closed {
-		return nil
-	}
-	m.update <- updateStatusEndTime()
-	close(m.update)
-	m.closed = true
+	m.closeOnce.Do(func() {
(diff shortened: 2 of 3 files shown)


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #2137, opened 2026-08-23
release-github.yml: float Go version to match other workflows

### LINKED ISSUE #2132: release-github.yml pins go-version 1.25.7, so release binaries miss stdlib security patches
## Summary

`.github/workflows/release-github.yml` pins an exact Go patch version:

```yaml
      - name: setup go environment
        uses: actions/setup-go@... # v7.0.0
        with:
          go-version: '1.25.7'
```

Every other workflow in the repo uses the floating minor `'1.25'`, which
`actions/setup-go` resolves to the latest available patch at run time:

| Workflow | `go-version` | Tracks patches? |
| --- | --- | --- |
| `build.yml` | `['1.25']` | yes |
| `codeql-analysis.yml` | `['1.25']` | yes |
| `golangci-lint.yml` | `['1.25']` | yes |
| `vulnerability-check.yml` | `'1.25'` | yes |
| `snapcraft.yaml` | `go/1.25/stable` | yes |
| **`release-github.yml`** | **`'1.25.7'`** | **no** |

Because `release-github.yml` is the workflow that builds and GPG-signs the
binaries attached to a GitHub release, the artifacts users actually download are
the one thing in the repo built on a frozen, silently ageing toolchain.

## Impact

Go 1.25.7 is currently 6 patch releases behind, and the gap includes at least
one stdlib security fix. As a concrete example, [GO-2026-5026] /
[CVE-2026-39821] (`golang.org/x/net/idna` accepting Punycode labels that decode
to an ASCII-only label, vendored into `net/http`) is fixed in Go **1.25.13**,
1.26.6, and 1.27.0-rc.3. Binaries built with 1.25.7 still carry the affected
stdlib, so downstream SCA scanners flag the published release artifacts.

Two things make this easy to miss:

- **Nothing bumps the pin automatically.** Dependabot's `gomod` and
  `github-actions` ecosystems update module requirements and action refs; neither
  rewrites a `go-version:` string literal. The pin only moves when someone edits
  it by hand, and it has not been touched in the recent history of the file.
- **`vulnerability-check.yml` will not catch it.** That workflow runs
  `govulncheck` under floating `'1.25'`, so it analyses against an up-to-date
  toolchain and reports clean even while the release job is still on 1.25.7.

Related: #2131 bumps the `Dockerfile` builder stage to `golang:1.26.6-alpine` for
this same CVE. That fixes the `ghcr.io/oras-project/oras` image only — the
release binaries are a separate build path and remain on 1.25.7.

## Proposed fix

In `.github/workflows/release-github.yml`, change the pin to the floating minor
so it matches the rest of the repo and picks up patch releases automatically:

```diff
-          go-version: '1.25.7'
+          go-version: '1.25'
```

If maintainers prefer an exact pin here for release reproducibility, the
alternative is bumping it to `'1.25.13'` (or newer) now and treating the pin as
something that needs periodic manual review — but the floating minor is
consistent with the other five build paths and needs no ongoing upkeep.

Worth confirming as part of the change: `go.mod` declares `go 1.25.7`, which is a
*minimum* language version rather than a toolchain pin, so it does not hold the
build back — any toolchain ≥ 1.25.7 satisfies it. No `toolchain` directive is
present.

## Verification

After the change, a release build should report a Go version ≥ 1.25.13 in the
`setup go environment` step log, and `govulncheck` against the produced binary
should no longer report GO-2026-5026.

[GO-2026-5026]: https://pkg.go.dev/vuln/GO-2026-5026
[CVE-2026-39821]: https://nvd.nist.gov/vuln/detail/CVE-2026-39821

sooraj-sky:
## Summary
- `.github/workflows/release-github.yml` pinned an exact Go patch version (`1.25.7`) for the job that builds and GPG-signs the release binaries.
- Every other workflow in the repo (`build.yml`, `codeql-analysis.yml`, `golangci-lint.yml`, `vulnerability-check.yml`, `snapcraft.yaml`) already uses the floating minor `'1.25'`, which `actions/setup-go` resolves to the latest patch at run time.
- The pin was 6 patches stale and missed a stdlib security fix (GO-2026-5026 / CVE-2026-39821, fixed in 1.25.13), and nothing bumps it automatically since Dependabot's `gomod`/`github-actions` ecosystems don't rewrite `go-version:` string literals.
- Fix: switch `release-github.yml` to the same floating minor used everywhere else in the repo, so release binaries pick up patch releases automatically like every other build path.

Fixes #2132

## Test plan
- [x] Confirmed the old exact pin (`1.25.7`) is gone and matches the floating-minor pattern (`'1.25'`) used by the other 5 workflows in the same repo
- Workflow itself can only be exercised on an actual release run (`workflow_dispatch`/tag push) in CI on merge — the change is a one-line value swap of an already-used, valid `setup-go` input

### DIFF
--- a/.github/workflows/release-github.yml
+++ b/.github/workflows/release-github.yml
@@ -29,7 +29,7 @@ jobs:
       - name: setup go environment
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.25.7'
+          go-version: '1.25'
       - name: Install Syft
         uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
         with:


TerryHowe (approved):
/lgtm

PR #2155, opened 2026-09-05
fix(ci): resolve licenserc go.mod path relative to config directory

yunaremaia:
## Summary

The `dependency.files` path in `.github/licenserc.yml` is resolved relative to the config file's directory (`.github/`), not the repo root. The previous value `go.mod` pointed to `.github/go.mod` which does not exist.

skywalking-eyes 0.9.0 added a `validateGoModFile()` check that opens the declared path before chdir, surfacing this pre-existing misconfiguration. Every other check on the skywalking-eyes bump PR #2149 passes — only `check-license` fails.

## The Fix

```diff
 dependency:
   files:
-    - go.mod
+    - ../go.mod
```

`filepath.Join(".github", "../go.mod")` cleans to `go.mod` at the repo root, which is the module we actually want scanned — the same module 0.8.0 was reaching accidentally via `go mod download` walking up from `.github/`. So the set of dependencies checked does not change.

## Why this is safe under 0.8.0

The fix is also correct under the currently pinned 0.8.0 (it resolves to the same root `go.mod` that version reaches by walking up), so it can land on its own first. Once on `main`, #2149 needs a rebase for its `check-license` run to go green.

## Prior art

Identical bug and fix in oras-go: oras-project/oras-go#1359 → oras-project/oras-go#1366. This repo never received the port.

Fixes #2154

PR #2158, opened 2026-09-06
fix(ci): resolve dependency license config path from repo root

### LINKED ISSUE #2154: check-license fails after skywalking-eyes 0.9.0: dependency.files path is resolved relative to .github/
### What happened

The `check-license` job fails on [#2149](https://github.com/oras-project/oras/pull/2149) (Dependabot bumping the `github-actions` group, which includes `apache/skywalking-eyes` 0.8.0 -> 0.9.0):

```
INFO  Loading configuration from file: .github/licenserc.yml
ERROR open /home/runner/work/oras/oras/.github/go.mod: no such file or directory
ERROR one or more errors occurred checking license compatibility
```

[Failing job](https://github.com/oras-project/oras/actions/runs/33467392152/job/99729983426). Every other check on that PR is green.

### Root cause

`.github/licenserc.yml` declares:

```yaml
dependency:
  files:
    - go.mod
```

`ConfigDeps.Finalize()` resolves relative `files` entries against **the config file's directory**, not the repo root — [`pkg/deps/config.go#L61-L65`](https://github.com/apache/skywalking-eyes/blob/v0.9.0/pkg/deps/config.go#L61-L65), and this code is byte-identical in 0.8.0 and 0.9.0. So `go.mod` has always meant `.github/go.mod`, which does not exist.

0.8.0 never noticed. Its `GoModResolver.Resolve` only did `os.Chdir(filepath.Dir(goModFile))`; chdir into `.github/` succeeds, and the subsequent `go mod download` walks *up* and finds the real root `go.mod`:

```console
$ cd oras/.github && go env GOMOD
/path/to/oras/go.mod
```

So the dependency check has been passing by accident, scanning the root module.

0.9.0 added a `validateGoModFile()` that opens the declared path before the chdir:

```go
func (resolver *GoModResolver) Resolve(goModFile string, config *ConfigDeps, report *Report) error {
	if err := validateGoModFile(goModFile); err != nil {   // os.ReadFile -> ENOENT
		return err
	}
	if err := os.Chdir(filepath.Dir(goModFile)); err != nil {
```

That `os.ReadFile` on the nonexistent `.github/go.mod` is the error above. This is a pre-existing misconfiguration in this repo that 0.9.0 surfaces, not a regression in skywalking-eyes.

### Fix

One line in `.github/licenserc.yml` — make the path relative to the config file, since that is what `Finalize()` joins against:

```diff
--- a/.github/licenserc.yml
+++ b/.github/licenserc.yml
@@ -44,4 +44,4 @@
 dependency:
   files:
-    - go.mod
+    - ../go.mod
```

That is the entire change: `.github/licenserc.yml` line 46, `- go.mod` -> `- ../go.mod`. `filepath.Join(".github", "../go.mod")` cleans to `go.mod` at the repo root, which is the module we actually want scanned — the same module 0.8.0 was reaching accidentally, so the set of dependencies checked does not change.

**Do not** change these, they are already correct:

- `header.paths-ignore` (lines 31-40) — header globs are matched against the repo root, so the `go.mod` / `go.sum` / `go.work` entries there are right as-is.
- The `config: .github/licenserc.yml` inputs in `.github/workflows/license-checker.yml` — the config location is fine; only the path *inside* it is wrong.

Moving `licenserc.yml` to the repo root would also work but is a larger, unnecessary change.

### Verifying locally

```console
$ go install github.com/apache/skywalking-eyes/cmd/license-eye@v0.9.0
$ license-eye -v info -c .github/licenserc.yml dependency check
```

Fails with the `open .../.github/go.mod` error before the change, passes after.

### Prior art

Identical bug and fix in oras-go: oras-project/oras-go#1359, fixed by oras-project/oras-go#1366. This repo never received the port.

### Merge order

The fix is also correct under the currently pinned 0.8.0 (it resolves to the same root `go.mod` that version reaches by walking up), so it can land on its own first. Once it is on `main`, #2149 needs a rebase — `@dependabot rebase` — for its `check-license` run to go green.

akshita317:
Fixes #2154

## Problem

`check-license` fails on any PR that touches dependencies (e.g. #2149, Dependabot bumping the `github-actions` group to `apache/skywalking-eyes` 0.9.0):

```
INFO  Loading configuration from file: .github/licenserc.yml
ERROR open /home/runner/work/oras/oras/.github/go.mod: no such file or directory
ERROR one or more errors occurred checking license compatibility
```

`.github/licenserc.yml` declares `dependency.files: [go.mod]`. skywalking-eyes' `ConfigDeps.Finalize()` joins relative `files` entries against the **config file's own directory**, so that entry has always meant `.github/go.mod` — which does not exist.

It passed by accident under 0.8.0: `GoModResolver.Resolve` only did `os.Chdir(filepath.Dir(goModFile))`, and chdir into `.github/` succeeds, after which `go mod download` walks *up* and finds the real root module. 0.9.0 added `validateGoModFile()`, which `os.ReadFile`s the declared path **before** the chdir and fails with ENOENT.

So this is a pre-existing misconfiguration in this repo that 0.9.0 surfaces, not a regression in skywalking-eyes. The dependency check has been silently scanning the root module by accident.

## Fix

One line: point the entry at `../go.mod`, so `filepath.Join(".github", "../go.mod")` cleans to the repo-root `go.mod`. That is the same module 0.8.0 was reaching accidentally, so **the set of dependencies checked is unchanged**.

Deliberately left alone:
- `header.paths-ignore` — header globs are matched against the repo root, so its `go.mod` / `go.sum` / `go.work` entries are already correct.
- The `config: .github/licenserc.yml` input in `license-checker.yml` — the config location is fine; only the path *inside* it was wrong.

## Verification

Locally with `license-eye` v0.9.0:

```console
$ go install github.com/apache/skywalking-eyes/cmd/license-eye@v0.9.0

# before
$ license-eye -v info -c .github/licenserc.yml dependency check
INFO  Loading configuration from file: .github/licenserc.yml
ERROR open .../oras/.github/go.mod: no such file or directory

# after
$ license-eye -v info -c .github/licenserc.yml dependency check
INFO  Loading configuration from file: .github/licenserc.yml
$ echo $?
0
```

`dependency resolve` after the change lists the root module's tree (`oras.land/oras-go/v2 v2.6.2`, `spf13/cobra`, `sirupsen/logrus`, `golang.org/x/crypto`, …), confirming it resolves the intended module rather than silently doing nothing.

## Prior art

Identical bug and fix in oras-go: oras-project/oras-go#1359, fixed by oras-project/oras-go#1366. This repo never received the port.

### DIFF
--- a/.github/licenserc.yml
+++ b/.github/licenserc.yml
@@ -43,7 +43,7 @@ header:
 
 dependency:
   files:
-    - go.mod
+    - ../go.mod
   licenses:
     # known issue: https://github.com/apache/skywalking-eyes/pull/107#issuecomment-1129761574
     - name: github.com/chzyer/logex


TerryHowe (changes requested):
DCO failure https://github.com/oras-project/oras/pull/2158/checks?check_run_id=101483809157 TLDR `git commit --amend -sa; git push --force`

TerryHowe:
Sorry, there was a PR for this issue earlier https://github.com/oras-project/oras/pull/2155

PR #2164, opened 2026-09-12
test(display): cover the template and descriptor metadata handlers

akshita317:
## Problem

Two metadata packages had **no test file at all**:

```
cmd/oras/internal/display/metadata/template     0.0% of statements
cmd/oras/internal/display/metadata/descriptor   0.0% of statements
```

Both render user-facing output — `--format go-template` for every command that supports it, and `oras manifest fetch --descriptor` — so a regression in either is immediately visible to users but was caught by nothing.

## Changes

Tests for every handler in both packages. They assert the **rendered bytes**, not merely that no error came back, so a change in output format fails the test.

**`descriptor`**
- compact and pretty output both round-trip back to the same descriptor
- only the pretty form is indented and newline-terminated; the compact form stays on one line so it can be piped
- a failing writer surfaces its error
- the reference and raw manifest bytes handed to `OnFetched` do not leak into the output

**`template`** — attach, discover, manifest fetch, push, repo ls, repo tags: each over its `Render` path and an unparsable template, plus the branches worth pinning down:
- `discover` rejects a referrer whose subject was never seen, and accepts a referrer of a previously discovered referrer
- `manifest fetch` tolerates non-JSON content by rendering a null manifest rather than failing
- `push` does not record a digest reference as a tag, and records nothing when no reference is given

```
cmd/oras/internal/display/metadata/template     85.3% of statements
cmd/oras/internal/display/metadata/descriptor   80.0% of statements
```

## One thing noticed while writing these

`model.NewPush` builds `ReferenceAsTags` with `var refAsTags []string` and only ever appends, so when nothing is tagged the field marshals as `null` rather than `[]`. `model.Repositories` and `model.Tags` both initialise their slices to `[]string{}` instead.

The visible effect is that a template calling `{{len .referenceAsTags}}` errors with `len of nil pointer` after an untagged push, while the same template shape works for `repo ls` and `repo tags`. I left it alone here to keep this PR test-only — happy to open an issue or a follow-up PR if you'd like it made consistent.

### DIFF
--- a/cmd/oras/internal/display/metadata/descriptor/manifest_fetch_test.go
+++ b/cmd/oras/internal/display/metadata/descriptor/manifest_fetch_test.go
@@ -1,0 +1,110 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package descriptor
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"reflect"
+	"testing"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const testDigest = "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+
+var testDescriptor = ocispec.Descriptor{
+	MediaType: "application/vnd.oci.image.manifest.v1+json",
+	Digest:    testDigest,
+	Size:      100,
+}
+
+// errWriter fails every write, to exercise the output error path.
+type errWriter struct{}
+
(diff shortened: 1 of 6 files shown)


TerryHowe (approved):
LGTM. I fetched the branch and ran it locally: tests pass, `-race` is clean, `go vet` and `gofmt` are clean, and the coverage numbers reproduce exactly as claimed (template 85.3%, descriptor 80.0%).

What I like here is that the assertions are on the exact rendered output rather than just a nil error, so a format regression actually fails. The branch selection is thoughtful too — unknown-subject rejection in `discover`, referrer-of-a-referrer acceptance, non-JSON content rendering a null manifest, and digest-reference-is-not-a-tag are exactly the branches that would otherwise rot silently. `OnFetched_ignoresNameAndContent` pins a real contract, and the descriptor round-trip test checking that only the pretty form is indented and newline-terminated is what protects the pipe-ability of the compact form.

Two follow-ups, neither blocking:

- `template/pull.go` is the one handler in the package still at 0%, and it is the entire remaining gap. Filed as #2165.
- `repo_test.go` covers both `repo_list.go` and `repo_tags.go`; every other test file under `display/metadata/` pairs 1:1 with its source. Worth splitting into `repo_list_test.go` and `repo_tags_test.go` whenever this area is next touched.

And your `model.NewPush` observation is correct — `var refAsTags []string` marshals to `null`, so `{{len .referenceAsTags}}` fails after an untagged push where the same template works for `repo ls` and `repo tags`. Agreed it belongs in a separate PR; please do open an issue for it.

PR #2168, opened 2026-09-14
test(display): cover the pull metadata handler

### LINKED ISSUE #2165: Add tests for the go-template pull metadata handler
## Problem

`cmd/oras/internal/display/metadata/template/pull.go` has no test coverage. Every method of `PullHandler` is at 0% of statements:

```
template/pull.go:38  NewPullHandler   0.0%
template/pull.go:47  OnPulled         0.0%
template/pull.go:52  Render           0.0%
template/pull.go:57  OnFilePulled     0.0%
template/pull.go:62  OnLayerSkipped   0.0%
```

#2164 added tests for the rest of the `template` package (attach, discover, manifest fetch, push, repo ls, repo tags) and brought it from 0% to 85.3%. `pull.go` is the entire remaining gap — covering it would take the package to roughly 100%.

This handler renders the user-facing output of `oras pull --format go-template`, so a change in its output is immediately visible to users but is currently caught by nothing.

## Suggested coverage

Following the style established in #2164 — assert the rendered bytes, not merely that no error came back:

- `Render` over a template exercising both the reference and the file list. Note it builds the reference as `path + "@" + root.Digest.String()` unconditionally, so the digest form is worth pinning.
- `Render` with an unparsable template returns an error.
- `Render` with no files pulled, i.e. before any `OnFilePulled` call.
- `OnFilePulled` for a file with a title annotation and for one without, plus the error path when `descPath` cannot be resolved.
- `OnPulled` records the root descriptor and ignores the target.
- `OnLayerSkipped` is a no-op returning nil.

## Related

- #2164 covered the rest of the package.

Soundcreates:

Adds focused unit tests for the go-template pull metadata handler. The tests pin the digest-form reference and rendered pulled-file metadata, cover empty pulls and unparsable templates, verify title and title-less descriptors, ensure the target is ignored by `OnPulled`, and exercise `OnLayerSkipped`.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>` format, will close issue(s) when PR gets merged)*:
Fixes #2165

- [x] Does the affected code have corresponding tests, e.g. unit test, E2E test?
- [ ] Does this change require a documentation update?
- [ ] Does this introduce breaking changes that would require an announcement or bumping the major version?
- [x] Do all new files have an appropriate license header?


cmd/oras/internal/display/metadata/template/pull_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestPullHandler_Render_noFiles(t *testing.T) {
+	buf := &bytes.Buffer{}
+	handler := NewPullHandler(buf, "localhost:5000/test", "{{.reference}}|{{len .files}}").(*PullHandler)
+	handler.OnPulled(nil, testDescriptor)
+
+	if err := handler.Render(); err != nil {
+		t.Fatalf("Render() error = %v, want nil", err)
+	}
+	want := "localhost:5000/test@" + testDigest + "|0"
```
TerryHowe: This test fails as written: `model.Pulled.files` is nil until the first `Add`, so `.files` marshals to `null` and `len` errors.

```
--- FAIL: TestPullHandler_Render_noFiles (0.00s)
    pull_test.go:84: Render() error = template: format output:1:17: executing "format output" at <len .files>: error calling len: len of nil pointer, want nil
```

`range` tolerates nil, so this keeps the empty-pull case covered:

```suggestion
	handler := NewPullHandler(buf, "localhost:5000/test", "{{.reference}}|{{range .files}}{{.path}};{{end}}").(*PullHandler)
	handler.OnPulled(nil, testDescriptor)

	if err := handler.Render(); err != nil {
		t.Fatalf("Render() error = %v, want nil", err)
	}
	want := "localhost:5000/test@" + testDigest + "|"
```

Side note for maintainers: the underlying behavior is arguably a bug — `oras pull --format '{{len .files}}'` errors when nothing was pulled. Defaulting to an empty slice in `model.NewPull` would fix it, but that's out of scope for a test-only PR.

cmd/oras/internal/display/metadata/template/pull_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestPullHandler_Render_noFiles(t *testing.T) {
+	buf := &bytes.Buffer{}
+	handler := NewPullHandler(buf, "localhost:5000/test", "{{.reference}}|{{range .files}}{{.path}};{{end}}").(*PullHandler)
+	handler.OnPulled(nil, testDescriptor)
+
+	if err := handler.Render(); err != nil {
+		t.Fatalf("Render() error = %v, want nil", err)
+	}
+	want := "localhost:5000/test@" + testDigest + "|""
```
TerryHowe:
```suggestion
	want := "localhost:5000/test@" + testDigest + "|"
```

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #2170, opened 2026-09-14
fix(display): report referenceAsTags as [] instead of null when untagged

### LINKED ISSUE #2169: Metadata models report null instead of [] for empty lists (push referenceAsTags, pull files)
## Problem

Two metadata models emit `null` instead of `[]` when their list is empty. The other list-carrying models initialise to an empty slice, so the same template or JSON consumer behaves differently depending on the command.

Raised in review of #2164.

| Model field | Initialised as | JSON when empty |
|---|---|---|
| `discover` `referrers` | `[]*Node{}` | `[]` |
| `repo ls` `repositories` | `[]string{}` | `[]` |
| `repo tags` `tags` | `[]string{}` | `[]` |
| **`push` `referenceAsTags`** | `var refAsTags []string`, append-only | **`null`** |
| **`pull` `files`** | never initialised; `Files()` returns `slices.Clone(nil)` | **`null`** |

### `oras push` with no tag

`model.NewPush` builds `referenceAsTags` with `var refAsTags []string` and only appends, so an untagged push marshals it as `null`:

```json
{ "referenceAsTags": null }
```

A go-template taking its length fails outright:

```console
$ oras push localhost:5000/repo file.txt --format go-template='{{len .referenceAsTags}}'
Error: template: format output:1:2: executing "format output" at <len .referenceAsTags>:
error calling len: len of nil pointer
```

The same shape works for `oras repo tags --format go-template='{{len .tags}}'` and `oras repo ls --format go-template='{{len .repositories}}'`.

### `oras pull` with no titled layer

`PullHandler.OnFilePulled` is only called for layers carrying `org.opencontainers.image.title` (`cmd/oras/root/pull.go`), so pulling an artifact with no titled layer leaves `Pulled.files` nil, and `Files()` returns `slices.Clone(nil)`, which is also nil. `"files"` then marshals as `null`.

`{{range .files}}` — what the e2e suite uses — tolerates either form, which is why this has gone unnoticed. `{{len .files}}` does not.

## Proposal

Initialise both to an empty slice so empty lists are always `[]`. For `push` this is a one-line change:

```diff
-	var refAsTags []string
+	refAsTags := make([]string, 0, len(tags))
```

I'll open a PR for `push` with regression tests at the model, JSON-handler and template-handler levels. `pull` is the same root cause in `model/pull.go`; happy to fold it into the same PR or do it separately, whichever you prefer. (#2002 touches that file, but only its import block.)

Nothing in the repository asserts on `null` here: the JSON push test compares with `slices.Equal`, which treats nil and empty alike, and the e2e suite only ranges over both fields.

akshita317:
Fixes #2169 (the `push` half).

## Problem

`model.NewPush` built `referenceAsTags` with `var refAsTags []string` and only appended, so an **untagged push** marshalled it as `null`:

```json
{ "referenceAsTags": null }
```

and a go-template taking its length failed outright:

```
error calling len: len of nil pointer
```

`repo ls` (`repositories`), `repo tags` (`tags`) and `discover` (`referrers`) all initialise to an empty slice and report `[]`, so the same template shape worked for those commands but not for `push`. Confirmed in review of #2164.

## Fix

```diff
-	var refAsTags []string
+	refAsTags := make([]string, 0, len(tags))
```

Empty now marshals as `[]`, and the slice is preallocated to its final length.

## Tests

Regression tests at each layer the value passes through, all written first and **confirmed failing on `main`**:

| Test | On `main` | With fix |
|---|---|---|
| `model` `TestNewPush_referenceAsTags` (nil / empty / 1 / 2 tags) | nil and empty cases: `referenceAsTags = null, want []` | pass |
| `json` `TestPushHandler_Render_untagged` | `referenceAsTags = null, want []` | pass |
| `template` `TestPushHandler_Render_untaggedLen` (`{{len .referenceAsTags}}`) | `len of nil pointer` | pass |

The model test compares the raw JSON value, so it distinguishes `null` from `[]` rather than accepting either. The existing tagged-push assertions are unchanged and still pass.

## Compatibility

For an untagged `oras push --format json`, `referenceAsTags` changes from `null` to `[]`, matching the other list fields. Nothing in the repository depended on `null`: the existing JSON push test compares with `slices.Equal`, which treats nil and empty alike, and the e2e suite only ranges over the field.

`go test ./...`, `gofmt` and `go vet` are clean.

## Not in this PR

`pull`'s `files` has the same root cause (see #2169) and lives in `model/pull.go`. I've kept this PR to `push` as discussed; happy to follow up with `pull` separately or add it here, whichever you prefer.


cmd/oras/internal/display/metadata/model/push.go
```diff
@@ -27,7 +27,7 @@ type push struct {
 
 // NewPush returns a metadata getter for push command.
 func NewPush(desc ocispec.Descriptor, path string, tags []string) any {
-	var refAsTags []string
+	refAsTags := make([]string, 0, len(tags))
```
TerryHowe:
```suggestion
	refAsTags := []string{}
```

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

Issue #1728, opened 2025-05-12
Resursively copying a multi-plarform image might fail when the index manifest has no referrer


qweeah:
### What happened in your environment?

Crafted below artifacts in a registry, where `A` is a image index with `manifests` field pointing a list of manifests that contains only image manifest `B`, and `C` is a referrer that points to `B` as its subject.

```mermaid
graph TD
    A -- manifests ---> B
    C -- subject --> B
```

When running `oras cp -r` with source as `A`, the copy process fails

### What did you expect to happen?

Should succeed to copy all the three artifacts.

### How can we reproduce it?

```bash
REG=localhost:6000
REPO=test
TAG=demo

SUBJECT=$(oras push $REG/$REPO -a test=true --format=go-template={{.reference}}) # generate image manifest
oras attach --artifact-type test/demo $SUBJECT -a test=true # attach a referrer to manifest
oras manifest index create  $REG/$REPO:$TAG $SUBJECT # create index
oras cp -r $REG/$REPO:$TAG --to-oci-layout sandbox:test # copy index to an OCI layout, will fail
```

The logs of `oras cp` is
```console
[2025-05-12T00:49:35.144358495Z][DEBUG]: --> Request #0
> Request URL: "http://localhost:6000/v2/test/manifests/demo"
> Request method: "HEAD"
> Request headers:
   "User-Agent": "oras/1.3.0-beta.3"
   "Accept": "application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, application/vnd.oci.artifact.manifest.v1+json"


[2025-05-12T00:49:35.14791852Z][DEBUG]: <-- Response #0
< Response Status: "401 Unauthorized"
< Response headers:
   "X-Content-Type-Options": "nosniff"
   "Date": "Mon, 12 May 2025 00:49:35 GMT"
   "Content-Length": "147"
   "Content-Type": "application/json; charset=utf-8"
   "Docker-Distribution-Api-Version": "registry/2.0"
   "Www-Authenticate": "Basic realm=\"test-basic\""
< Response body:
   No response body to print


[2025-05-12T00:49:35.148017021Z][DEBUG]: --> Request #1
> Request URL: "http://localhost:6000/v2/test/manifests/demo"
> Request method: "HEAD"
> Request headers:
[...]

Issue #1403, opened 2024-06-12
Oras Discover - json output include details of the image itself and not just directly attached artifacts (option for all artifacts)


akcrisp:
### What is the version of your ORAS CLI

1.2

### What would you like to be added?

update the --format json option to include details of the image itself including digest and application type. 

Can we also have a flag to output all artifacts whether direct or indirect in the json output. So for example if i have an sbom attached as an artifact, which is signed,  - the json output shows the sbom details but no mention of the signature for the sbom.

### Why is this needed for ORAS?

At moment because the json doesn't include the image details i am havig to discover that info by other means. Ideally the three supported output formats of json / tree and table should be aligned so they all output the same - ideally everything.(or indirect via a switch if need be)

### Are you willing to submit PRs to contribute to this feature?

- [ ] Yes, I am willing to implement it.

qweeah:
I can see two requirements raised in this issue:
1) Metadata of the subject image should be included in the JSON output of `oras discover`
2) Support showing indirect referrers in the JSON output of `oras discover` (This requires redesign fields in [v1.2.0's model](https://github.com/oras-project/oras/blob/v1.2.0/docs/proposals/formatted-output.md#oras-discover) and might introduce breaking changes)

/cc @FeynmanZhou @yizha1 

@akcrisp Feel free to add more details

sajayantony:
Nested artifacts can be challenging but the subject of the referrers can be included as a descriptor in the format output.

qweeah:
https://github.com/qweeah/oras/releases/tag/v1.2.0-discover
Sharing a demo of my own private build back in June. This private build 

1) Adds a new field `manifests` to the format model of `oras discover` to show metadata of the referrers.
2) Adds a new flag `-r/recursive` to show indirect referrers.

To install
```bash
## 1. Installation
VERSION="1.2.0-discover" &&\
curl -LO "https://github.com/qweeah/oras/releases/download/v${VERSION}/oras_${VERSION}_linux_amd64.tar.gz" &&\
mkdir -p oras-install/ &&\
tar -zxf oras_${VERSION}_*.tar.gz -C oras-install/
```

Here is a demo of listing indirect referrers in the formatted output
```console
> ./oras-install/oras discover mcr.microsoft.com/oss/kubernetes/kubectl:v1.29.1 --format json -r
{
[...]

PR #2174, opened 2026-09-18
fix: keep an index's own children out of its reported referrers

### LINKED ISSUE #2148: cp -r fails with 'MANIFEST_UNKNOWN'

Since `cosign copy` tells me to use `oras cp -r` instead, I started testing oras. But the copy of the GitLab images fail with a 'maninfest unknown'. 

When doing the copy without the `-r`, the copy succeeds but no signature is copied.

When doing the `copy -r` again after that, the copy succeeds but still no signature is copied.

A succesfull copy of GitLab registry images, including signatures so we can verify images with cosign.

### How can we reproduce it?

- Start a registry (tested with harbor and docker registry)
```
docker run -p 5000:5000 --restart always --name registry registry:3
```
- Copy `registry.gitlab.com/gitlab-org/build/cng/gitlab-workhorse-ee:v19.3.0` to the registry
```
oras cp -dr --to-insecure registry.gitlab.com/gitlab-org/build/cng/gitlab-workhorse-ee:v19.3.0 localhost:5000/gitlab-workhorse-ee:v19.3.0 > cp-r_1.txt 2>&1
```
- See the 'manifest unknown' error
```
{"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown","detail":"unknown tag=sha256-11e4068791cb99f2871ddbbeb5b3c213814a7cba4299d33acc5d31b1171064cb"}]}
```

Attached the logfile of the `copy -rd`

Copy with `oras cp -dr` fails
[cp-r_1.txt](https://github.com/user-attachments/files/31644043/cp-r_1.txt)

Copy with `oras cp -d` succeeds
[cp.txt](https://github.com/user-attachments/files/31644042/cp.txt)

Copy with `oras cp -dr` succeeds, but no signature
[cp-r_2.txt](https://github.com/user-attachments/files/31644044/cp-r_2.txt)

Version:        1.3.4
Go version:     go1.25.14
OS/Arch:        windows/amd64
Git commit:     db9e29505c3059f2b8fde34ae8cae266c5c765e9
Git tree state: clean

Windows 11 (git bash)

somaz94:

`oras cp -r` fails with `MANIFEST_UNKNOWN` when the source registry does not implement the Referrers API. ORAS falls back to the referrers tag schema and requests `sha256-<hex>`; a registry that resolves that as the digest `sha256:<hex>` answers with the index itself, so the index's own children are read back as its referrers.

`prepareCopyOption` (cmd/oras/root/cp.go:293) passes those through untouched. `oras.ExtendedCopyGraph` then walks upwards from the index into the children, takes them for the graph roots and copies only their sub-graphs, so the index is never pushed. The `dst.Tag` at cmd/oras/root/cp.go:279 asks the destination for a manifest that is not there, which is the 404 at the end of the log in the issue. It is also why a plain `oras cp` followed by `cp -r` succeeds: the index is already at the destination by then.

A successor can never be a predecessor, so this drops any reported referrer the index already lists in its own `manifests` (cmd/oras/root/cp.go:327). The index is left with no predecessors, becomes a graph root itself and is copied before it is tagged. Genuine referrers are untouched, and a child that both appears in `manifests` and declares the index as its subject is still copied, as a successor.

The upstream half of this report, oras-project/oras-go#1396, was fixed in oras-project/oras-go#1403 on oras-go `main`, which is v3. This repo is on `oras.land/oras-go/v2 v2.6.2` and no v2 release carries it, so the failure still reproduces here.

Scope: the fix covers `cp -r` and `backup`, since `backup --include-referrers` goes through the same `recursiveCopy`. `restore` and `countReferrers` read OCI layouts, so they were never affected. `oras discover` still lists the phantom referrers, because it calls `registry.Referrers` on the remote directly (cmd/oras/root/discover.go:173), and `oras attach` has the same exposure one layer down in oras-go's `updateReferrersIndex`. Those two wait on a v2 backport of oras-project/oras-go#1403.

On the alternative suggested in the issue, an `oras.CopyGraph` of the root before `dst.Tag`: I built that first and measured a regression in `cp -r --force`. `--force` wraps the destination in `contentutil.TraversingTarget` (cmd/oras/root/cp.go:197), whose `Exists` returns false for every manifest media type (internal/contentutil/traverse.go:56), so the extra `CopyGraph` cannot short-circuit and walks the manifest graph a second time. Against two local registries that turned a 14 line `cp -r --force` into 20 lines, with the index and both child manifests pushed twice. Clearing the status callbacks does not cover the TTY path, because the progress rows come from the tracked target's `Push` rather than from the handler. I am happy to switch to that shape if you prefer it.

Fixes #2148

**Local validation** (macOS arm64, Go 1.26.7):

`go test -run Test_recursiveCopy_strandedRoot ./cmd/oras/root/` on unmodified main: `FAIL cp_test.go:666: recursiveCopy() error = failed to perform "Tag" on destination: sha256:d2ef7640...: application/vnd.oci.image.index.v1+json: not found`. With the change: `ok`.

`go test -race ./...`: all packages ok. `go test -race -count=50 ./cmd/oras/root/ -run Test_recursiveCopy`: ok. `golangci-lint run ./...`: 0 issues. `gofmt -l cmd internal`: clean. `go vet ./...`: clean.

End to end against two local `registry:3` instances, comparing this build with one built from `main`. Copying a two platform index that carries a referrer: first copy, idempotent recopy and `cp -r --force` recopy all produce output identical to `main`. The #1728 shape, an index with no referrer of its own but a referrer on one child, is also identical, and the child's referrer arrives at the destination.


cmd/oras/root/cp.go
```diff
@@ -318,16 +318,37 @@ func prepareCopyOption(ctx context.Context, src oras.ReadOnlyGraphTarget, _ oras
...
+			return slices.ContainsFunc(index.Manifests, func(child ocispec.Descriptor) bool {
+				return content.Equal(child, predecessor)
+			})
+		}), nil
+	}
+
+	if len(referrers) == 0 {
+		// no child referrers
+		return opts, root, nil
+	}
```
TerryHowe, line 350: Suggestion for the two points above.

**Updated.** This originally fetched each candidate and kept a child that declares `root` as its subject. That branch is unreachable: a child of index `I` cannot carry `digest(I)` in its subject, because `digest(I)` is taken over JSON that already contains the child's digest, so the pair is unconstructible. It could only ever fall through to the drop — at the cost of an extra manifest GET per child on every `FindPredecessors(root)` call, and a new hard-failure path when a listed child is not fetchable. Removed; the invariant is stated in the comment instead.

What is left:

- collects into a fresh slice, since `descs` belongs to the wrapped implementation — `slices.DeleteFunc` edits it in place and zeroes the tail, and the later `append(rootReferrers, referrers...)` can write into that same backing array;
- keeps `src.Predecessors` as the effective default on the no-child-referrers path, so predecessor discovery there is otherwise unchanged (moot for the CLI, which always sets `FindPredecessors` in `doCopy`, but it matters for library callers);
- names the wrapper so both call sites share it.

Verified on your branch with this applied: `go build`, `go vet`, `gofmt`, `go test ./...` and `go test -race ./...` all clean, and all ten existing `Test_prepareCopyOption_*` / `Test_recursiveCopy_*` tests still pass.

```suggestion
	// A registry without Referrers API support may resolve the referrers tag
	// sha256-<hex> as the digest sha256:<hex> and answer with the index itself,
	// reporting the index's own children as its referrers. A child can never be
	// a genuine referrer of its own parent: it would have to carry digest(root)
	// in its subject, while digest(root) is taken over JSON that already
	// contains the child's digest, so the pair is unconstructible. Left in
	// place these phantom referrers make extended copy walk up into the
	// children, take those for the graph roots and never copy the index, so the
	// root tagging in recursiveCopy has nothing to tag.
	// Reference: https://github.com/oras-project/oras/issues/2148
	type findPredecessorsFunc = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error)
	dropOwnChildren := func(findPredecessors findPredecessorsFunc) findPredecessorsFunc {
		return func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
			descs, err := findPredecessors(ctx, src, desc)
			if err != nil || !content.Equal(desc, root) {
				return descs, err
			}
			// Collect into a new slice: descs belongs to the wrapped
			// implementation and must not be modified in place.
			kept := make([]ocispec.Descriptor, 0, len(descs))
			for _, predecessor := range descs {
				if slices.ContainsFunc(index.Manifests, func(child ocispec.Descriptor) bool {
					return content.Equal(child, predecessor)
				}) {
					continue
				}
				kept = append(kept, predecessor)
			}
			return kept, nil
		}
	}

	if len(referrers) == 0 {
		// No child referrers, but root still needs the guard above. Wrap
		// oras-go's own default instead of registry.Referrers so that
		// predecessor discovery on this path is otherwise unchanged.
		findPredecessors := opts.FindPredecessors
		if findPredecessors == nil {
			findPredecessors = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
				return src.Predecessors(ctx, desc)
			}
		}
		opts.FindPredecessors = dropOwnChildren(findPredecessors)
		return opts, root, nil
	}

	if opts.FindPredecessors == nil {
		opts.FindPredecessors = func(ctx context.Context, src content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
			return registry.Referrers(ctx, src, desc, "")
		}
	}
	opts.FindPredecessors = dropOwnChildren(opts.FindPredecessors)
```

cmd/oras/root/cp_test.go
```diff
@@ -601,3 +601,121 @@ func Test_recursiveCopy_tagFailure(t *testing.T) {
...
+	if !errors.As(err, &copyErr) {
+		t.Fatalf("recursiveCopy() error = %v, want *oras.CopyError", err)
+	}
+	if copyErr.Op != "Push" {
+		t.Errorf("recursiveCopy() error Op = %q, want %q", copyErr.Op, "Push")
+	}
+	if copyErr.Origin != oras.CopyErrorOriginDestination {
+		t.Errorf("recursiveCopy() error Origin = %v, want %v", copyErr.Origin, oras.CopyErrorOriginDestination)
+	}
+}
```
TerryHowe, line 721: One more, on the same #2148 filter: nothing currently asserts it does not *over*-filter. Both new tests cover the phantom-referrer case; neither pins down that a genuine referrer of the root survives alongside it. That is the one behaviour the filter could plausibly break, so it is worth a guard.

Appending here since it is a pure addition to the end of the file.

Notes on it:

- It passes against both your current version and the `dropOwnChildren` variant I suggested on `cp.go` — it guards the filter, not one particular implementation of it.
- Mutation-checked: making the filter drop everything fails it with `left the referrer out of the destination`, while `Test_recursiveCopy_strandedRoot` still passes under that same mutation. So it covers something the existing tests do not.
- Unlike the other two, this one also passes on unmodified `main`, because the genuine referrer's `subject` edge drags the index in as a successor. That is worth having pinned down: it is exactly why #2148 only bites when the index has no real referrers of its own.
- `gofmt` clean, passes under `-race`.

```suggestion
}
// Test_recursiveCopy_genuineRootReferrerSurvives guards the filter above
// against over-filtering: a real referrer of the root must still be copied
// even when the same FindPredecessors call also reports the index's own
// children as referrers.
func Test_recursiveCopy_genuineRootReferrerSurvives(t *testing.T) {
	ctx := context.Background()
	src := memory.New()
	push := func(blob []byte, mediaType string) ocispec.Descriptor {
		desc := ocispec.Descriptor{
			MediaType: mediaType,
			Digest:    digest.FromBytes(blob),
			Size:      int64(len(blob)),
		}
		if err := src.Push(ctx, desc, bytes.NewReader(blob)); err != nil {
			t.Fatal(err)
		}
		return desc
	}

	configDesc := push(configContent, configMediaType)
	childDesc := push([]byte(fmt.Sprintf(`{"schemaVersion":2,"mediaType":%q,"config":{"mediaType":%q,"digest":%q,"size":%d},"layers":[]}`,
		ocispec.MediaTypeImageManifest, configDesc.MediaType, configDesc.Digest, configDesc.Size)), ocispec.MediaTypeImageManifest)
	indexDesc := push([]byte(fmt.Sprintf(`{"schemaVersion":2,"mediaType":%q,"manifests":[{"mediaType":%q,"digest":%q,"size":%d}]}`,
		ocispec.MediaTypeImageIndex, childDesc.MediaType, childDesc.Digest, childDesc.Size)), ocispec.MediaTypeImageIndex)
	// A genuine referrer of the index, i.e. a manifest declaring it as subject.
	referrerDesc := push([]byte(fmt.Sprintf(`{"schemaVersion":2,"mediaType":%q,"artifactType":"application/vnd.test.referrer","config":{"mediaType":%q,"digest":%q,"size":%d},"layers":[],"subject":{"mediaType":%q,"digest":%q,"size":%d}}`,
		ocispec.MediaTypeImageManifest, configDesc.MediaType, configDesc.Digest, configDesc.Size,
		indexDesc.MediaType, indexDesc.Digest, indexDesc.Size)), ocispec.MediaTypeImageManifest)

	opts := oras.DefaultExtendedCopyGraphOptions
	opts.FindPredecessors = func(_ context.Context, _ content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
		if content.Equal(desc, indexDesc) {
			// The phantom child and the genuine referrer, reported together.
			return []ocispec.Descriptor{childDesc, referrerDesc}, nil
		}
		return nil, nil
	}

	dst := memory.New()
	if err := recursiveCopy(ctx, src, dst, "v1", indexDesc, opts); err != nil {
		t.Fatalf("recursiveCopy() error = %v, wantErr false", err)
	}
	for name, desc := range map[string]ocispec.Descriptor{
		"index":    indexDesc,
		"child":    childDesc,
		"referrer": referrerDesc,
	} {
		exists, err := dst.Exists(ctx, desc)
		if err != nil {
			t.Fatal(err)
		}
		if !exists {
			t.Errorf("recursiveCopy() left the %s out of the destination", name)
		}
	}
}
```

cmd/oras/root/cp.go
somaz94: Applied in 9dcaca5, thanks. Also rebased on main.

cmd/oras/root/cp_test.go
somaz94: Added in 9dcaca5, with one blank line before the doc comment so gofmt stays clean. I reran your mutation check and got the same result: dropping every referrer fails this test and leaves `Test_recursiveCopy_strandedRoot` passing.

TerryHowe (commented):
Two follow-ups, plus a scope note.

Scope: `oras discover` still reports the phantom referrers — `discover.go:173` calls `registry.Referrers` on the remote repo directly and has no equivalent filter, so an index there still lists its own children (the `visited` map stops the recursion, not the wrong output). `oras attach` has the same exposure one layer down, since oras-go's `updateReferrersIndex` shares the `referrersFromIndex` call that oras-go#1403 fixed. Not asking you to widen the PR — but worth a line in the description that `cp -r` and `backup` are the intended blast radius (`backup --include-referrers` gets this for free via `recursiveCopy`, `restore` and `countReferrers` read OCI layouts so they were never affected), and that the rest waits on a v2 backport of #1403.

cmd/oras/root/cp_test.go
```diff
@@ -601,3 +601,178 @@ func Test_recursiveCopy_tagFailure(t *testing.T) {
...
+	} {
+		exists, err := dst.Exists(ctx, desc)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !exists {
+			t.Errorf("recursiveCopy() left the %s out of the destination", name)
+		}
+	}
+}
```
TerryHowe, line 778: Suggestion: nothing currently pins `cp.go:372`, the second `dropOwnChildren` wrap. Verified three ways before posting:

- passes on your head;
- with line 372 reverted it fails with the exact #2148 error — `failed to perform "Tag" on destination: sha256:d2ef7640…: application/vnd.oci.image.index.v1+json: not found`;
- it takes `cp.go:384-397`, the #1728 `copyRoot` path, from statement coverage `0` to covered.

Under that same mutation the rest of the suite stays green, so this is the only thing that would catch a regression there. It is also the shape `oras attach` produces in practice: the referrer sits on a child, not on the index. `gofmt` clean, passes under `-race`.

Appending at the end of the file, same as before.

```suggestion
}

// Test_recursiveCopy_strandedRootWithChildReferrer covers the #1728 shape of
// https://github.com/oras-project/oras/issues/2148: the index has no referrer
// of its own, but a child does, so FindPredecessors reports the phantom
// referrers on the path where the child referrer list is non-empty.
func Test_recursiveCopy_strandedRootWithChildReferrer(t *testing.T) {
	ctx := context.Background()
	src := memory.New()
	push := func(blob []byte, mediaType string) ocispec.Descriptor {
		desc := ocispec.Descriptor{
			MediaType: mediaType,
			Digest:    digest.FromBytes(blob),
			Size:      int64(len(blob)),
		}
		if err := src.Push(ctx, desc, bytes.NewReader(blob)); err != nil {
			t.Fatal(err)
		}
		return desc
	}

	configDesc := push(configContent, configMediaType)
	childDesc := push([]byte(fmt.Sprintf(`{"schemaVersion":2,"mediaType":%q,"config":{"mediaType":%q,"digest":%q,"size":%d},"layers":[]}`,
		ocispec.MediaTypeImageManifest, configDesc.MediaType, configDesc.Digest, configDesc.Size)), ocispec.MediaTypeImageManifest)
	indexDesc := push([]byte(fmt.Sprintf(`{"schemaVersion":2,"mediaType":%q,"manifests":[{"mediaType":%q,"digest":%q,"size":%d}]}`,
		ocispec.MediaTypeImageIndex, childDesc.MediaType, childDesc.Digest, childDesc.Size)), ocispec.MediaTypeImageIndex)
	// A genuine referrer of the child, not of the index.
	referrerDesc := push([]byte(fmt.Sprintf(`{"schemaVersion":2,"mediaType":%q,"artifactType":"application/vnd.test.referrer","config":{"mediaType":%q,"digest":%q,"size":%d},"layers":[],"subject":{"mediaType":%q,"digest":%q,"size":%d}}`,
		ocispec.MediaTypeImageManifest, configDesc.MediaType, configDesc.Digest, configDesc.Size,
		childDesc.MediaType, childDesc.Digest, childDesc.Size)), ocispec.MediaTypeImageManifest)

	opts := oras.DefaultExtendedCopyGraphOptions
	opts.FindPredecessors = func(_ context.Context, _ content.ReadOnlyGraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
		switch {
		case content.Equal(desc, indexDesc):
			// Phantom: the index's own child, reported as its referrer.
			return []ocispec.Descriptor{childDesc}, nil
		case content.Equal(desc, childDesc):
			return []ocispec.Descriptor{referrerDesc}, nil
		}
		return nil, nil
	}

	dst := memory.New()
	if err := recursiveCopy(ctx, src, dst, "v1", indexDesc, opts); err != nil {
		t.Fatalf("recursiveCopy() error = %v, wantErr false", err)
	}
	for name, desc := range map[string]ocispec.Descriptor{
		"index":    indexDesc,
		"child":    childDesc,
		"referrer": referrerDesc,
	} {
		exists, err := dst.Exists(ctx, desc)
		if err != nil {
			t.Fatal(err)
		}
		if !exists {
			t.Errorf("recursiveCopy() left the %s out of the destination", name)
		}
	}
}
```

cmd/oras/root/cp.go
```diff
@@ -318,8 +318,49 @@ func prepareCopyOption(ctx context.Context, src oras.ReadOnlyGraphTarget, _ oras
...
+			}
+			return kept, nil
+		}
+	}
+
 	if len(referrers) == 0 {
-		// no child referrers
+		// No child referrers, but root still needs the guard above. Wrap
+		// oras-go's own default instead of registry.Referrers so that
+		// predecessor discovery on this path is otherwise unchanged.
```
TerryHowe: Comment-only. `doCopy` sets `FindPredecessors` unconditionally at cp.go:180, so the nil branch below is unreachable from the CLI — statement coverage on it is `0`. Keeping it is right for library callers, but the current wording reads as though this is the path `oras cp` takes.

```suggestion
		// No child referrers, but root still needs the guard above. The CLI
		// always sets FindPredecessors in doCopy; this nil case is for library
		// callers. Default to src.Predecessors rather than registry.Referrers
		// so that predecessor discovery on this path is otherwise unchanged.
```

cmd/oras/root/cp_test.go
somaz94: Added in d7f7abd. I reran your mutation check and got the same result: without the second `dropOwnChildren` wrap this test fails with the Tag not found error, and the rest of the suite still passes. The description now also has the scope note from your review.

cmd/oras/root/cp.go
somaz94: Applied in d7f7abd, thanks.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #2175, opened 2026-09-19
fix(display): report files as [] instead of null when a pull produces none

akshita317:
Refs #2169 — the `pull` half.

A note on that issue: it tracked both halves, but it auto-closed when #2170 merged because I used a closing keyword there while deliberately fixing only `push`. That was my mistake; happy to reopen it if you would rather track the remaining half there.

## Problem

`model.NewPull` passes the caller's slice straight into `files`, and `Pulled.Files()` returns `slices.Clone` of a nil slice — which is also nil. So a pull that produced no files marshals as:

```json
{ "files": null }
```

and a go-template taking its length fails outright:

```
error calling len: len of nil pointer
```

`OnFilePulled` only fires for layers carrying `org.opencontainers.image.title`, so an artifact with no titled layer reaches this path.

This is the same inconsistency fixed for `push` in #2170: `repositories`, `tags` and `referrers` all report `[]` when empty, so the same template shape worked for every command except `pull`.

## Fix

```diff
 func NewPull(digestReference string, files []File) any {
+	if files == nil {
+		files = []File{}
+	}
```

`NewPull` is the single serialisation boundary — both the JSON and template handlers call it with `ph.pulled.Files()` — so one guard covers both.

## Tests

Written first and **confirmed failing on `main`**:

| Test | On `main` | With fix |
|---|---|---|
| `model` `TestNewPull_files` (nil / empty / one file) | `files = null, want []` | pass |
| `json` `TestPullHandler_Render_noFiles` | `files = null, want []` | pass |
| `template` `TestPullHandler_Render_noFilesLen` (`{{len .files}}`) | `len of nil pointer` | pass |

The model test compares the raw JSON value, so it distinguishes `null` from `[]` rather than accepting either.

## Compatibility

For a pull with no files, `--format json` changes `files` from `null` to `[]`. Nothing depended on `null`: the template `TestPullHandler_Render_noFiles` added in #2168 uses `{{range .files}}`, which tolerates either form and still passes, and the e2e suite only ranges over the field.

`go test ./...`, `gofmt` and `go vet` are clean.


TerryHowe (approved):
/lgtm

PR #2176, opened 2026-09-19
test(display): cover the metadata model types

akshita317:
## Problem

`cmd/oras/internal/display/metadata/model` is at **13.7% of statements**. `push.go` was the only type with a test file, so the types behind `oras attach`, `discover`, `manifest fetch`, `repo ls` and `repo tags` had no tests at all — despite defining the JSON and go-template output users consume.

## Changes

One test file per source file, following the 1:1 pairing used elsewhere under `display/metadata` (and asked for on #2164). The assertions are on the marshalled shape, not merely that a constructor returns non-nil:

| File | What it pins |
|---|---|
| `descriptor_test.go` | `NewDigestReference`; `FromDescriptor` copying the core fields, inlining `reference`/`mediaType`/`digest`/`size` as templates address them, and **deliberately dropping** `URLs`, `Data` and `Platform` |
| `discover_test.go` | `NewNode` starting `Referrers` as an empty slice; a referrer of a previously discovered referrer attaching to *that* node, not the root; an unknown subject rejected with `unexpected subject descriptor` |
| `attach_test.go` | the rendered reference and descriptor fields |
| `fetched_test.go` | content round-tripping, and nil content marshalling as `null` rather than failing — what lets the fetch handlers tolerate non-JSON content |
| `repo_list_test.go` / `repo_tags_test.go` | both lists initialised so an empty listing marshals as `[]`, not `null` |
| `tag_test.go` | `Tags` sorting, which `push` relies on for `referenceAsTags` |

The `[]`-not-`null` assertions are deliberate: that is the regression class fixed in #2170 and #2175, and nothing was pinning it for the lists that already behaved correctly.

## Result

```
before: coverage: 13.7% of statements
after:  coverage: 58.8% of statements
```

Every function is now at 100% except the four in `pull.go` (`newFile`, `NewPull`, `Files`, `Add`). Those are left alone on purpose to avoid conflicts: #2175 covers `NewPull`, and #2002 also touches that file. Happy to fold them in here instead if you would rather.


TerryHowe (approved):
/lgtm

PR #2177, opened 2026-09-19
fix: only infer platform from OCI image configs

### LINKED ISSUE #2178: Incorrect platform inference for non-image configs

oras manifest index create incorrectly infers a platform from a non-image config.

In my local environment, I created an OCI artifact with a custom config media type containing "os" and "architecture" fields. When creating an index from the artifact, ORAS added a  platform field to the descriptor based on those fields, even though the config was not an OCI image config.

For example, the config contained:

```json
{"os":"linux","architecture":"amd64"}
```

with the config media type:

```text
application/vnd.example.fake.config+json
```

The resulting index descriptor was incorrectly assigned:

```json
"platform": {
  "os": "linux",
  "architecture": "amd64"
}
```

ORAS should not infer a platform from a config unless the config uses the OCI image config media type.

For a non-image config, the resulting index descriptor should have no platform field, even if the config JSON contains os and architecture fields.

### Steps to reproduce

1.  Create a config containing os and architecture:
    
```json
{"os":"linux","architecture":"amd64"}
```

2.  Push an OCI artifact using a non-image config media type:
    
```bash
oras push \
  --oci-layout /tmp/oras-weird-platform:fake \
  --artifact-type application/vnd.example.fake \
  --config fake-config.json:application/vnd.example.fake.config+json \
  oras-amd64.txt:application/octet-stream
```

3.  Create an index from the artifact:

```bash
oras manifest index create \
  /tmp/oras-weird-platform:index \
  fake \
  --oci-layout
```

4.  Inspect the resulting index:
    
```bash
oras manifest fetch \
  --oci-layout /tmp/oras-weird-platform:index
```

The descriptor for the artifact incorrectly contains:

```json
"platform": {
  "os": "linux",
  "architecture": "amd64"
}
```

The config media type is application/vnd.example.fake.config+json, not application/vnd.oci.image.config.v1+json

Version: 1.3.4
Go version: go1.26.7
OS/Arch: linux/amd64

CachyOS Linux (linux/amd64)

TanvirTian:
oras manifest index create currently infers a platform from the config blob of every OCI image manifest, regardless of the config media type.

This means a non-image artifact with a JSON config containing os and architecture fields can incorrectly receive a platform field when added to an image index.

This PR limits platform inference to manifests whose config media type is application/vnd.oci.image.config.v1+json, which is the OCI image config media type.
A regression test was added to verify that platform information is not inferred from a non-image config containing os and architecture fields.

Fixes #2178 


cmd/oras/root/manifest/index/create.go
```diff
@@ -183,6 +183,9 @@ func fetchSourceManifests(ctx context.Context, displayStatus status.ManifestInde
 }
 
 func getPlatform(ctx context.Context, target oras.ReadOnlyTarget, manifest *ocispec.Manifest) (*ocispec.Platform, error) {
+	if manifest.Config.MediaType != ocispec.MediaTypeImageConfig {
+		return nil, nil
+	}
```
TerryHowe: Docker-format manifests (`vnd.docker.distribution.manifest.v2+json`) also reach here via `descriptor.IsImageManifest`, and their config is `vnd.docker.container.image.v1+json` — so this drops their platform. (Needs `MediaTypeConfig` added to `internal/docker/mediatype.go` + the import here.)

```suggestion
	if manifest.Config.MediaType != ocispec.MediaTypeImageConfig &&
		manifest.Config.MediaType != docker.MediaTypeConfig {
		return nil, nil
	}
```

TanvirTian: thanks. now updated to also allow the docker config media type and added a test for it :)

TerryHowe (changes requested):
Lint issue fix and I think we are g2g

TanvirTian:
> Lint issue fix and I think we are g2g
Fixed the lint issue and pushed the changes. All checks are passing now. Thanks :)

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #2182, opened 2026-09-21
test(display): cover the raw content handlers

akshita317:
## Problem

`cmd/oras/internal/display/content` is at **37.9% of statements**. `manifest_fetch.go` has no test file at all — neither of its functions is exercised — and `manifest_index.go` is covered only on the path where creating the output file fails.

These handlers produce what `oras manifest fetch` and `oras manifest index create` write to stdout or to `--output`, so a regression is immediately visible to users but was caught by almost nothing.

## Changes

`manifest_fetch_test.go` (new, pairing 1:1 with its source) and additional cases in `manifest_index_test.go`. The existing index test is unchanged. Assertions are on the bytes that come out, not merely that no error is returned:

- **`--pretty` is dropped for a file** and kept for stdout and for `-`. Both constructors do this deliberately (`// ignore --pretty when output to a file`) but nothing pinned it, so the behaviour could have been lost silently.
- **The compact form is byte-for-byte the fetched content**, so it stays pipeable; only the pretty form is indented.
- **`--output` writes to the file and leaves stdout untouched.**
- **Error paths**: a path that cannot be created reports `failed to open`, a failing writer surfaces its error, and content that is not JSON fails while prettifying rather than being written out half-formed.

## Result

```
before: coverage: 37.9% of statements
after:  coverage: 100.0% of statements
```

Every function in the package is now at 100%:

```
discard.go            OnContentFetched               100.0%
discard.go            OnContentCreated               100.0%
discard.go            NewDiscardHandler              100.0%
manifest_fetch.go     OnContentFetched               100.0%
manifest_fetch.go     NewManifestFetchHandler        100.0%
manifest_index.go     NewManifestIndexCreateHandler  100.0%
manifest_index.go     OnContentCreated               100.0%
```


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #2183, opened 2026-09-21
fix: fix concurrency in Tagged

### LINKED ISSUE #2181: Tagged.Tags() sorts the shared slice under a read lock
## What happened

`model.Tagged.Tags()` sorts the shared slice while holding only a **read** lock:

https://github.com/oras-project/oras/blob/main/cmd/oras/internal/display/metadata/model/tag.go#L37-L42

```go
func (tag *Tagged) Tags() []string {
	tag.lock.RLock()
	defer tag.lock.RUnlock()
	slices.Sort(tag.tags)   // <-- writes to tag.tags under RLock
	return tag.tags
}
```

`slices.Sort` mutates the backing array in place, but `RLock` is shared — two concurrent `Tags()` calls both acquire it and sort the same array simultaneously. The mutex therefore does not provide the exclusion the type appears to promise.

`Tags()` also returns the internal slice directly, so a caller can mutate `tag.tags` after the lock is released.

## Reproduction

Dropping this into `cmd/oras/internal/display/metadata/model` and running `go test -race`:

```go
func TestTaggedRaceProbe(t *testing.T) {
	var tagged Tagged
	for _, s := range []string{"c", "a", "b"} {
		tagged.AddTag(s)
	}
	var wg sync.WaitGroup
	for i := 0; i < 8; i++ {
		wg.Add(1)
		go func() { defer wg.Done(); _ = tagged.Tags() }()
	}
	wg.Wait()
}
```

```
WARNING: DATA RACE
Read at 0x00c000180ae0 by goroutine 12:
  slices.insertionSortOrdered[go.shape.string]()
  slices.Sort[go.shape.[]string,go.shape.string]()
  ...model.(*Tagged).Tags()
      cmd/oras/internal/display/metadata/model/tag.go:41

Previous write at 0x00c000180ae0 by goroutine 10:
  slices.insertionSortOrdered[go.shape.string]()
  ...model.(*Tagged).Tags()
      cmd/oras/internal/display/metadata/model/tag.go:41
```

## Impact

**Latent, not currently reachable.** The only two call sites each invoke `Tags()` once, from `Render()`:

- `cmd/oras/internal/display/metadata/json/push.go:62`
- `cmd/oras/internal/display/metadata/template/push.go:64`

`AddTag` vs. `Tags` is correctly excluded (write lock vs. read lock), so no user-visible bug exists today. The problem is that `Tagged` is written as a concurrency-safe type — that is the entire reason it carries a `sync.RWMutex` — and it does not hold up if a second `Tags()` caller is ever added. That would be an easy and quiet regression to introduce.

## Suggested fix

Either:

1. Sort in `AddTag` under the write lock, leaving `Tags()` a pure read; or
2. Have `Tags()` take the write lock and return `slices.Clone(tag.tags)`.

(2) also fixes the returned-internal-slice aliasing. A concurrent `Tags()` subtest under `-race` would lock the behavior in.

## Context

Noticed while reviewing #2176, which adds the first test coverage for `Tagged` and pins the sorting behavior. That PR is test-only and correct as-is; this is a separate production-code follow-up.

TanvirTian:
Tagged.Tags() sorts the shared tag.tags slice while holding a read lock, which can cause a data race when called concurrently.

This changes Tags() to use an exclusive lock while sorting and return a cloned slice so callers cannot modify the internal slice.

Added tests covering concurrent access and returned-slice isolation.

Fixes #2181 


TerryHowe (approved):
/lgtm

TanvirTian:
> /lgtm

thanks :) . could you merge it when you get a chance?

TerryHowe:
Thanks!

